PR #35, opened 2016-12-14
Open-Issue: Sort the forms in the briefcase alphabetically #1

carstendev:
Sorry for not exactly following the contribution workflow, i´m new to contributing to OS projects.

I tackled the issue of sorting the FormTransferTable alphabetically. I basically tried to stick to the code style and used a private member class to encapsulate the logic.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/FormTransferTable.java
+++ b/src/org/opendatakit/briefcase/ui/FormTransferTable.java
@@ -26,13 +26,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import javax.swing.JButton;
-import javax.swing.JOptionPane;
-import javax.swing.JTable;
-import javax.swing.UIManager;
-import javax.swing.table.AbstractTableModel;
-import javax.swing.table.TableCellRenderer;
-import javax.swing.table.TableColumnModel;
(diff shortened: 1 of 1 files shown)


src/org/opendatakit/briefcase/ui/FormTransferTable.java
```diff
@@ -26,13 +26,8 @@
...
 import java.util.Map;
 
-import javax.swing.JButton;
-import javax.swing.JOptionPane;
-import javax.swing.JTable;
-import javax.swing.UIManager;
-import javax.swing.table.AbstractTableModel;
-import javax.swing.table.TableCellRenderer;
-import javax.swing.table.TableColumnModel;
+import javax.swing.*;
```
lognaturel: There isn't yet a style guide for this repo but I think that since it has been a convention throughout the code and is in most style guides (for example, [Google's](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports)), we will not use wildcard imports. Could you please change these back? I know it's hard to adhere to an invisible style guide, sorry!

carstendev: Sure, I´ll revert the changes on the imports, to not use wildcards.

PR #36, opened 2016-12-16
The text areas in the "Detailed Transfer Status for [form]" Dialog an…

### LINKED ISSUE #26: The text area in the Details... dialog should not be editable
#### Software and hardware versions 
ODK Briefcase v1.4.9 Production, Java 1.8.0_25, MacOS 10.12.1

#### Problem description
The text area in the `Detailed Transfer Status for [form]` dialog is editable.

1. Connect to a server with forms.
1. Click on a form's `Details...` button.
1. Type in the `Detailed status` text area.

#### Expected behavior
The text area should not be editable.

carstendev:
…d the "Export Details for [form]" Dialog are now set to editable false.  Fix #26.

I set the <tt>isEditable</tt> flag on the <tt>JEditorPane</tt> that gets used to display the details to <tt>false</tt>.
By doing this, the editor pane can´t be edited anymore. 

The <b>problem description</b> only mentioned, that the text area in the <tt>Detailed Transfer Status for [form]</tt> Dialog was editable. The same behavior could also be observed in the <tt>Export Details for [form]</tt> Dialog.

The implemented fix also ensures, that the <tt>Export Details for [form]</tt> Dialog editor pane can´t be edited. <br><br>I believe, this is the desired behavior for both Dialogs.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/ScrollingStatusListDialog.java
+++ b/src/org/opendatakit/briefcase/ui/ScrollingStatusListDialog.java
@@ -70,6 +70,7 @@ public class ScrollingStatusListDialog extends JDialog implements ActionListener
     getRootPane().setDefaultButton(cancelButton);
 
     JEditorPane editorArea = new JEditorPane("text/plain", statusHtml);
+    editorArea.setEditable(false);
   //Put the editor pane in a scroll pane.
     JScrollPane editorScrollPane = new JScrollPane(editorArea);
     editorScrollPane.setVerticalScrollBarPolicy(


lognaturel:
Thanks for taking the time to notice the other places where this problem existed and fixing it across the board!

mitchellsundt:
Note sure whether editable also impacts select-ability. This text should be able to be selected and copied for pasting into bug reports, etc..

carstendev:
Hey @mitchellsundt, JEditorPane's are editable by <i>default</i>. Invoking <tt>setEditable(false)</tt> makes the JEditorPane uneditable. It is still <i>selectable</i> and the user can copy data from it, but the user <b>cannot</b> change the JEditorPane's contents.

PR #45, opened 2016-12-26
This resolves #22. The last successful server setting for both push and

### LINKED ISSUE #22: Server settings should persist between restarts
It's pretty annoying to retype server settings on application restart. Server settings (both for pull and push) should persist across restarts.

carstendev:
pull will now be saved as a preference.

I am not entirely sure, why GitHub is marking the entire files as changed. I tried every whitespace setting I could find, but nothing changes. My local diff also seems to be working fine. If reviewing the pull request like this is too much work, I will investigate a little more and hopefully find a solution. 

To implement the feature, I extended the functionality of the <tt>BriefcasePrefereces</tt> class, so that we are now able to manage preferences per class and not only application wide. 

The username and URL are saved as a preference after they have been successfully used to connect to a server. If an <b>Aggregate 0.9.x</b> server is used, the URL and token are saved. The preferences for Aggregate 1.0 and Aggregate 0.9.x servers are managed <b>separately</b> by the pull tab.

To load the preferences into the server connection dialog, I used the already supplied <tt>ServerConnectionInfo</tt> class and added the username/ token and URL on instance initialization.
The contents of the instance are then displayed by the connection dialog.

### DIFF
--- a/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
+++ b/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
@@ -62,14 +62,43 @@ public class BriefcasePreferences {
     return new BriefcasePreferences(node, PreferenceScope.CLASS_NAME);
   }
   
+  /**
+   * Returns the value associated with the specified key in this preference
+   * node. Returns the specified default if there is no value associated with
+   * the key, or the backing store is inaccessible.
+   * 
+   * @param key
+   *          key whose associated value is to be returned.
+   * @param defaultValue
+   *          the value to be returned in the event that this preference node
+   *          has no value associated with key.
+   * @return the value associated with key, or defaultValue if no value is associated
+   *         with key, or the backing store is inaccessible.
+   */
   public String get(String key, String defaultValue) {
     return preferences.get(key, defaultValue);
   }
   
+  /**
+   * Associates the specified value with the specified key in this preference
+   * node.
+   * 
+   * @param key
+   *          key with which the specified value is to be associated.
+   * @param value
+   *          value to be associated with the specified key.
+   */
   public void put(String key, String value) {
     preferences.put(key, value);
   }
   
+  /**
+   * Removes the value associated with the specified key in this preference
+   * node, if any.
+   * 
+   * @param key
+   *          key whose mapping is to be removed from the preference node.
+   */
   public void remove(String key) {
     preferences.remove(key);
   }
--- a/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
+++ b/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
@@ -47,14 +47,6 @@ public class BriefcasePreferences {
   }
   
   /**
-   * Factory that returns the application scoped <tt>BriefcasePreferences</tt> object.
-   * The method always return the same instance.
-   */
-  public static BriefcasePreferences applicationScoped() {
-    return Preference.APPLICATION_SCOPED;
-  }
-  
-  /**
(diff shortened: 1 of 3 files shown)


lognaturel:
Hey @carstendev, thanks so much for doing this! I haven't forgotten about you. I'm not too familiar with some of what's in here so I want to take my time to educate myself so I can thoughtfully evaluate some of the changes and additions. The behavior is just as desired. Happy New Year!

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
```diff
@@ -21,39 +21,133 @@
...
+      @Override
+      public Preferences preferenceFactory(Class<?> node) {
+        return Preferences.userRoot().node(node.getName());
+      }
+    };
+    public abstract Preferences preferenceFactory(Class<?> node);
+  }
+  
+  /**
+   * Static member class, that holds the instance to the application scoped
```
lognaturel: Thanks for writing a bit of an explanation here. I think it's still a little obscure so I'd recommend linking out to https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom so someone looking at the code can learn more about the pattern and why it works.

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
```diff
@@ -21,39 +21,133 @@
...
+   */
+  public void remove(String key) {
+    preferences.remove(key);
+  }
 
   public static void setBriefcaseDirectoryProperty(String value) {
-    if ( value == null ) {
-      getApplicationPreferences().remove(BRIEFCASE_DIR_PROPERTY);
+    if (value == null) {
+      Preference.APPLICATION_SCOPED.remove(BRIEFCASE_DIR_PROPERTY);
```
lognaturel: I believe that every use of `Preference.APPLICATION_SCOPED` should be replaced by `BriefcasePreferences.applicationScoped()` to take advantage of the lazy loading.

src/org/opendatakit/briefcase/ui/PullTransferPanel.java
```diff
@@ -1,536 +1,553 @@
...
-  }
-
-  @EventSubscriber(eventClass = RetrieveAvailableFormsFailedEvent.class)
-  public void formsAvailableFromServer(RetrieveAvailableFormsFailedEvent event) {
-    ODKOptionPane.showErrorDialog(PullTransferPanel.this,
-        "Accessing the server failed with error: " + event.getReason(), "Accessing Server Failed");
-  }
-
-}
+/*
```
lognaturel: These are really the strangest diffs ever. I have no idea what's causing the weirdness! Maybe character encoding? Strange!

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
carstendev: This is a very good idea, I will change the explanation and have it include a link.

src/org/opendatakit/briefcase/ui/PullTransferPanel.java
carstendev: That's exactly what I am thinking, but every character encoding change I tried, made it worse.

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
carstendev: You're right, I wrote it to provide access to the <tt>Preference.APPLICATION_SCOPED</tt> instance from other classes. But since we're only accessing the instance through its specific methods, we don't need the factory.

PR #47, opened 2016-12-30
Added description in connection error message 

### LINKED ISSUE #24: Error when protocol isn't specified could be more helpful.
#### Software and hardware versions 
ODK Briefcase v1.4.9 Production, Java 1.8.0_25, MacOS 10.12.1

#### Problem description
The error message shown when a URL without a protocol is specified looks like:
```
Unable to fetch formList: Invalid url:
opendatakit.appspot.com/formList.
Failed with error: no protocol: opendatakit.appspot.com/formList
```

1. Click the "Connect..." button to open the "Aggregate v1.x Server Connection" dialog.
1. Enter in `opendatakit.appspot.com` or another URL without a specified protocol (e.g. `http://`, `https://`)
1. Click the `Connect` button

#### Expected behavior
The error is not very human-friendly. It should say something like "Did you forget to specify http:// or https://?" to help the user.

anakiou:
Added description in connection error message in case protocol is not specified Fixes #24

**Before**

before

**After**

after

### DIFF
--- a/src/org/opendatakit/briefcase/util/AggregateUtils.java
+++ b/src/org/opendatakit/briefcase/util/AggregateUtils.java
@@ -208,6 +208,9 @@ public class AggregateUtils {
       e.printStackTrace();
       String msg = description.getFetchDocFailed() + "Invalid url: " + urlString + ".\nFailed with error: "
           + e.getMessage();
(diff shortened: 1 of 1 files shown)


lognaturel:
At first I thought the String test should be replaced by a test on the exception type but [MalformedURLException](https://docs.oracle.com/javase/7/docs/api/java/net/MalformedURLException.html) can be thrown for reasons other than a missing protocol so this is good.

PR #50, opened 2017-01-08
Remove unneeded libraries

yanokwa:
These libraries are not used. I also compiled with mainClassName in build.gradle set to the various main classes.


build.gradle
```diff
@@ -11,18 +11,11 @@ dependencies {
...
-    compile group: 'commons-logging', name: 'commons-logging', version: '1.2'
     compile group: 'org.bushe', name: 'eventbus', version: '1.4'
-    compile group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.2'
-    compile group: 'org.apache.httpcomponents', name: 'httpcore', version: '4.4.4'
     compile group: 'org.apache.httpcomponents', name: 'httpmime', version: '4.5.2'
     compile group: 'net.sf.kxml', name: 'kxml2', version: '2.3.0'
-    compile group: 'eu.medsea.mimeutil', name: 'mime-util', version: '2.1.3'
-
     compile files('lib/javarosa-libraries-2016-04-21.jar')
-    compile files('lib/smallsql-0.21.jar')
```
lognaturel: This library is called manually and shouldn't be removed.

yanokwa:
Added smallsql library back. Tested to make sure I could download forms from a server and export to CSV.

PR #55, opened 2017-01-11
Changed the ScrollingStatusListDialog to live update.

### LINKED ISSUE #53: Status in Detailed Transfer Status window should live-update
#### Software versions 
Briefcase v1.4.9

#### Problem description
The status in the Detailed Transfer Status window is not updated as the transfer status changes.

1. Select a form
1. Click 'Push' or 'Pull'
1. Click on the 'Details...' button for that form

The Detailed Transfer Status window is opened but it does not update as the transfer status changes.

#### Expected behavior
The Detailed Transfer Status window should be updated as the transfer status changes.

#### Other information 
@carstendev I believe this is why the text area had been made editable.

carstendev:
This PR Resolves #53.

To implement this, I had to modify the <tt>ScrollingStatusListDialog</tt> to hold an instance of the form, it displays information for. This was needed, because <tt>ScrollingStatusListDialog</tt> instances have to filter out FormStatusEvents, that are <b>concurrently</b> published for different forms (e.g. when downloading multiple forms).


lognaturel:
Fabulous! Thank you so much for also thinking of the export screen, too. ✨

PR #57, opened 2017-01-23
Update version to 1.4.10.

lognaturel:


mitchellsundt:
FYI: What I have typically done is update the version # to:

   max( ODK Collect, ODK Aggregate) 

And often try to bundle these releases with an ODK Collect release.

The critical lock-step version # requirement is when you update the javarosa library.

lognaturel:
Ah, thanks for pointing that out. The key thing is that users must have a way of knowing which tool versions are compatible with each other, right? It might be good to think up a way to make that even more explicit somehow.

mitchellsundt:
Yes. We haven't strictly followed major/minor/point numbering, or we'd be up around 1.14 now.

Every bump in javarosa library produces a potential incompatible backward compatibility if a form is designed that uses the new features -- that form will break any older server. But, if those new features are not used, these ARE just point releases.

This is the tricky thing with this labeling and one of the reasons I began date-stamping the javarosa library so I could keep track of cross-(in)compatibilities.

PR #61, opened 2017-02-22
push pull button fixes

### LINKED ISSUE #52: Don't switch to Push tab when Pull is initiated
#### Software versions 
Briefcase v1.4.9

#### Problem description
Sometimes, when I initiate a pull of a form, I am switched to the Push tab automatically. I haven't investigated when it happens in detail but it may be limited to the first time a form is pulled after app launch.

1. Select a form
1. Click 'Pull'

#### Expected behavior
The user should remain on the pull tab until explicitly switching to the push tab.

#### Other information 
Similarly, triggering a push moves the user automatically to the export tab.

zakupower:
When the application starts, pull button was enabled. Now it's not. Pull and
push buttons switched to another tab when pressed. Now they don't.
fixes #52

### DIFF
--- a/src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java
+++ b/src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java
@@ -284,7 +284,7 @@ public class MainBriefcaseWindow implements WindowListener {
       public Component getComponentAfter(Container arg0, Component arg1) {
         ArrayList<Component> componentOrdering = new ArrayList<Component>();
         for (;;) {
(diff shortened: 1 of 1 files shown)


lognaturel:
Hello @zakupower and welcome to ODK Briefcase! We hang out at http://slack.opendatakit.org and encourage you to come say hi there.

In the future, please do comment on an issue you're planning to take on so we know it's being addressed. In this particular case, @carstendev had already claimed it! It's ok because this is a nice quick fix that is complementary to his eventual refactor but just be careful of that in the future. Thanks! 😊

PR #66, opened 2017-03-05
Included a chronological sorting for file instances in csv export

### LINKED ISSUE #58: Export to CSV should sort like Aggregate sorts
Feature request from a user: "I notice that the records are not chronologically export in the csv sheet as when you download directly from the server."

rclakmal:
Fixes  #58

Instances for CSV export are sorted in the chronological order by submission date. if no submission date, we assume it to be latest.


yanokwa:
@nribeka I've checked the functionality. How does this code look to you?

src/org/opendatakit/briefcase/util/ExportToCsv.java
```diff
@@ -28,12 +28,7 @@
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import java.text.DateFormat;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
nribeka: This is probably IDE config making star import when you're using more than 5 imports.

src/org/opendatakit/briefcase/util/ExportToCsv.java
```diff
@@ -134,6 +129,33 @@ public boolean doAction() {
...
     File[] instances = instancesDir.listFiles();
 
+    //Sorts the instances by the submission date. If no submission date, we assume it to be latest.
+    Arrays.sort(instances, new Comparator<File>() {
+      File submission1, submission2;
+      String submissionDate1String, submissionDate2String;
+      Date submissionDate1, submissionDate2;
+      public int compare(File f1, File f2) {
+        try {
+          submission1 = new File(f1, "submission.xml");
```
nribeka: We could potentially improve this by doing simple memoization of already parsed submission. That way we don't need to parse the same file again.

nribeka: Store the date of already parsed. That way when we encounter the same file, it will just read the cache instead of reparsing the file. I hope that make sense?

src/org/opendatakit/briefcase/util/StringUtils.java
```diff
@@ -0,0 +1,10 @@
+package org.opendatakit.briefcase.util;
```
nribeka: Need some copyright info here.

rclakmal:
@nribeka Thank you very much for the review. I have incorporated review comments.

src/org/opendatakit/briefcase/util/StringUtils.java
yanokwa: The code is yours, @rclakmal. Put your name and the correct date.

rclakmal: Done

src/org/opendatakit/briefcase/util/ExportToCsv.java
```diff
@@ -134,6 +136,46 @@ public boolean doAction() {
 
     File[] instances = instancesDir.listFiles();
 
+		// Sorts the instances by the submission date. If no submission date, we
+		// assume it to be latest.
+		Arrays.sort(instances, new Comparator<File>() {
```
lognaturel: Instances could be null at this point so it would be good to do a null check before sorting.

src/org/opendatakit/briefcase/util/ExportToCsv.java
```diff
@@ -134,6 +136,46 @@ public boolean doAction() {
 
     File[] instances = instancesDir.listFiles();
 
+		// Sorts the instances by the submission date. If no submission date, we
+		// assume it to be latest.
+		Arrays.sort(instances, new Comparator<File>() {
+			HashMap<File, String> submissionDateToFileMap = new HashMap<File, String>();
+			File submission1, submission2;
```
lognaturel: It's good practice to declare variables in the tightest scope possible. In this case, `submission1` and `submission2` should just be declared when they are first initialized (`File submission1 = new File(f1, "submission.xml");`)

`submissionDate1String` and `submissionDate2String` can be declared above the conditionals that set them. 

`submissionDate1` and `submissionDate2` can be declared where they are initialized.

src/org/opendatakit/briefcase/util/ExportToCsv.java
```diff
@@ -134,6 +136,46 @@ public boolean doAction() {
...
+				try {
+					if (submissionDateToFileMap.get(f1) == null) {
+						submission1 = new File(f1, "submission.xml");
+						submissionDate1String = XmlManipulationUtils.parseXml(submission1).getRootElement()
+								.getAttributeValue(null, "submissionDate");
+						submissionDateToFileMap.put(f1, submissionDate1String);
+					} else {
+						submissionDate1String = submissionDateToFileMap.get(f1);
+					}
+					if (submissionDateToFileMap.get(f2) == null) {
```
lognaturel: I would put a blank line before this conditional to indicate that it's a separate section of the code.

nribeka:
I think the comparator class will just gets created once (so there's only 1 map I think used to sort it) and get passed around during the sorting process. If the sorting is called on just a few file, I think it's fine to leave it out.

rclakmal:
@lognaturel Thank you for the review :) Incorporated your review comments. 
Regarding the map to cache the results, as @nribeka  pointed out, I think only one map is being created for a single `export` since the map belongs to the one time created Comparator class.

lognaturel:
Nice! @nribeka @rclakmal As the code is written, a new `Comparator` with a new map is created on each sort. If you wanted to reuse the map, you'd have to do something to keep the `Comparator` object around. I would recommend not doing this optimization, though.

rclakmal:
@lognaturel Removed the optimization done using the hashmap.

yanokwa:
I've confirmed this works on a dataset with all dates and with some dates. After @rclakmal cleans up the indentation, this will be ready for @lognaturel to merge.

yanokwa:
I spoken too soon! I found a bug...

If you have an invalid file (e.g., a hidden `.DS_Store`) in your Briefcase Storage Directory Briefcase throws an exception in the log. 
```
org.opendatakit.briefcase.model.FileSystemException: Failed while reading submission xml: java.io.FileNotFoundException: /Users/yanokwa/Library/Application Support/ODK/ODK Briefcase Storage/forms/choice_filter/instances/.DS_Store/submission.xml (Not a directory)
        at org.opendatakit.briefcase.util.XmlManipulationUtils.parseXml(XmlManipulationUtils.java:290)
        at org.opendatakit.briefcase.util.ExportToCsv$1.compare(ExportToCsv.java:150)
        at org.opendatakit.briefcase.util.ExportToCsv$1.compare(ExportToCsv.java:142)
        at java.util.TimSort.binarySort(TimSort.java:296)
        at java.util.TimSort.sort(TimSort.java:221)
        at java.util.Arrays.sort(Arrays.java:1438)
        at org.opendatakit.briefcase.util.ExportToCsv.doAction(ExportToCsv.java:142)
        at org.opendatakit.briefcase.util.ExportAction$TransformFormRunnable.run(ExportAction.java:57)
        at java.lang.Thread.run(Thread.java:745)
```

rclakmal:
@yanokwa Thanks for the catch. Added a fix to handle invalid files in the instances directory. Also fixed the indentation.

PR #72, opened 2017-03-18
Pulling can be triggered multiple times for the same form leading to ConcurrentModificationException #49

### LINKED ISSUE #49: Pulling can be triggered multiple times for the same form leading to ConcurrentModificationException
#### Software versions 
Briefcase v1.4.9

#### Problem description
Pulling can be triggered multiple times for the same form. This can sometimes result in a `ConcurrentModificationException` if the same instance happens to be downloaded at the same time. Plus, it's inefficient.

1. Select a form
1. Click on the pull button
1. Unselect the form and then reselect it, this makes the pull button active again
1. Click on the pull button, the form will be pulled a second time.

#### Expected behavior
The pull button should always be inactive when only forms that are currently being pulled are selected.

#### Other information 
I believe I first triggered this by starting a pull on one form, realizing I also wanted to pull a second one, and selecting that second one without unchecking the first. This is probably a more likely scenario. 

The easiest fix might be to make it impossible to select a form that is currently downloading.

rclakmal:
Fixes #49


lognaturel:
This currently blocks every row from being unchecked when there's an active transfer which has a strange consequence -- if I check a row while a pull is happening, I can't uncheck it. This is confusing so I think we should aim for only blocking the rows that have active transfers.

rclakmal:
@lognaturel How would you define an active transfer? Ones which have a pull status such as "fetching..." or "getting..."? In this case, we will have to do a string comparison of the pull status.

rclakmal:
@lognaturel What do you think about handling it in this way? Basically, we only block the row if it has a non-empty status.

lognaturel:
Nice idea! I think it's almost there but I can still trigger a double download by checking one form, clicking on 'pull', selecting another form while that first one is downloading and then clicking pull again.

rclakmal:
@lognaturel Handled this situation. Thanks for the catch :-)

Now, push/pull action acts only on forms which are selected AND not already in a fetch/upload mode.

rclakmal:
This is open for review.

lognaturel:
Thank you, @rclakmal and I apologize for having trouble keeping up!

Here's another strange case -- check one form, start a pull/push. Then check another one and start a pull/push. Once one of them succeeds, the other one can now be checked/unchecked even if it's still ongoing!

rclakmal:
@lognaturel Thanks for your detailed testing to make it right :-)

I realized this happened because the `transferStateActive` variable I used for infact a static variable and not reflecting the instance state of different transfers.  As an alternative, to find out whether the transfer is active, I used the termination text (Success! & Failed.) comparison. 

I tested it out and it seems to work fine. But a second look from you will certainly help.

lognaturel:
Ooh, I hadn't even noticed that it was static, good catch. This looks good! 

I do think it would be nice to have some actual state that represents the current status rather than having to read a UI string but we can do that incrementally.

PR #78, opened 2017-03-23
Fix packaging to use dependent jars

### LINKED ISSUE #65: Exporting encrypted csv file using ODK Briefcase fails on Windows
> I have been trying to export my encrypted test data using ODK Briefcase and my public-private key pair, but every time I export it, the csv file is blank except for the variable names in the header.  Has anyone run into this problem before or does anyone know if I am doing anything wrong.
> 
> I never noticed the message in the corner but no it did not complete.  I forgot to download the JCE but after I did so, it still failed.  Here is my error: 
> 
> Error decrypting submission uuid8d2c8480-d625-4058-86e2-57aab74f124a Cause: org.opendatakit.briefcase.model.CryptoException: Error decrypting base64EncryptedKey Cause: java.security.NoSuchAlgorithmException: Cannot find any provider supporting RSA/NONE/OAEPWithSHA256AndMGF1Padding
> FAILED!
> 
> I am using java 8 on my computer but downloaded JCE for the Java 7 version since it said so in the manual.  Should I downgrade my java to java 7?
> 
– bendo28@gmail.com

#### Software versions 
Briefcase v1.x.x, Java v1.x.x, operating system, Aggregate v1.x.x, Collect v1.x.x...

#### Other information 
Things you tried, stack traces, related issues, suggestions on how to fix it...

yanokwa:
* Stopped exploding dependent jars because it breaks BouncyCastle signing
* Included Eclipse classloader that came with previous versions of Briefcase to load jars
* Confirmed that Eclipse license is compatible ([see number 30](https://eclipse.org/legal/eplfaq.php)) and updated license docs
* Added explodedJar task to make for easier debugging
* Confirmed working on Windows 7 UI
* Confirmed broken on macOS (#79)
* Fixes #65


shivam-tripathi:
Solves the problem on Ubuntu 16.04 Linux.

rclakmal:
OS X Sierra (10. 12)
Double clicking on jar works. However, the encrypted export still fails. Updated with findings in the issue #79

PR #83, opened 2017-03-26
Duplication not allowed when pulling forms from collect(fixed issue #8)

icemc:


icemc:
@yanokwa  please could you review the changes i made. Thanks

icemc:
@lognaturel  please could you review the changes i made. Thanks

src/org/opendatakit/briefcase/util/TransferFromODK.java
```diff
@@ -216,12 +223,46 @@ public boolean accept(File dir, String name) {
...
+                  File[] contents = scratchInstance.listFiles(new FilenameFilter() {
+
+                    @Override
+                    public boolean accept(File dir, String name) {
+                      return name.endsWith(".xml");
+                    }});
                   if ( contents == null || contents.length == 0 ) break;
+
+                  if (contents.length == 1){
+                    String itsInstanceId  = null;
```
dcbriccetti: Extra space.

icemc:
@dcbriccetti  Thanks for the review. extra space removed

icemc:
@opendatakit/briefcase

src/org/opendatakit/briefcase/util/TransferFromODK.java
```diff
@@ -25,12 +25,8 @@
 
 import org.apache.commons.io.FileUtils;
 import org.bushe.swing.event.EventBus;
-import org.opendatakit.briefcase.model.FileSystemException;
-import org.opendatakit.briefcase.model.FormStatus;
-import org.opendatakit.briefcase.model.FormStatusEvent;
-import org.opendatakit.briefcase.model.BriefcaseFormDefinition;
-import org.opendatakit.briefcase.model.OdkCollectFormDefinition;
-import org.opendatakit.briefcase.model.TerminationFuture;
+import org.opendatakit.briefcase.model.*;
```
yanokwa: No star imports! Just import the classes you need.

src/org/opendatakit/briefcase/util/TransferFromODK.java
```diff
@@ -208,6 +205,16 @@ public boolean accept(File dir, String name) {
               }
               
               if (xml.exists()) {
+                //Check if the instance has an instanceID
+                try {
+                  XmlManipulationUtils.FormInstanceMetadata sim =
```
yanokwa: Confusing variable name. What does this actually represent?

src/org/opendatakit/briefcase/util/TransferFromODK.java
```diff
@@ -216,12 +223,45 @@ public boolean accept(File dir, String name) {
...
-                  String[] contents = scratchInstance.list();
+                  File[] contents = scratchInstance.listFiles(new FilenameFilter() {
+
+                    @Override
+                    public boolean accept(File dir, String name) {
+                      return name.endsWith(".xml");
+                    }});
                   if ( contents == null || contents.length == 0 ) break;
+                  if (contents.length == 1){
+                    String itsInstanceId  = null;
```
yanokwa: Nit pick, but there is an extra space before the =

yanokwa:
Hi @icemc! In your PR, please be sure to include how you tested that this functionality.

I tested this by generating two forms. One with an instanceID (`yes-id`) and one without (`no-id`). I filled each form twice on ODK Collect. Then I copied this [odk](https://github.com/opendatakit/briefcase/files/900016/odk.zip) folder with the forms and instances onto my desktop computer and pulled that data twice into Briefcase.

As expected, my Briefcase Storage folder had two submissions for `yes-id` and four submissions for `no-id`. 
screen shot 2017-04-05 at 08 58 20

I then exported to CSV and as expected, `yes_id.csv` had only two rows and `no_id.csv` had four rows. I also confirmed that with master, `yes_id.csv` had four rows, so this PR addresses the issue.

There is a subtle problem though. With Aggregate you, can de-dupe because you can guarantee that the instance data will not change. With Collect, you do not have this guarantee. That is, you can have an instance with an instanceID. You can pull that instance with Briefcase, then edit the form in Collect, and re-pull that data. The instance has changed, but the instanceID is the same. With this PR, you won't get the updated changes.

@lognaturel Proposed a good solution. If the instanceID is the same, then do an MD5 hash comparison. If the hash matches, then ignore the file. If it doesn't copy the file and append a number to the folder (just like Briefcase currently does).

icemc:
@yanokwa working on the changes requested, thanks

icemc:
@yanokwa I tested this by generating two forms. One with an instanceID (yes-id) and one without (no-id). I filled each form twice on ODK Collect. Then I copied this odk folder with the forms and instances onto my desktop computer and pulled that data twice into Briefcase.

As expected, my Briefcase Storage folder had two submissions for yes-id and four submissions for no-id
multiple pull
i then exported the forms with yes_id had only two rows while no_id had four rows. Repeated the procedure with master and yes_id had four rows.

Regarding the problem of pulling updated instances from collect. I updated one of the instances of yes_id and when i pulled the ODK data into Briefcase the updated instance was pulled even though there was another instance having the same instance ID in Briefcase
filechanged

So this pull request solves #8

yanokwa:
This works exactly how I imagined, @icemc. Well done!

@lognaturel @nribeka Can you review the code?

src/org/opendatakit/briefcase/util/TransferFromODK.java
```diff
@@ -216,12 +228,46 @@ public boolean accept(File dir, String name) {
...
+                    String itsInstanceId = null;
+                    try {
+                      XmlManipulationUtils.FormInstanceMetadata formInstanceMetadata =
+                              XmlManipulationUtils.getFormInstanceMetadata(XmlManipulationUtils.parseXml(contents[0])
+                                      .getRootElement());
+                      itsInstanceId = formInstanceMetadata.instanceId;
+
+                      //Check if the above ODK file(xml) instanceId is equal to this briefcase file instanceId then compare their MD5 hashes
+                      //if yes don't copy it, skip to next file
+                      if (itsInstanceId != null && instanceId.equals(itsInstanceId) &&
```
lognaturel: I think the null check here should be on `instanceId`. If `itsInstanceId` is null, that doesn't matter because you don't dereference it. But if `instanceId` is null, then you get a null pointer exception when calling `equals`.

icemc: true let me fix that and test

lognaturel: (or you can just flip the call so it's `itsInstanceId.equals(instanceId)`, either way is fine!)

icemc: @lognaturel  fixed the changes requested

yanokwa:
Thanks for making those changes, @icemc!

PR #84, opened 2017-03-27
Skip bad encrypted form instances during export. Fix #10

### LINKED ISSUE #10: Briefcase pulls incomplete instances of encrypted forms, causes error when decrypting
<a href="https://github.com/mitchellsundt"></img></a> **Issue by [mitchellsundt](https://github.com/mitchellsundt)**
_Thursday Jul 09, 2015 at 19:01 GMT_
_Originally opened as https://github.com/opendatakit/opendatakit/issues/671 (1 comment(s))_

----

Originally reported on Google Code with ID 670

```
What steps will reproduce the problem?
1.Use briefcase to pull instances from an encrypted form off of a device
2. One or more instances is incomplete
3. Decryption/exporting is interrupted and fails with the following error: Error decrypting
submission .... Cause:
org.opendatakit.briefcase.model.ParsingException: Missing
base64EncryptedKey element in encrypted form.


What is the expected output? What do you see instead?
Expected output is to export/decrypt without interruption, somehow skipping over the
incomplete instances that lack the encrypted key element.  Instead, the export fails
and it is necessary to manually remove incomplete instance from briefcase storage before
completing decryption.

What version of the product are you using? On what operating system?
Using ODK Collect v1.2 rev 1013, ODK Briefcase v1.0.2.1, XLSForm on Android 2.2

Please provide any additional information below.
```

Reported by `abarnett5989` on 2012-08-22 17:05:53

icemc:


icemc:
@yanokwa could you please review this code. What do u think about the approach i used?

lognaturel:
Thank you, @icemc! I haven't had a chance to review the functionality yet although it looks like a good approach but I notice a few instances that don't match the style of the rest of the file. [This style guide](https://google.github.io/styleguide/javaguide.html) has a good description of common style rules. In particular, there should be a space after `if`, a single space after `}`, all if/else/else if blocks should have curly braces even if they only include one line and else should be on the same line as the two braces around it (`} else {`). Please make those changes!

icemc:
@longnaturel will do them ASAP.

icemc:
@lognaturel  formatted the code as required

yanokwa:
@icemc Before we review this, please describe how you tested that this works.

icemc:
@yanokwa @ln During decryption of encrypted form instances several errors may occur such as parsing errors, decryption errors or even file system errors (in java ParsingException | CryptoException | FileSystemException). In master when such errors occur during export of an instance the export fails and doesn't proceed. But with this fix when these errors occur a popup dialog appears asking the user to skip the bad instance. if the user selects yes the export continue with other instances and skips the bad instance but if the user selects no then the export operation is aborted. That is how it worked for me.
The steps i used to produce the bad encrypted form is described [here](https://github.com/opendatakit/briefcase/issues/13#issuecomment-289858122)

yanokwa:
This is great, @icemc. I love the functionality! 

My concerns have to do with the dialogs. I don't think we should be using dialogs because we don't use them currently! Also, having users ignore the dialog forever doesn't feel great to me.

Instead, I wonder if we can just write the exception in the log and change the alert to `Succeeded with errors. See details`. What do you think?

icemc:
@yanokwa during pulling of form instances if one of the instances could not be downloaded then the is skipped and the whole process is marked as FAILED! even though the process completed and was not aborted. For my opinion i think such events should be marked as "SUCCESS WITH SOME ERRORS". So i went forward to implement  such an event during CSV export. If its that looks nice i could extend it to the pull process. Here is a screen shot
successerrors
Note that if all the files were skipped then the export process is marked as "FAILED".

PR #96, opened 2017-04-07
Re-download deleted submission file during next pull from aggregate server fixes #92

### LINKED ISSUE #92: Re-download deleted submission files during next pull from aggregate
Software versions 
Briefcase v1.4.10,  operating system  windows 8,

 Problem description
When a form instance(submission file) obtained by pulling the form from aggregate is deleted from its folder(usually named as the instance id). During next pull Briefcase does verify that instance file is present before skipping it instead verify if the instance folder is present. By doing so deleted instances are not re-downloaded. This will probably happen to users who play around with the briefcase storage directory and form instances

Steps to reproduce the problem

- Pull a form say **New Widgets** from an aggregate server (in my case aggregate 1.0). After downloading the instances move to your briefcase storage location and locate where the instances of the particular form are stored(in my case ....\ODK Briefcase Storage\forms\New Widgets\instances). 

- enter into one or more instance directory and delete the submission file (submission.xml)

- Go to briefcase and re-pull the form instances. The deleted instances are not re-downloaded

 Expected behavior
The deleted instances should be re-downloaded during the next pull

Other information

icemc:
When the submission file is deleted from the the instance folder. During the next pull from aggregate the submission file should be re-downloaded but master doesn't re-download the submission file.
A submission file is deleted as shown here
beforepull

During the next pull it is re-downloaded as shown
afterpull

so this pull request solves issue #92

### DIFF
--- a/src/org/opendatakit/briefcase/util/ServerFetcher.java
+++ b/src/org/opendatakit/briefcase/util/ServerFetcher.java
@@ -351,7 +351,8 @@ public class ServerFetcher {
       if ( instanceFolder != null ) {
           //check if the submission file is present in the folder before skipping
           File instance = new File(instanceFolder, "submission.xml");
-          if (instance.exists()) {
+          File instanceEncrypted = new File(instanceFolder, "submission.xml.enc");
+          if (instance.exists() || instanceEncrypted.exists()) {
               logger.info("already present - skipping fetch: " + uri );
               return;
           }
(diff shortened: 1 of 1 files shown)


yanokwa:
This looks good to me, @icemc! @lognaturel and @nribeka, can you take a look at the code?

lognaturel:
@icemc Thanks for including screenshots of what you've done to verify your changes!! That is super helpful.

PR #107, opened 2017-04-19
util/ExportToCsv: Only duplicate media on export when they are unique

### LINKED ISSUE #77: Only duplicate media on export when there are truly unique
Collect uses timestamps to name files. On a very large campaign, two devices can take images at the same time, upload them both to Aggregate, then download them to Briefcase. When those images are exported, what happens?

shivam-tripathi:
While exporting, if the destination folder already contains a file with the
same name, compute the MD5 hash of the files and compare them. If
same, skip copying; else append a suffix to differentiate the files.
For better performance, cache the already computed hash in a HashMap with
fileName-hashValue as key-value pair.

Fix #77

### DIFF
--- a/src/org/opendatakit/briefcase/util/ExportToCsv.java
+++ b/src/org/opendatakit/briefcase/util/ExportToCsv.java
@@ -70,6 +70,7 @@ public class ExportToCsv implements ITransformFormAction {
   BriefcaseFormDefinition briefcaseLfd;
   TerminationFuture terminationFuture;
   Map<TreeElement, OutputStreamWriter> fileMap = new HashMap<TreeElement, OutputStreamWriter>();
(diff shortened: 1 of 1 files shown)


shivam-tripathi:
Case: Same name, same image different instances.
Expected behaviour: Single image in media folder and csv file

Case: Same name, different image, different instances.
Expected behaviour:  Two images in media folder and csv file, one with suffix added.

Case: Same name, same image with different size, different instances 
Expected behaviour: Two images in media folder and csv file, one with suffix added.

Output matches expected behaviour.

shivam-tripathi:
Open for review.

yanokwa:
I can confirm this works exactly as imagined. Thanks for detailing the test cases, @shivam-tripathi!

PR #108, opened 2017-04-20
Addition of Progress Bars in export event 

shivam-tripathi:
Display progress bars while exporting

### DIFF
--- a/src/org/opendatakit/briefcase/util/ExportToCsv.java
+++ b/src/org/opendatakit/briefcase/util/ExportToCsv.java
@@ -27,6 +27,7 @@ import org.kxml2.kdom.Node;
 import org.opendatakit.briefcase.model.BriefcaseFormDefinition;
 import org.opendatakit.briefcase.model.CryptoException;
 import org.opendatakit.briefcase.model.ExportProgressEvent;
+import org.opendatakit.briefcase.model.ExportProgressPercentageEvent;
 import org.opendatakit.briefcase.model.FileSystemException;
 import org.opendatakit.briefcase.model.ParsingException;
 import org.opendatakit.briefcase.model.TerminationFuture;
@@ -77,6 +78,7 @@ public class ExportToCsv implements ITransformFormAction {
   boolean overwrite = false;
   int totalFilesSkipped = 0;
   int totalInstances = 0;
+  int processedInstances = 0;
   
   
   // Default briefcase constructor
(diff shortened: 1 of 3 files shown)


rclakmal:
@shivam-tripathi 

I tested this on Linux Mint. UI changes looks good :-)

One thing I noticed, if I press Cancel, while progress bar is half way through, I still end up with a "SUCCEEDED!" message.  When you check the exported files, they also seems to be exported correctly. Could it be that progress bar is actually triggered after the export is done? Because you also notice a delay in the beginning

shivam-tripathi:
@rclakmal: The progress bar depends on a separate event, and so it doesn't affect the label in any way. 
I tried pressing Cancel mid way, I get  a "Failed" message. Can you give me something so that I can see when it fails? :)

shivam-tripathi:
Open for review.

rclakmal:
@shivam-tripathi Sorry for the delay in checking back. I checked up with a larger set of forms data set and it works as expected.  Pls ignore my last comment because I cannot reproduce it anymore. Probably it was a  one time thing. 

In Linux Mint and Window 7, this is how it looks. There seems to be a little alignment issue of the progress bar in respect to the buttons. 

Windows
capture

Linux Mint
progressbar

src/org/opendatakit/briefcase/util/ExportToCsv.java
```diff
@@ -794,6 +796,7 @@ private boolean processInstance(File instanceDir) {
       return false;
     }
     EventBus.publish(new ExportProgressEvent("Processing instance: " + instanceDir.getName()));
+    EventBus.publish(new ExportProgressPercentageEvent((++processedInstances*100.0)/totalInstances));
```
lognaturel: Using the `++` operator inline like this is always a bit concerning to me. It's really easy to introduce strange off by one bugs if someone doesn't understand the difference between the prefix and suffix operator. I think it's worth adding another line to first increment the `processedInstances` variable and then calculate the percent.

Also please make sure to have spaces around all your operators (`processedInstances * 100.0` rather than processedInstances*100.0`).

shivam-tripathi:
@lognaturel @rclakmal Thank you for the feedback!
I have kept the alignment as it was earlier with the label, thinking there must be some reason for it being so. I too think alignment should be fixed, and will modify accordingly for both this and the code style.
Again, thanks for your feedback! :-)

shivam-tripathi:
Open for review.

lognaturel:
Nice! 👏👏

I think the alignment wasn't very visible when it was just the text far away from the buttons so no one noticed it!

PR #116, opened 2017-04-26
add eclipse project support to build.gradle

daveycrockett:
I'm more comfortable with Eclipse than IntelliJ, and since this was a one-line change I thought it would be acceptable to give folks options?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ local.properties
 .idea
 .DS_Store
 build
+.project
+.classpath
+.settings
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,7 @@ plugins {
     id 'application'
     id 'checkstyle'
     id 'de.fuerstenau.buildconfig' version '1.1.8'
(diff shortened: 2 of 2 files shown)


yanokwa:
Hi @daveycrockett! This is one of those decisions might have subtle effects, so let's have a discussion!

My concerns are: 
1. The maintainers don't use Eclipse. Adding this implies making sure Eclipse support works. And when it doesn't work,  we'll have an increased support burden.
1. We've settled on Android Studio on Collect and IntelliJ is functionally the same thing. Given that, you'll have to learn IntelliJ one way or another.
1. Adding Eclipse to the build file adds to the stuff people have to download to work with Briefcase. It's probably a small download, but we should check.

Does adding this Gradle plugin make your life significantly easier?

daveycrockett:
It doesn't, it's actually just as easy to add the line to build.gradle temporarily just to run `./gradlew eclipse`, generate the dotfiles, and then remove it.  Closing this out now...

PR #120, opened 2017-04-28
ui/MainBriefcaseWindow: Make "f" flag more generic

### LINKED ISSUE #118: Change "-f" (export_filename) flag CSV specific nature to a more general one in CLI option
Presently for flag 'f' in command line the argument name has been set to "name.csv". Briefcase will hopefully support other formats in future, so it must do away with .csv name attached.

shivam-tripathi:
Change "-f" (export_filename) flag from CSV specific to a more general one in CLI option

Fix #118

### DIFF
--- a/src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java
+++ b/src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java
@@ -552,10 +552,10 @@ public class MainBriefcaseWindow implements WindowListener {
         .build();
 
     Option exportFilename = Option.builder("f")
-        .argName("name.csv")
+        .argName("name.format")
         .hasArg()
         .longOpt(EXPORT_FILENAME)
-        .desc("File name for exported CSV")
+        .desc("File name for exported format")
         .build();
 
     Option overwrite = Option.builder("oc")


daveycrockett:
I'm not sure I understand the purpose of this issue.  Is the goal just to future-proof the _name_ of the CLI arg, or to actually make the export functionality itself more extensible?  If it's the former this seems to fulfill the goal, if it's the latter I would expect there would need to be more work in BriefcaseCLI, either adding an additional arg specifying output format or by looking at the extension in the output file argument, and in either case using a Factory pattern to determine which outputter class gets instantiated (ExportToCsv versus some other format ITransformFormAction implementation).

shivam-tripathi:
@daveycrockett Hi!
The aim is to just future proof the name of the argument. 
Also, it aims to be a sub part of the bigger issue of extending the CLI functionality, upon which I am working presently (apparently very slowly).

daveycrockett:
Makes sense, thanks for the clarification!  Sounds like this should be ready to merge whenever we're out of code freeze?

daveycrockett:
@opendatakit/briefcase anything remaining TODO with this PR?

yanokwa:
I think this PR is too early. We should first support different formats (https://github.com/opendatakit/briefcase/issues/4), then change the flag after that is done. I'm closing for that reason.

PR #132, opened 2017-06-15
Fix #128.

### LINKED ISSUE #128: Spinner in proxy port settings don't increment or persist well
* If I put in 999 as the port number and try to increment with the +, it doesn't increment. Anything below 999 works though. 
* Port numbers don't seem to persist in the UI between launches. It persists as far as connections though.

daveycrockett:
o  New flavor of stackoverflow approach for number-only JSpinners
o  JSpinners don't fire focus events, so use ChangeListener instead
o  Actually ran ./gradlew test check this time (sorry)


yanokwa:
This is looking good to me! I've verified that I can type arbitrary numbers and increment/decrement. Thanks for the quick fix!

PR #136, opened 2017-06-20
Add caching to prevent form processing on first run

### LINKED ISSUE #82: Do not load forms in JavaRosa at launch
On launch, ODK Briefcase seems to load each local form in JavaRosa. This can be a slow process and it's not clear to me why it is needed because all you need to load in UI is the form name.

nribeka:
Closes #82 

* Added caching
* Make some classes serializable
* Make some fields transient in the javarosa parser (this one probably gonna need testing the most), since transient fields won't get serialized, meaning it won't get deserialized when cache is read from the local file.


src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
```diff
@@ -311,10 +312,10 @@ private static synchronized final XFormParserWithBindEnhancements parseFormDefin
   /**
    * The ODK Id that uniquely identifies the form
    */
-  protected final FormDef rootJavaRosaFormDef;
+  protected transient final FormDef rootJavaRosaFormDef;
```
yanokwa: How did you decide which member variables are transient?

src/org/opendatakit/briefcase/util/FileSystemUtils.java
```diff
@@ -74,11 +78,46 @@
...
+  static Map pathToHashMap;
+  static Map pathToDefinitionMap;
+
   public static final String getMountPoint() {
     return System.getProperty("os.name").startsWith("Win") ? File.separator + ".." : (System
         .getProperty("os.name").startsWith("Mac") ? "/Volumes/" : "/mnt/");
   }
 
+  public static void saveFormDefinitionCache() {
+    if (!cacheFile.exists() || cacheFile.canWrite()) {
```
yanokwa: Are you sure this should be an OR? I'd expect AND.

src/org/opendatakit/briefcase/util/FileSystemUtils.java
```diff
@@ -74,11 +78,46 @@
   static final String ENCRYPTED_FILE_EXTENSION = ".enc";
   static final String MISSING_FILE_EXTENSION = ".missing";
 
+  static final File cacheFile = new File(getFormsFolder().getParent(), "cache.ser");
```
yanokwa: We should probably name this something clearer? Maybe `Form Definitions.cache` or `forms.cache`. Whatcha think @lognaturel?

yanokwa:
@daveycrockett Can you try and review this PR?

@nribeka What kind of testing do you think we need to validate that the transient fields work? Is an issue of big forms? Or forms that have every type of control?

yanokwa:
@nribeka I can confirm that the application loads faster on the second time, so great job there! 

There seems to be a bug where on the second launch of the application, exports fail with 
```
java.lang.NullPointerException
        at org.opendatakit.briefcase.util.ExportToCsv.populateRepeatGroupsIntoFileMap(ExportToCsv.java:699)
        at org.opendatakit.briefcase.util.ExportToCsv.processFormDefinition(ExportToCsv.java:772)
        at org.opendatakit.briefcase.util.ExportToCsv.doAction(ExportToCsv.java:127)
        at org.opendatakit.briefcase.util.ExportAction$TransformFormRunnable.run(ExportAction.java:58)
        at java.lang.Thread.run(Thread.java:748)
```
Can you describe briefly how you tested the functionality?

nribeka:
@yanokwa I added reparsing of the form definition. I only test with pull, push and export to csv on some simple form that you sent to me. Can you test more on this?

src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
nribeka: I make all all non-serializable class transient. I think this is why you're getting the NPE in your test because it reference some elements that are not restored from the deserialization process.

src/org/opendatakit/briefcase/util/FileSystemUtils.java
nribeka: This should create the definition cache file if it's not exists, or if it's exists then only try to write to it if it's writable. I think the writable part is not needed?

src/org/opendatakit/briefcase/util/FileSystemUtils.java
jpknox: The .ser filetype is reserved for the output of serialization from Java apps. I personally like "cache.ser" but I don't mind if it's renamed.

jpknox: This line could be made a little clearer with the following amendment.
`static final File cacheFile = new File(getBriefcaseFolder(), "cache.ser");`

nribeka: Updated to the `getBriefcaseFolder()` call.

src/org/opendatakit/briefcase/util/FileSystemUtils.java
```diff
@@ -180,7 +219,7 @@ public static final boolean isODKInstancesParentFolder(File pathname) {
     return foi.exists() && foi.isDirectory();
   }
 
-
+  @SuppressWarnings("unchecked")
```
daveycrockett: Is it possible to parameterize the generics pathToHashMap and pathToDefinitionMap?  Could remove this warning suppression if so.

src/org/opendatakit/briefcase/util/FileSystemUtils.java
```diff
@@ -74,11 +78,46 @@
...
+      try (ObjectOutputStream objectOutputStream = new ObjectOutputStream(new FileOutputStream(cacheFile))) {
+        objectOutputStream.writeObject(pathToHashMap);
+        objectOutputStream.writeObject(pathToDefinitionMap);
+      } catch (IOException e) {
+        e.printStackTrace();
+      }
+    }
+  }
+
+  public static void loadFormDefinitionCache() {
```
daveycrockett: Need to add warning suppression to this method (whether Maps are parameterized or not).

src/org/opendatakit/briefcase/util/FileSystemUtils.java
```diff
@@ -74,11 +78,46 @@
   static final String ENCRYPTED_FILE_EXTENSION = ".enc";
   static final String MISSING_FILE_EXTENSION = ".missing";
 
+  static final File cacheFile = new File(getBriefcaseFolder(), "cache.ser");
+
+  static Map pathToHashMap;
```
daveycrockett: Small, nit-picking preference to calling this something other than ___HashMap (just to avoid any potential confusion).  pathToMd5Map perhaps?

src/org/opendatakit/briefcase/model/BriefcaseFormDefinition.java
```diff
@@ -365,6 +366,15 @@ public boolean isFileEncryptedForm() {
   }
 
   public TreeElement getSubmissionElement() {
+    TreeElement treeElement = formDefn.getSubmissionElement();
```
daveycrockett: Just looking through the code (which I admittedly don't fully understand), it looks like BriefcaseFormDefinition.getSubmissionKey(String uri) may also need this lazy-load treatment...its value depends on formDefn.getSubmissionKey(uri) which depends on submissionElement.getName(); submissionElement is transient.

nribeka: I changed the call to use getSubmissionElement() instead of calling directly from submissionElement.

src/org/opendatakit/briefcase/util/FileSystemUtils.java
nribeka: I make it generic and I moved the supress warning to the loading from the file process.

src/org/opendatakit/briefcase/util/FileSystemUtils.java
```diff
@@ -190,7 +229,19 @@ public static final boolean isODKInstancesParentFolder(File pathname) {
         if (f.isDirectory()) {
           try {
             File formFile = new File(f, f.getName() + ".xml");
-            formsList.add(new BriefcaseFormDefinition(f, formFile));
+            String formFileHash = getMd5Hash(formFile);
+            String existingFormFileHash = String.valueOf(pathToHashMap.get(formFile.getAbsolutePath()));
```
dcbriccetti: I don’t see the “value of” String.valueOf here. (Get it? Value of?)

lognaturel:
I'm afraid I lost track of the status here. 😖 @nribeka what did you think of the idea to move the cache code to its own class? Don't like it? Like it but haven't had time to look at it yet? ...?

src/org/opendatakit/briefcase/util/CacheUtils.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.util.HashMap;
+import java.util.Map;
+
+public class CacheUtils {
```
dcbriccetti: Why `CacheUtils` with static methods and not an instantiable `FormCache` or something? Not critical in my view, I just tend to think there are too many “utils” in the world.

nribeka: I think it need to be either singleton or a static utils since it will be shared between classes, right? Yeah, too many utils in here 🕺

dcbriccetti: Did you consider instantiating it once, and passing a reference to where it is needed?

nribeka: I think that will require quiet a bit of changes in method and constructor signature.

We're using this class in the static method `FileSystemUtils.getBriefcaseFormList()`, so we can't just pass it to the `FileSystemUtils`'s constructor. So we need to change the method signature for `getBriefcaseFormList`.

The `getBriefcaseFormList` method is called in 3 different classes, so we need to change the constructor for those 2 of those classes (`PushTransferPanel` and `ExportPanel`) and pass it from the main class `MainBriefcaseWindow`.

What do you think @dcbriccetti?

dcbriccetti: I don’t want to hold you up. I could come along afterward and refactor. But how about adding a static `setFormCache` method to `FileSystemUtils`, and calling it when you instantiate `FormCache`?

dcbriccetti:
I’m rebasing this on master and Yaw will merge.

yanokwa:
Closing in favor of https://github.com/opendatakit/briefcase/pull/190

Issue #94, opened 2017-04-07
Fix style throughout the codebase


yanokwa:
The code base is currently not very consistent when it comes to code style. We already have a good [checkstyle file](https://github.com/opendatakit/briefcase/blob/master/config/checkstyle/checkstyle.xml). It can be run with `gradle check` and eventually added to continuous integration. I also like to use the [checkstyle IDEA plugin](https://plugins.jetbrains.com/plugin/1065-checkstyle-idea) in IntelliJ.

To make sure this is done in an incremental and easy to test manner, contributors should **pick ONE check to uncomment**, update this issue so others know you've taken it, then review the errors, fix them committing changes for each file and then issue a pull request.

shivam-tripathi:
Taking `<module name="CommentsIndentation"/>`

shivam-tripathi:
Comment indentation.

rclakmal:
Empty catch clauses

rclakmal:
Single line java docs

macalac:
Hey, just came across this project via up-for-grabs
Jumping on `WhitespaceAround`

rockydcoder:
Indentation

zstojanovic:
I'd like to help. OverloadMethodsDeclarationOrder.

zstojanovic:
Also, I see that there is a number of Checkstyle rules which could be uncommented without issues - the code is already conforming to them.

@yanokwa - would it be desirable to uncomment all such rules? Should I prepare a PR with such change?

lognaturel:
@zstojanovic Yes, that'd be a very nice PR to submit so we are sure those issues aren't introduced and we don't have to think about the checks again as others are enabled. Thanks!

icemc:
Multiple variable declarations

icemc:
Array type style

icemc:
Missing switch default

mayank8318:
Add an Intellij Settings file in the repository so that anyone can just import the settings and start coding :)
@yanokwa @ggalmazor

ggalmazor:
> Add an Intellij Settings file

I think that could make an easier transition for first-time contributors :)

yanokwa:
@ggalmazor I hesitate here because I'm not sure how we do this incrementally. That is, seems like it'd be ideal to have an IntelliJ settings file matches the checkstyle that CI evaluates the code against.

PR #139, opened 2017-07-07
Impl of date ranges for briefcase export

daveycrockett:
- Modified underlying exportAction api
- Barebones UI impl (not very i18n-friendly, no start/end labels, no calendar picker)

I'm looking for a code and U/X review on the basic impl of date ranges for export.  Couple things to point out:

1. I couldn't find a way to get a localized date format string (i.e., imperial time for US MM/dd/yyyy versus normal sane people time dd/MM/yyyy for everywhere else), so I opted for yyyy-MM-dd which may or may not be acceptable.
1. The easiest way to add a calendar picker would be to pull in an outside dependency, wasn't sure if anyone had opinions on whether it's worth it.
1. The layout may be a little too light-weight: a checkbox toggles, and there's no labels for "from" and "to"
1. To solve the above, there's no enforcement of which date is earlier than the other...the UI will always "do the right thing" for the ranges.  Not sure if this is intuitive or obscure U/X, and if there are strong opinions there.
1. Also added a //TODO to the code that might be worth discussing

Any other comments welcome!


lognaturel:
I think the UI is great! @daveycrockett you are awesome. Thanks for rolling up your sleeves and getting deep into the guts. 

I do see what you mean about the light-weightness though. What about matching the proxy settings UI where the checkbox goes before the section label and the from and to are indented below?

yyyy-MM-dd looks great to me. I think an outside dependency for a calendar picker would be fine since it's the type of thing that doesn't change much over time (lib abandonment doesn't matter) and should be pretty targeted/small. That component may also provide some better options for date format and internationalization.

I don't have strong feelings about using the smallest date as `from`and biggest as `to`. I wonder whether some calendar picker libs might make validating that `to` is after `from` really easy. If so, that would be my preference.

I agree with your `TODO`,  explicitly setting the state for individual components seems like a good idea.

daveycrockett:
Thank you for the feedback @lognaturel! 

- I'll mess around with checkbox location and see what looks good.
- I'm playing around with net.sourceforge.jdatepicker now, we'll see what sort of features it enables...
- I also don't have strong feelings about range validation...so I'll defer to your preference! :D

daveycrockett:
Okay, JDatePicker and validation added!  As I said in the second commit message, we should discuss any potential risks we see with adopting a dependency that has open issues and no activity since 2015 :-/ Unfortunately this was the best Java-Swing implementation I could find (but alternative suggestions are welcome!)

I noticed two issues during this impl that I wanted to raise as potentially separate issues rather than muddy this PR (if there's agreement, I'll go ahead and file them):

- ExportPanel.enabledExportButton(boolean state) is only ever called with a value of 'true,' would like to remove the parameter and branch
- Already mentioned in the //TODO I added, ExportPanel.setEnabled(boolean enabled) is brittle, and I in fact found a forth small bug I went ahead and fixed, where "Export Details..." is enabled on initial pane load.  The full fix is something I'd like to put in a separate issue.

yanokwa:
I'm liking how this is turning out, Dave! I tried it out and it did what you said it'd do 😀

I think we can do better on date picker. https://github.com/LGoodDatePicker/LGoodDatePicker is actively updated, looks good visually (at least on OS X), has internationalization and lots of other goodies. Did you consider it?

We probably should include some language in the UI about whether these dates are inclusive or exclusive. In the CLI of Briefcase, start date is inclusive and end date is exclusive and so we could do that. But, Clint has a nice UI for this in Build that might be good to replicate if you wanted to go above and beyond.

screen shot 2017-07-19 at 15 40 08

As far as the rest of the UI, I like the check mark to enable or disable range. My gut says that date range should go after type or maybe after directory because it's likely to be used more than the private key, but I don't feel strongly about that.

What I do feel strongly about is alignment! I think From/To are top aligned and to match the rest of the elements on that tab, they should be center aligned.

yanokwa:
One more idea to consider. It might be sensible to match the text in the UI with the CLI. So maybe...
* Date Range -> Export filter
* From -> Start date (inclusive)
* End -> End date (exclusive)

daveycrockett:
Thanks for the feedback @yanokwa!  I had not stumbled upon LGoodDatePicker, and from what little I've glanced at I like the API much better.  Only downside:

"The backport is distributed as jar files. (The backport it is not distributed on Maven Central)."

We'd need to add LGoodDatePicker-Backport-release-10.3.1 to the briefcase repo to support as low as Java 7.  If I get a :+1: or two that this is an acceptable approach, I'll go ahead and move forward!

lognaturel:
Using the backport sounds fine to me and the library definitely seems better.

Kind of a separate conversation but maybe we can start it here. Java 8 is about 3 years old now, could we consider targeting it? If there's an obvious reason why we shouldn't or can't without significant work then we can end the conversation here but seems worth exploring, perhaps on the dev category of the forum if there needs to be more feedback to make the decision.

Looping in @dcbriccetti who has mostly been focusing on JavaRosa recently but said on today's dev call that he'd be interested in getting involved here as well! 🎉

daveycrockett:
Thanks for all the great feedback!  Latest changes/thoughts:

- I changed the labels to reflect Yaw's feedback, happy with where that's at
- I held off on adding "inclusive" and "exclusive" to keep this PR incremental, I'll open a separate issue to add it
- I think I got all the alignment correct...all the existing form controls used BASELINE; all my added ones use BASELINE except the row with the checkbox, which really only looks right with CENTER alignment.  Please spot-check, clearly my eye for aesthetics is not the best! :)
- Backport LGoodDatePicker now part of the repo, seems to behave just fine
- I'll open a separate issue for upgrading to Java 8, we can discuss there? (I agree it's a good idea!)

src/org/opendatakit/briefcase/ui/ExportPanel.java
```diff
@@ -358,6 +389,16 @@ public ExportPanel(TerminationFuture terminationFuture) {
     btnPemFileChooseButton = new JButton("Choose...");
     btnPemFileChooseButton.addActionListener(new PEMFileActionListener());
 
+    JLabel lblDateRange = new JLabel("Specific Date Range:");
+    checkDates = new JCheckBox("", false);
```
dcbriccetti: I should be able to click here to toggle the checkbox. You cheated by separating the checkbox from the text that goes with it. I assume it’s to line things up the way you like. Try another approach, or don’t require the text to line up?
jcheckbox

src/org/opendatakit/briefcase/ui/ExportPanel.java
```diff
@@ -333,6 +341,12 @@ public void enableExportButton(boolean state) {
         btnExport.setEnabled(false);
         return;
       }
+
+      if ( checkDates.isSelected() && (pickStartDate.getDate() == null || pickEndDate.getDate() == null)) {
```
dcbriccetti: I think we aren’t putting a space after “if (”.

src/org/opendatakit/briefcase/ui/ExportPanel.java
```diff
@@ -88,8 +89,8 @@
   private DetailButton btnDetails;
   private JButton btnExport;
   private JButton btnCancel;
-  private JDatePickerImpl pickStartDate;
-  private JDatePickerImpl pickEndDate;
+  private DatePicker pickStartDate;
```
dcbriccetti: Making these final lets the reader know these are not instantiated multiple times. Reduces cognitive load.
```
  private final DatePicker pickStartDate;
  private final DatePicker pickEndDate;
  private final JCheckBox checkDates;
```

src/org/opendatakit/briefcase/ui/ExportPanel.java
```diff
@@ -194,6 +205,20 @@ public void actionPerformed(ActionEvent e) {
     }
   }
 
+  class DateRangeActionListener implements ActionListener, DateChangeListener {
```
dcbriccetti: I suggest breaking apart the two listeners, as detailed in comments to come.

src/org/opendatakit/briefcase/ui/ExportPanel.java
```diff
@@ -358,6 +402,21 @@ public ExportPanel(TerminationFuture terminationFuture) {
     btnPemFileChooseButton = new JButton("Choose...");
     btnPemFileChooseButton.addActionListener(new PEMFileActionListener());
 
+    JLabel lblDateRange = new JLabel("Submission Date Filter");
+    JLabel lblDateFrom = new JLabel("Start Date (inclusive):");
+    JLabel lblDateTo = new JLabel("End Date (exclusive):");
+    DateRangeActionListener dateRangeListener = new DateRangeActionListener();
+    checkDates = new JCheckBox("", false);
+    checkDates.addActionListener(dateRangeListener);
```
dcbriccetti: I’d use an ItemListener (not ActionListener), and put it inline, like this:
```
    checkDates.addItemListener(new ItemListener() {
        @Override
        public void itemStateChanged(ItemEvent e) {
            pickStartDate.setEnabled(checkDates.isSelected());
            pickEndDate.setEnabled(checkDates.isSelected());
            ExportPanel.this.setEnabled(true);
        }
    });
```

src/org/opendatakit/briefcase/ui/ExportPanel.java
```diff
@@ -358,6 +402,21 @@ public ExportPanel(TerminationFuture terminationFuture) {
...
+
+    pickStartDate = new DatePicker();
```
dcbriccetti: I’d make the DateChangeListener like this:
```
    final DateChangeListener dateChangeListener = new DateChangeListener() {
        @Override
        public void dateChanged(DateChangeEvent dateChangeEvent) {
            ExportPanel.this.setEnabled(true);
        }
    };
    pickStartDate = new DatePicker();
    pickStartDate.setEnabled(false);

    pickStartDate.addDateChangeListener(dateChangeListener);

    pickEndDate = new DatePicker();
    pickEndDate.setEnabled(false);
    pickEndDate.addDateChangeListener(dateChangeListener);
```

src/org/opendatakit/briefcase/ui/ExportPanel.java
```diff
@@ -194,6 +205,20 @@ public void actionPerformed(ActionEvent e) {
     }
   }
 
+  class DateRangeActionListener implements ActionListener, DateChangeListener {
+    @Override
+    public void actionPerformed(ActionEvent e) {
+      pickStartDate.setEnabled(checkDates.isSelected());
+      pickEndDate.setEnabled(checkDates.isSelected());
+      ExportPanel.this.setEnabled(true);
```
dcbriccetti: Why call setEnabled on the whole ExportPanel here?

dcbriccetti: OK, I read above where you want to address this in another change. That’s fine with me.

dcbriccetti:
@daveycrockett, I’m happy to pair with you and take care of this setEnabled business if you like. I’m free now for a couple hours, and all day tomorrow Pacific time.

src/org/opendatakit/briefcase/ui/ExportPanel.java
daveycrockett: Would this be something covered in https://github.com/opendatakit/briefcase/issues/94 all at once? I saw this convention all throughout this file, was just "when in Rome"-ing.

dcbriccetti: I can respect your "when in Rome"-ing. Leave as is, if you like.

daveycrockett:
Thanks for the feedback @dcbriccetti! I got rid of the checkbox and it cleaned everything up nicely.  Pairing on the setEnabled() refactor sounds fun, but 1) can we create a separate issue for it, and 2) can I get with you on availability in two weeks? I'm about to be on vacation for the next week and a half (will also hit you up on the Slacks).

One final observation: I'm not sure where in the order of form elements the date fields should go, but below "Pem file" doesn't seem right.  Anyone have any strong opinions here?

yanokwa:
I forgot to squash 😭

Issue #140, opened 2017-07-09
Recorded instance check performance degrades proportionally with total instances


batkinson:
Seen while working on a prototype for #138, based on 1.6.10.

With the improvement in download performance of the #138 prototype, I noticed instance fetch performance became unexpectedly poor for forms with larger number of submissions. Also, resuming a form download after a previous batch of fetches was slow even when no instance downloads occur.

The reason for this behavior is the implementation of the form instance tracking. Specifically, `DatabaseUtils.hasRecordedInstance(...)` gets proportionally slower with the number of form instances tracked for the form. Since this method is called for each form instance, the linear increase in execution time of each call results in an exponential increase in the time needed to fetch all instances. The problem is *very noticeable* once a form's `recorded_instance` table approaches the 50,000 record range.

The acute problem is the sql query `SELECT directory FROM recorded_instance WHERE instanceId = ?` requires a full table scan. The solution: index the `instanceId` column. However, smallsql does not appear to support indexes.

To confirm the fix: I swapped instance tracking to use hsql with an indexed `instanceId` column. Instance check performance improved considerably even for small numbers of instances, and the performance level did not degrade proportionally with the number of tracked instances.

PR #142, opened 2017-07-09
Briefcase-138

batkinson:
Improved fetch performance over high-latency connections and related changes.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/MessageStrings.java
+++ b/src/org/opendatakit/briefcase/ui/MessageStrings.java
@@ -80,6 +80,6 @@ public class MessageStrings {
   public static final String PROXY_SET_ADVICE =
       "If you are behind a proxy, try setting up your proxy details through 'Settings' tab.";
 
-  public static final String PARALLEL_DOWNLOADS = "Use Parallel Downloads";
+  public static final String PARALLEL_PULLS = "Pull submissions in parallel (experimental)";
 
 }
--- a/src/org/opendatakit/briefcase/ui/MessageStrings.java
+++ b/src/org/opendatakit/briefcase/ui/MessageStrings.java
@@ -80,4 +80,6 @@ public class MessageStrings {
   public static final String PROXY_SET_ADVICE =
       "If you are behind a proxy, try setting up your proxy details through 'Settings' tab.";
 
+  public static final String PARALLEL_DOWNLOADS = "Use Parallel Downloads";
+
 }
--- a/src/org/opendatakit/briefcase/util/ServerFetcher.java
+++ b/src/org/opendatakit/briefcase/util/ServerFetcher.java
@@ -270,6 +270,7 @@ public class ServerFetcher {
     public Thread newThread(Runnable r) {
       Thread t = new Thread(r, namePrefix + threadNumber.getAndIncrement());
       t.setPriority(Thread.MIN_PRIORITY);
+      t.setDaemon(true);
       return t;
     }
   }
--- a/src/org/opendatakit/briefcase/util/ServerFetcher.java
+++ b/src/org/opendatakit/briefcase/util/ServerFetcher.java
@@ -241,87 +241,75 @@ public class ServerFetcher {
     return allSuccessful;
   }
 
-  public static class SubmissionDownloadChunk {
+  public static class SubmissionChunk {
     final String websafeCursorString;
     final List<String> uriList;
 
-    public SubmissionDownloadChunk(List<String> uriList, String websafeCursorString) {
+    public SubmissionChunk(List<String> uriList, String websafeCursorString) {
       this.uriList = uriList;
       this.websafeCursorString = websafeCursorString;
     }
   };
 
   private boolean downloadAllSubmissionsForForm(File formInstancesDir, DatabaseUtils formDatabase, BriefcaseFormDefinition lfd,
-      FormStatus fs) {
-    boolean allSuccessful = true;
-
-    RemoteFormDefinition fd = (RemoteFormDefinition) fs.getFormDefinition();
-
+                                                FormStatus fs) {
     int count = 1;
-    String baseUrl = serverInfo.getUrl() + "/view/submissionList";
+    boolean allSuccessful = true;
+    RemoteFormDefinition fd = (RemoteFormDefinition) fs.getFormDefinition();
+    ExecutorService execSvc = Executors.newFixedThreadPool(MAX_CONNECTIONS_PER_ROUTE);
+    CompletionService<SubmissionChunk> chunkCompleter = new ExecutorCompletionService(execSvc);
+    CompletionService<String> submissionCompleter = new ExecutorCompletionService(execSvc);
 
-    String oldWebsafeCursorString = "not-empty";
-    String websafeCursorString = "";
-    for (; !oldWebsafeCursorString.equals(websafeCursorString);) {
-      if ( isCancelled() ) {
-        fs.setStatusString("aborting fetching submissions...", true);
+    String oldWebsafeCursorString, websafeCursorString = "";
+
+    chunkCompleter.submit(new SubmissionChunkDownload(fs, fd.getFormId(), websafeCursorString));
+
+    boolean cursorFinished;
+
+    try {
+      do {
+        if (isCancelled()) {
+          fs.setStatusString("aborting fetching submission chunks...", true);
+          EventBus.publish(new FormStatusEvent(fs));
+          return false;
+        }
+
+        fs.setStatusString("processing submission chunk...", true);
         EventBus.publish(new FormStatusEvent(fs));
-        return false;
-      }
 
-      fs.setStatusString("retrieving next chunk of instances from server...", true);
-      EventBus.publish(new FormStatusEvent(fs));
+        oldWebsafeCursorString = websafeCursorString; // remember what we had...
+        SubmissionChunk chunk;
+        try {
+          chunk = chunkCompleter.take().get();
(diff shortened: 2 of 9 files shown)


lognaturel:
@batkinson Have you tried this with a real Aggregate server? On AppEngine?

I'm unfortunately not very well versed in this part of the system yet and have not yet understood Mitch's concern well enough to be able to gauge whether this approach avoids it. :/ Will do my best to get up to speed and provide useful feedback! I think it would be good to get several pairs of eyes on this one.

batkinson:
Hi @lognaturel. This was tested on a production deployment of Aggregate running Tomcat on AWS. I didn't have a working AppEngine deployment to use. Mitch's concern was that low-bandwidth would have to be shared amongst multiple downloads if you go concurrent. That would actually *increase* the transmission times for those files so it could make those downloads fail where they otherwise wouldn't.

My observation here is based on running briefcase on an *actual* deployment. The briefcase download protocol is not well-designed for what it attempts to do. The http request-response style of communication for single submission file download means that latency dominates, even with high-bandwidth (like a satellite connection).

batkinson:
Oh, one thing I forgot. I fully anticipated that the concurrency level could be configurable, which should give users the ability to configure for their environment if needed. However, my guess would be that users would probably go the *other way* and want more concurrency. I balanced this to minimize lock contention on the shared object monitors.

lognaturel:
That's also roughly how I was understanding what Mitch said. And this is specifically a problem in the context of App Engine because of the 60-second transmission cap. However, it's not quite clear to me how likely that 60 sec limit is to be surpassed. With text-only form submissions, I'm guessing it's not common even with very low bandwidth. It's when media comes into play that things get problematic and the limit could be reached constantly. I'm not totally clear on whether that limit is inevitable, though -- [this Google reference](https://cloud.google.com/appengine/articles/deadlineexceedederrors) suggests that the server does have a chance to catch the exception though perhaps under those conditions there's not enough time. I don't know whether Aggregate tries this or not.

I think you're right that having the level of concurrency be configurable is the ideal. Mitch suggested "any concurrent-request logic needs to adjust its level of concurrency to account for the available bandwidth." I think the size of the elements being pulled would also need to be taken into account. I believe that one could get dramatically different results on the same connection and Aggregate instance depending on whether the submissions being pulled have attached media or not, for example. Does that sound correct? Do you have any media in your submissions, @batkinson?

How about adding a setting for concurrent connections that is off by default for now? I'm imagining a top-level checkbox that is something like "Try multiple connections at once" and a sub-setting that is a dropdown of plausibly reasonable choices. Then we could generate some really good help text on how to anticipate what the right value will be for each form that needs to be pulled. As more experiments happen under different conditions, the default could be changed.

If that general approach appeals to you @batkinson, same thing as with the DB change, someone else can do that piece of it on a branch that builds off this one since I know the setting is of less value to you.

batkinson:
> With text-only form submissions, I'm guessing it's not common even with very low bandwidth. It's when media comes into play that things get problematic and the limit could be reached constantly.
Yes, I suspect the parallel downloads would only really influence things greatly when downloading multiple _large_ media attachments. One way that can could be handled is scheduling media attachments on a single thread (and http connection). However, for small media files, you're likely still better off unless you're saturating your network connection. From my testing, I couldn't get submission downloads even close to saturating my network link.

batkinson:
I followed the advice from @lognaturel and added a setting to enable parallel downloads that is disabled by default. I also rebased this to eliminate a merge conflict after #140 was merged into master. Also, I fixed checkstyle warnings caused by IntelliJ collapsing imports from the same package into a wildcard import.

mitchellsundt:
Things to keep in mind:

1.  Push of data to the server should be serialized.  Within ODK Aggregate, there is a **global mutex** ("TaskLock" across all server instances, mediated by the datastore layer) in the server when inserting submissions. Having multiple push requests occurring simultaneously will cause them to block on the mutex, chewing up their 60-second request limit, as they get processed in single file no matter how many server instances are spun up. 

2. That 60-second request limit can be very commonly exceeded over low-bandwidth connections, and even text-only submissions can be impacted on satellite connections. That is why ODK Collect splits submissions into multiple 10MB submission requests. The timer starts upon receipt of the first byte, so a slow connection can account for a sizeable portion of those 60 seconds. Ditto for sending a response. The processing on the server is generally negligible in relation to the transmission times.

3. Pull of data can be overlapped. However, for submissions with many media attachments, keep in mind that ODK Aggregate holds all media attachments in memory, then copies them into the HttpEntity, so this can consume a fair amount of memory (everything is held in memory at least twice). *The same memory consumption applies on pushes / submission of forms.*  I.e., you probably want to manage how many media attachment download requests you have in play at the same time.

4. Spinning up of copies of the frontend will incur faster quota usage on AppEngine. For that reason, the Aggregate configuration [here](https://github.com/opendatakit/aggregate/blob/master/eclipse-default/WebContent/WEB-INF/appengine-web.xml) specifies a 14-second queuing time threshold before a new instance is spun up.  Only if at least one request is queued for longer than 14 seconds will a new instance be spun up, and then that new instance will take about 30 seconds to become live. Leaving a 15-second processing interval. This is why ODK Collect tried twice before failing a submit.

The two worst of these limitations, the global mutex and the in-memory copies/full-packet-assembly, are a result of implementing on top of AppEngine and its Datastore. 

A server that used database transactions and that used streaming servlet 3.0 functionality would have less trouble with concurrent requests.

src/org/opendatakit/briefcase/util/WebUtils.java
```diff
@@ -307,39 +315,37 @@ public static final HttpPost createOpenRosaHttpPost(URI uri) {
...
+
+		HttpClientBuilder clientBuilder = HttpClientBuilder.create()
+				.setConnectionManager(connectionManager)
+				.setConnectionManagerShared(true)
+				.setDefaultSocketConfig(socketConfig)
+				.setDefaultRequestConfig(requestConfig);
+
+		HttpHost proxy = BriefcasePreferences.getBriefCaseProxyConnection();
+		if (proxy != null) {
+			clientBuilder.setRoutePlanner(new DefaultProxyRoutePlanner(proxy)).build();
```
lognaturel: I think this call on `build` is redundant. Not sure whether it matters.

src/org/opendatakit/briefcase/util/ServerFetcher.java
```diff
@@ -238,88 +250,175 @@ public boolean downloadFormAndSubmissionFiles(List<FormStatus> formsToTransfer)
...
+        for (String uri : chunk.uriList) {
+          if (isCancelled()) {
+            fs.setStatusString("aborting requesting submissions...", true);
+            EventBus.publish(new FormStatusEvent(fs));
+            return false;
+          }
+          submissionCompleter.submit(new SubmissionDownload(formInstancesDir, formDatabase, lfd, fs, uri));
+        }
+
+        for (int i = 0; i < chunk.uriList.size(); i++) {
```
lognaturel: I think this can be replaced with a for-each loop. It confused me since there's the same loop above so I was trying to figure out a reason for it being different. I'm apparently easily confused!

src/org/opendatakit/briefcase/ui/MessageStrings.java
```diff
@@ -80,4 +80,6 @@
   public static final String PROXY_SET_ADVICE =
       "If you are behind a proxy, try setting up your proxy details through 'Settings' tab.";
 
+  public static final String PARALLEL_DOWNLOADS = "Use Parallel Downloads";
```
lognaturel: I think it would be good to indicate that this is for pulls only.

src/org/opendatakit/briefcase/util/ServerFetcher.java
lognaturel: Or perhaps you did it this way to make it clear that no elements are actually used and that it's purely a counting loop? That actually makes sense.

lognaturel:
Really appreciate your insights, @mitchellsundt, those are so helpful. From what you say, I think this level of parallelism will mostly be helpful for those who don't have media attachments and/or don't run on AppEngine. The point about in-memory copies is particularly interesting and it seems that might be a good next step for refining this implementation. 

When we release this, we really need to be thoughtful about how we communicate to users when parallelism might help them and when it could hurt them.

src/org/opendatakit/briefcase/util/ServerFetcher.java
batkinson: Yes, I didn't use the for-each here because the second loop doesn't actually use the items in the list. It only needs to match up with the first loop, since each submitted job in the first loop should have a matching call to `take()` in the second. I can add a comment to clarify and prevent the reader from having to dive directly into the loop semantics.

src/org/opendatakit/briefcase/util/WebUtils.java
batkinson: Yes, you're right. At the very least, it's confusing. I just missed this when converting the original code. I'll remove it.

src/org/opendatakit/briefcase/ui/MessageStrings.java
batkinson: Ok, I thought download implied that, but it's the only place where the word download exists in the UI. I will reword this to make it more clear.

yanokwa:
Nice work, @batkinson! I've confirmed that this is a lot faster on a small form with no attachments on App Engine and a local Tomcat server. I've also confirmed that it works via an HTTP proxy.

Marking parallel downloads as an experimental feature is a good way to communicate to users that the feature might not work for them, so great job there too!

Issue #141, opened 2017-07-09
App responsiveness degrades proportionally with form history length during pull


batkinson:
Seen while working on a prototype for #138, based on 1.6.10.

I noticed after fixing #140 that while instance checks were no longer slow, the application still seemed to slow down as the duration of the sync continued. However, it appeared not to affect the progress of instance fetching and processing and instead only affected the responsiveness of the user interface. It was also correlated with increasingly high CPU usage.

After taking a look with a profiler, I discovered that the reason for the slow down was extremely high allocator activity. Specifically, allocation/deallocation of objects in the Swing text package. The activity coming exclusively from the AWT event thread.

I was able to track this back to `ScrollingStatusListDialog.onEvent(...)`. The problem is that the contents of the JEditorPane are replaced with the entire status history. There are multiple problems with the implementation:

- For a single status message update, the amount of work required to update the dialog is proportional to the entire status history, not just the single message - much like #140, the performance slows exponentially
- Updating the contents of the dialog happens regardless of whether the dialog is actually showing
- The `JEditorPane` is updated using the `setText(...)` method, which updates the existing internal model which as the documentation of the method states: 
> Leaving the existing model in place means that the old view will be torn down, and a new view created, where replacing the document would avoid the tear down of the old view.

Issue #138, opened 2017-07-04
Better utilize concurrency when pulling form submissions


batkinson:
This is [an issue that was previously raised on the Google Group with submission download performance](https://groups.google.com/d/msg/opendatakit/kzuRmSSPkfg/N6fgFDodyHEJ). However, I stumbled upon this independently and after doing some investigation on it myself, would ask that we reopen the discussion.

Specifically, I believe that adding support for concurrent submission downloads would improve  performance considerably in environment with _high latency_, regardless of the available bandwidth. Mitch's assessment did not talk about the design of the download protocol itself. Over high-latency network connections, the time waiting for previous requests to complete can easily dominate the time necessary to complete a pull.

My basic understanding of the protocol, for each form being pulled:
- Fetch the form definition, the media manifest, and files listed in the manifest
- Fetch a chunk of 100 (max) submission ids for the form, starting from the last cursor value
- One-by-one fetch each submission, including the media files

Consider an example pull:

- A single form definition with 10 media files
- A significant number of submissions, let's say 1000
- Each submission has an average of 1.5 attached media files
- The client's round-trip latency averages 350 milliseconds
- Each http request is serviced and sent instantaneously, eliminating bandwidth from the discussion
- Assume TCP connection time is near-constant with regards to the number of requests using HTTP 1.1 persistent connections

Because each request is performed serially, the time it will take to complete the pull is the round-trip time multiplied by the number of http requests:

- 1 form definition
- 1 form manifest
- 10 media files
- 10 id chunks containing 100 submission ids each (covering all 1000 submissions)
- 1000 form submissions
- 1500 submission media files

That works out to be 2,522 round trips. Multiplying that by our 350ms round trip time, and even with infinite processing speed on the server and infinite bandwidth, the pull takes: 2,522 requests * 350ms = 882,700 ms (approximately 14.7 minutes).

My reason for investigating this was a pull of a single form definition with over 200,000 submissions. When I did the basic arithmetic assuming the same infinite processing speed and bandwidth, I found that the pull (ignoring media files) would take around (2,000 id chunks + 200,000 submissions) * 350ms = 70,700,00ms (approximately 19.6 hours).

By introducing concurrency, we can effectively divide the time spent waiting for round trips by the number of concurrent downloads. For example, given 4 independent threads, 19.6 hours turns into 4.9 hours - a significant improvement.

yanokwa:
[...]

PR #143, opened 2017-07-09
Fixed #140: Recorded instance check performance degrades proportional…

### LINKED ISSUE #140: Recorded instance check performance degrades proportionally with total instances

batkinson:
Significantly improves performance of instance tracking by migrating briefcase form instance db to use hsql. Necessary due to the lack of an indexing feature in smallsql.

### DIFF
--- a/src/org/opendatakit/briefcase/util/DatabaseUtils.java
+++ b/src/org/opendatakit/briefcase/util/DatabaseUtils.java
@@ -18,15 +18,14 @@ package org.opendatakit.briefcase.util;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.opendatakit.briefcase.model.FileSystemException;
 
 import java.io.File;
-import java.sql.Connection;
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Statement;
+import java.sql.*;
 import java.util.Set;
 import java.util.TreeSet;
+
+import static org.opendatakit.briefcase.util.FileSystemUtils.getFormDatabaseUrl;
 
 /**
  * This class abstracts all the functionality of the instance-tracking
@@ -57,7 +56,7 @@ public class DatabaseUtils {
   public DatabaseUtils(Connection connection) {
     this.connection = connection;
   }
-  
+
   public void close() throws SQLException {
     if ( getRecordedInstanceQuery != null ) {
       try {
@@ -201,6 +200,14 @@ public class DatabaseUtils {
     } catch ( SQLException e ) {
       log.error("failure while pruning instance registry", e);
     }
+  }
+
+  public static DatabaseUtils newInstance(File formDirectory) throws FileSystemException, SQLException {
+    return new DatabaseUtils(getConnection(getFormDatabaseUrl(formDirectory)));
+  }
+
+  static Connection getConnection(String jdbcUrl) throws SQLException {
+    return DriverManager.getConnection(jdbcUrl);
   }
 
   public static void migrateData(Connection old_c, Connection new_c) throws SQLException {
--- a/src/org/opendatakit/briefcase/util/DatabaseUtils.java
+++ b/src/org/opendatakit/briefcase/util/DatabaseUtils.java
@@ -80,21 +80,19 @@ public class DatabaseUtils {
     if (hasRecordedInstanceTable) return;
 
     try (Statement stmt = connection.createStatement()) {
-
       try {
         stmt.execute(ASSERT_SQL);
+        if (log.isDebugEnabled()) {
+          stmt.execute(SELECT_ALL_SQL);
+          ResultSet rs = stmt.getResultSet();
+          while (rs.next()) {
+            log.debug("recorded: " + rs.getString(1) + " @dir=" + rs.getString(2));
+          }
+        }
       } catch (SQLException e) {
         log.debug("assertion failed, attempting to create instance table");
         createRecordedInstanceTable(connection);
         hasRecordedInstanceTable = true;
-      }
-
-      if (log.isDebugEnabled()) {
-        stmt.execute(SELECT_ALL_SQL);
-        ResultSet rs = stmt.getResultSet();
-        while (rs.next()) {
-          log.debug("recorded: " + rs.getString(1) + " @dir=" + rs.getString(2));
-        }
       }
     }
   }
--- a/src/org/opendatakit/briefcase/util/DatabaseUtils.java
+++ b/src/org/opendatakit/briefcase/util/DatabaseUtils.java
@@ -222,14 +222,16 @@ public class DatabaseUtils {
     }
   }
 
-  public static void migrateData(Connection old_c, Connection new_c) throws SQLException {
-    createRecordedInstanceTable(new_c);
-    try (PreparedStatement selectStmt = old_c.prepareStatement(SELECT_ALL_SQL);
-         ResultSet rset = selectStmt.executeQuery();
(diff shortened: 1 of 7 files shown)


lognaturel:
This definitely seems like a really great idea! The big question for me is how to minimize the impact on existing users. If I'm understanding correctly, what would happen with the db change as written is that no downloaded submissions would be recognized after the Briefcase update because although the files would still exist, the DB records would no longer be recognized. Users would continue to have a stale `info.db` folder.

I think this might be pretty confusing and it might be better to have some kind of migration to make the change a little more seamless. What do you think, @yanokwa?

yanokwa:
Agreed with @lognaturel that it would be good to have some migration story. 

I'm thrilled with the removal of SmallSQL since it hasn't been updated since 2011, but I'm curious why HSQLDB was used instead of SQLite?

batkinson:
Too much to say in response, but of the PRs for #138, #140 and #141 this is the one I would knew would be most likely to be rejected. However, I submitted it because it's a working fix for a debilitating scale problem users have *today*.

In terms of impact, you may be over-stating it, especially compared with the problems the issue causes. I suspect users that have large data sets just abandoned the idea of using briefcase once they tried it in practice. We almost did.

@yanokwa, the hsql choice was fairly arbitrary. HSQL is popular enough, as is H2, and it's more powerful than sqlite. However, that misses the point. *Any* database that supports indexing can be used to fix the problem.

If you want, I can resubmit with sqlite, but it's unlikely I have the cycles to build a complete migration solution (also, I am not convinced it's overly important given how fast re-verification once this is fixed). However, we can't use the existing tool as-is so I figured I would offer it up.

batkinson:
@yanokwa, not sure I understand the motivation for using sqlite here. The purpose is fast lookup of the instance directory based on instanceId, which the previous implementation fails at because it doesn't use indexes. If the popularity of the db tech matters, I guess I am missing the point. Smallsql wasn't exactly popular, but it worked (at least someone thought it), right?

Additionally, I re-implemented another branch using sqlite and the size of the driver is *huge*. The briefcase jar goes from 5MB to over 11MB with sqlite. The extra size is likely due to the native code, most of which won't even be used in a normal installation. For an internal database with 1 table containing 2 files, I'm not sure I'm sold that is a better option given its purpose.

We are arguing about the color of the bike shed. As long as a suitable database is used, and instance id lookup is O(lg n) - presumably using indexing - the issue will be fixed as it is in the PR.

batkinson:
> If I'm understanding correctly, what would happen with the db change as written is that no downloaded submissions would be recognized after the Briefcase update because although the files would still exist, the DB records would no longer be recognized.

@lognaturel, yes this PR only fixes the acute issue. However, most of the reason download is painful are the three issues I just submitted PRs for. Also, those most affected are people with large data sets, which probably just don't use briefcase (over the net) right now. Though that's complete speculation. My point is that fixing these *significantly* improves the whole experience downloading, *especially* for those in resource-constrained environments.

>  Users would continue to have a stale info.db folder.

Yep, but even with a migration I probably wouldn't want to remove it. You really think people would be confused by it? How many people actually know what the existing one is for? I actually didn't until I needed to fix this.

lognaturel:
I think @yanokwa was asking about the database choice to have a record of your rationale, not to second-guess it. The jar size difference alone is plenty compelling and I believe that's why smallsql was chosen initially.

As I understand it, many people use Briefcase to do incremental backups. That is, they pull records down once every N days to build up a local mirror of their data. They expect that the download will continue where it was left off and starting the download over could be a problem depending on available bandwidth, how they pay for it, etc.

For the migration, I was imagining using a preference to keep track of whether the DB layer has been changed yet or not. If a check on load shows it hasn't, then the hsql dbs for each form could be built up with some kind of "update in progress" dialog blocking user interaction. The smallsql `info.db` folder could stay there as it's not very big. I guess it would be helpful in the case where someone goes down to an earlier version of Briefcase.

@batkinson If someone built that migration on a branch off this one, would you be willing to do a verification pass?

batkinson:
Hi @lognaturel. Ah ok. Yeah, hopefully my explanation made it clear; no big reason, almost any would do and so I used one I that I had cached locally to save the download. Given the limited use and the obscurity of smallsql I figured almost anything was better as long as it didn't make deployment more difficult. My apologies if my response seemed overly defensive.

> @batkinson If someone built that migration on a branch off this one, would you be willing to do a verification pass?

Sure. So, I actually did a quick pass at a basic working solution. I can just push it to this PR. If someone wants to improve on it, I can take a look.

batkinson:
To whomever reviews, apologies for the long list of commits, but this was a bit of a slog due to the state of the code. Anyway, I tried to keep the commits small so you can review easily. Note that there will now be a small conflict (easily resolved) if applying the scale fixes (#138, #140, and #141) together.

lognaturel:
Super thoughtful migration implementation. Thanks so much, @batkinson!

In terms of manual verification, I looked at the following:
- existing smallsql db for a particular form -- info.hsqldb folder is created, info.db is moved to info.db.bak; downloads resume where expected (and much faster than before)
- form that has never been pulled before -- info.hsqldb folder created as expected; downloads resume where expected on subsequent pulls
- Briefcase running twice, `unable to open form database: Database lock acquisition failure` is shown

src/org/opendatakit/briefcase/util/DatabaseUtils.java
```diff
@@ -95,6 +90,17 @@ private void assertRecordedInstanceTable() throws SQLException {
     }
   }
 
+  private void dumpRecordedInstanceTable() throws SQLException {
+    try (Statement stmt = connection.createStatement()) {
+      stmt.execute(SELECT_ALL_SQL);
+      try (ResultSet rs = stmt.getResultSet()) {
+        while (rs.next()) {
+          log.debug("recorded: " + rs.getString(1) + " @dir=" + rs.getString(2));
```
lognaturel: @batkinson you mentioned quickly that you had noticed the return of a slowdown and I think it was based on this logging, is that right? The issue being that `isDebugEnabled` always true?

With my setup, I can't seem to get these debug statements to show up at all although `isDebugEnabled` does evaluate to true. If I put a `log.error` I can see it, though. In other classes, info messages show up at my IntelliJ console. I'm having a tough time following how logging is intended to work. @batkinson, if you're around for a quick chat in #briefcase-code, it might be helpful to talk live for a minute and strategize on how to approach getting more sane logging both for performance and general improvements.

batkinson: >The issue being that isDebugEnabled always true?

You got it exactly. Something with the logging setup smells off.

PR #144, opened 2017-07-09
Fixed #141: App responsiveness degrades proportionally with form hist…

### LINKED ISSUE #141: App responsiveness degrades proportionally with form history length during pull

batkinson:
Improves application performance when form status history grows by bounding the history size, reducing memory allocations by using alternative `JEditorPane` API, and only updating the `JEditorPane` when it is actually showing.

### DIFF
--- a/src/org/opendatakit/briefcase/model/FormStatus.java
+++ b/src/org/opendatakit/briefcase/model/FormStatus.java
@@ -60,7 +60,6 @@ public class FormStatus {
     }
     statusHistory.append("\n");
     statusHistory.append(statusString);
-    // statusHistory.append("</p>");
     this.isSuccessful = this.isSuccessful && isSuccessful;
   }
 
--- a/src/org/opendatakit/briefcase/model/FormStatus.java
+++ b/src/org/opendatakit/briefcase/model/FormStatus.java
@@ -25,7 +25,7 @@ public class FormStatus {
   private String statusString = "";
   private final StringBuilder statusHistory = new StringBuilder();
   private boolean isSuccessful = true;
-  private int historyMaxSize = 8096;
+  private int historyMaxSize = 8192;
 
   public FormStatus(TransferType transferType, IFormDefinition form) {
     this.transferType = transferType;
--- a/src/org/opendatakit/briefcase/model/FormStatus.java
+++ b/src/org/opendatakit/briefcase/model/FormStatus.java
@@ -18,6 +18,9 @@ package org.opendatakit.briefcase.model;
 
 
 public class FormStatus {
+
+  public static final int STATUS_HISTORY_MAX_BYTES = 1024 * 1024;
+
   public enum TransferType { GATHER, UPLOAD };
   private final TransferType transferType;
   private boolean isSelected = false;
@@ -25,7 +28,6 @@ public class FormStatus {
   private String statusString = "";
   private final StringBuilder statusHistory = new StringBuilder();
   private boolean isSuccessful = true;
-  private int historyMaxSize = 8192;
 
   public FormStatus(TransferType transferType, IFormDefinition form) {
     this.transferType = transferType;
@@ -55,7 +57,7 @@ public class FormStatus {
 
   public void setStatusString(String statusString, boolean isSuccessful) {
     this.statusString = statusString;
-    if (statusHistory.length() > historyMaxSize) {
+    if (statusHistory.length() > STATUS_HISTORY_MAX_BYTES) {
       trimHistory(statusString.length());
     }
     statusHistory.append("\n");
--- a/src/org/opendatakit/briefcase/model/FormStatus.java
+++ b/src/org/opendatakit/briefcase/model/FormStatus.java
@@ -56,16 +56,20 @@ public class FormStatus {
   public void setStatusString(String statusString, boolean isSuccessful) {
     this.statusString = statusString;
     if (statusHistory.length() > historyMaxSize) {
-      statusHistory.delete(0, statusString.length() + 1);
-      int lineEnd = statusHistory.indexOf("\n");
-      if (lineEnd >= 0) {
-        statusHistory.delete(0, lineEnd+1);
-      }
+      trimHistory(statusString.length());
     }
     statusHistory.append("\n");
     statusHistory.append(statusString);
     // statusHistory.append("</p>");
     this.isSuccessful = this.isSuccessful && isSuccessful;
+  }
+
+  private void trimHistory(int len) {
+    statusHistory.delete(0, len + 1);
+    int lineEnd = statusHistory.indexOf("\n");
+    if (lineEnd >= 0) {
+      statusHistory.delete(0, lineEnd + 1);
+    }
   }
 
   public String getStatusHistory() {
--- a/src/org/opendatakit/briefcase/model/FormStatus.java
+++ b/src/org/opendatakit/briefcase/model/FormStatus.java
@@ -25,6 +25,7 @@ public class FormStatus {
   private String statusString = "";
   private final StringBuilder statusHistory = new StringBuilder();
   private boolean isSuccessful = true;
+  private int historyMaxSize = 8096;
 
   public FormStatus(TransferType transferType, IFormDefinition form) {
     this.transferType = transferType;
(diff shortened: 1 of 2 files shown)


src/org/opendatakit/briefcase/model/FormStatus.java
```diff
@@ -25,6 +25,7 @@
   private String statusString = "";
   private final StringBuilder statusHistory = new StringBuilder();
   private boolean isSuccessful = true;
+  private int historyMaxSize = 8096;
```
lognaturel: I'm not able to figure out where this limitation comes from! Is it something you figured out empirically? Could you please add a little comment?

I hesitate to make style suggestions out here because it's the Wild West but might be good to make this a constant if you don't mind! 😊

src/org/opendatakit/briefcase/model/FormStatus.java
```diff
@@ -54,6 +55,13 @@ public void clearStatusHistory() {
 
   public void setStatusString(String statusString, boolean isSuccessful) {
     this.statusString = statusString;
+    if (statusHistory.length() > historyMaxSize) {
+      statusHistory.delete(0, statusString.length() + 1);
```
lognaturel: It took me a bit to understand the delete bounds so just to have a record -- the first `+ 1` is needed to take into account the newline that will go with the new status. The second `+ 1` is because `lineEnd` is an index, not a length.

src/org/opendatakit/briefcase/ui/ScrollingStatusListDialog.java
```diff
@@ -128,8 +130,12 @@ private ScrollingStatusListDialog(Frame frame, String labelText, String title, I
   public void onEvent(FormStatusEvent event) {
     // Since there can be multiple FormStatusEvent's published concurrently,
     // we have to check if the event is meant for this dialog instance.
-    if (event.getStatus().getFormDefinition().equals(form)) {
-      editorArea.setText(event.getStatus().getStatusHistory());
+    if (isShowing() && event.getStatus().getFormDefinition().equals(form)) {
+      try {
+        editorArea.read(new StringReader(event.getStatus().getStatusHistory()), null);
```
lognaturel: That works well but it still seems unfortunate to need to build the model every time! Additionally, it means the scroll position doesn't update as more status messages are added -- I believe some users watch this live and the scrolling is convenient in that context.

What about something like:
```
      try {
        Document doc = editorArea.getDocument();
        String latestStatus = event.getStatus().getStatusString() + "\n";
        doc.insertString(doc.getLength(), latestStatus, null);
      } catch (BadLocationException e) {
        log.warn("Couldn't append status", e);
      }
```

I haven't profiled this so I may be wrong but I _think_ it will also be fast.

src/org/opendatakit/briefcase/model/FormStatus.java
batkinson: Hi @lognaturel, I'm not surprised you didn't see it jumping into this cold. The selection of 8K for the bound was arbitrary. Also, I made it a field and not a constant intentionally. If someone wanted to make it a configurable setting, it's ready to go. However, the introduction of the bound is the key to fixing the issue. The issue is O(n) allocations, n being status events, for every status update. Everyone that works on briefcase should get really acquainted with #138, #140 and #141. These are essentially the same problem in different forms. O(n) complexity in latency, storage and memory allocations, respectively. Briefcase is nearly *useless* for large data sets because of these issues.

src/org/opendatakit/briefcase/ui/ScrollingStatusListDialog.java
batkinson: So, this should only be an improvement on the previous code in terms of semantics. It also was done based on actual profiling of the allocations. It's effectively setText with the entire history just as before, but history is bounded so allocations don't grow linearly.

Careful with that suggestion - you just made the JEditorPane's buffer O(n) for memory consumption. I didn't attempt to gold-plate the existing solution, only fix the scale issue. If I were to spend any more time on this myself, my first attempt would be to use the logging framework and make these views simply bounded buffers that are polled periodically, not on every status updated.

lognaturel: Thanks for the clarifications -- I didn't realize that the issue was the total memory usage, I thought it was only the repeated allocation and deallocation of ever larger objects. I would have expected the buffer to be able to hold several MBs without trouble on a modern computer!

src/org/opendatakit/briefcase/model/FormStatus.java
lognaturel: I answered your comment below first and I'm having a little trouble reconciling it with this one! If updating the `JEditorPane` doesn't require allocating new objects, is there really a need for such a low limit? Is this because your machine is very memory constrained? I'm not saying the optimization is a bad one but I just want to make sure we have a record of why it's needed so we don't scratch our heads too much later.

I believe 8Kb is actually 8192 bytes which makes the choice of 8096 seem more meaningful than it is.

src/org/opendatakit/briefcase/ui/ScrollingStatusListDialog.java
batkinson: > I didn't realize that the issue was the total memory usage, I thought it was only the repeated allocation and deallocation of ever larger objects. 

No, you're right. *This* issue is to fix the allocator activity. My point in mentioning the memory usage is that by bounding the buffer you avoid yet another similar memory-bound issue as well.

My apologies in advance for any offense taken to the following, but it is meant only as direct and honest critique of the **code**, not the **authors** and especially not **you**.

If you're *really* careful, keeping around MB-scale log strings in memory is possible as you say. However, it's going to be a MB-scale log multiplied by each form *in memory at the same time* given the current implementation. And for what utility? That's just awful resource usage and for something of questionable value. Keeping a status log in an unbounded buffer, especially given the contents of that log, is at least a questionable decision. That it even has the *opportunity* to threaten the critical functionality of the tool, especially when the user is putting lots of data in it, seems more like a thoughtless design.

There are many ways to offer the same, or even better feedback, without sacrificing critical functionality at scale. One is squelch meaningless status messages to begin with, another is using a more appropriate feedback indicator for progress (like progress bars), another would be just moving the same status to storage and not memory (in other words, using machine resources more appropriately).

An unbounded buffer here is really hard to defend here IMO. 

I don't really have a ton of time, but if the unbounded buffer is a deal-breaker to fix this, I will  resubmit the PR with the unbounded status buffer using a technique similar to what @lognaturel suggested above.

src/org/opendatakit/briefcase/model/FormStatus.java
batkinson: Yep, that 8096 is just me screwing up my powers of two. Updating the `JEditorPane`'s entire buffer *does* require allocating new objects, at the scale of the contents. The difference between `setText` and the alternative I used is it cuts it in half by avoiding the view teardown. Using either approach really only works at scale by bounding the buffer given that the allocations are proportional to the editor content.

batkinson:
@lognaturel I just rebased this so it applies cleanly after recent PR merges. I also incorporated your feedback while limiting java.text allocations *and* bound the memory consumption. Now, each form status history can grow up to 1MB in size. When the limit is hit, the form's history is trimmed as new messages are appended. However, when the status dialog is opened, is only uses the form's history initially. New status messages are appended directly, rather than with `setText(..)` or similar, thus avoiding allocations proportional to the entire history. This also allows the dialog text to grow unbounded until you close it.

lognaturel:
✨✨✨

I totally agree that using some kind of progress bar for the successful cases would be a big improvement. Thanks for making this incremental fix for now.

Issue #147, opened 2017-07-17
Application logs exceptions to console


batkinson:
#### Software versions 
Briefcase v1.6.0 and lower

Due to numerous uses of `Exception.printStackTrace()`, the application logs exceptional conditions to the console rather than to a logging framework. All such stack dumps should be logged so that they can be managed using logging configuration.

batkinson:
Fixed with the integration of the referenced PR.

PR #148, opened 2017-07-17
Logging cleanup

batkinson:
Logging cleanup, mostly centered around #147.


lognaturel:
Wow, thanks so much for this much-needed cleanup. We will all breathe a little easier.

PR #150, opened 2017-07-19
Non-portable storage directories

### LINKED ISSUE #149: Fetched instances not recognized when storage directory is moved
#### Software versions 
Briefcase v1.6.0 and earlier

- Initiate a pull of a form to create tracked instances
- Close briefcase
- Move the storage directory
- Restart briefcase
- Reconfigure briefcase to point to the same storage directory in its new location
- Initiate a pull of the same form you fetched previously

#### Expected behavior

For the second pull, chunk processing should be significantly faster than the first since it should recognize the instances already fetched and skip the instance downloads. Instead, it is just as slow as the first run since it is re-fetching the same form instances. The problem is that the instance directories are stored as absolute paths so the storage directory is not portable. It needs to exist in the exact location or previously fetched instances are not recognized.

batkinson:
Fixes #149 by detecting and fixing instance tracking databases that use absolute paths. It makes them portable by using paths relative to the form directory.


lognaturel:
Thanks for this, it seems pretty inherent to the usefulness of Briefcase.

PR #106, opened 2017-04-19
Add success with errors event during pull . Fixes #102

icemc:
This PR addresses issue #102 . @yanokwa  Please could you review this

PR #151, opened 2017-07-19
Add success with errors. Fixes #102.

### LINKED ISSUE #102: Mark pulling process as "Succeeded with some errors" instead of "Failed" if some form instances were skipped and the pulling process completed
#### Software versions 
Briefcase v1.4.10,  operating system Windows 8

#### Problem description
During pulling(of all types Aggregate, Collect directory etc) of form instances if one of the instances could not be downloaded then it is skipped and the pulling continues but at the end the whole process is marked as FAILED! even though the process completed and was not aborted. I think such events should be marked as "Succeeded with some errors"

Try pulling a form if an error occurs such that an instance is not downloaded the pulling process is marked as "FAILED" even though it completed
#### Expected behavior
if some form instances could not be downloaded during pull the pull process should be marked as "Succeeded with some errors" rather than "FAILED"

yanokwa:
Credit goes to @icemc. Original issue is at https://github.com/opendatakit/briefcase/issues/102 and  original PR is at https://github.com/opendatakit/briefcase/pull/106.

### DIFF
--- a/src/org/opendatakit/briefcase/model/FinalStatus.java
+++ b/src/org/opendatakit/briefcase/model/FinalStatus.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright (C) 2011 University of Washington.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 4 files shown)


yanokwa:
Hi @icemc! Since you filed the original issue, do you have an easy test case for us to trigger the Success with Errors message?

icemc:
Hi @yanokwa sorry for reply this late. I came across this issue when I was downloading forms and my Internet connection wasn't the best. At some point I wasn't able to fetch some forms and I got a failed message. Initially I had an empty folder but at the end of the process I realized that briefcase was able to fetch some forms. This happened twince and I wasn't able to reproduce this issue intentionally.

yanokwa:
Thanks for the response, @icemc! I'm going to close the PR because if we can't reproduce the problem, then we can't confirm that the changes you made fixed said problems. If you end up being able to reproduce it, file another issue and PR and we'll gladly take another look!

PR #153, opened 2017-07-20
Fixes #152: Can not re-initiate submission pulls

### LINKED ISSUE #152: Loss in connection prevents Briefcase from downloading submissions
In Briefcase v1.6.0, pull forms from an Aggregate server where you know the number of submissions (or the number of submissions is high). (e.g., https://nafundi-test.appspot.com, login with test/test).

Turn off your connection halfway through the pull. You now have fewer submissions than the total number of submissions.

Then turn your connection back on and try to pull again. Briefcase will not download the remaining submissions. It is only after a relaunch of the app that pulling will resume.

@lognaturel believes that a UI bug that was introduced in https://github.com/opendatakit/briefcase/issues/49 and she's likely correct. I have confirmed this problem did not exist in 1.4.10.

batkinson:
Fixes the implementation of #49 so you can't initiate pulls twice, while allowing re-initiation of previous pulls.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/FormTransferTable.java
+++ b/src/org/opendatakit/briefcase/ui/FormTransferTable.java
@@ -35,7 +35,6 @@ import javax.swing.table.TableCellRenderer;
 import javax.swing.table.TableColumnModel;
 import javax.swing.table.TableModel;
 import javax.swing.table.TableRowSorter;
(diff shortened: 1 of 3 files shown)


lognaturel:
Thank you! I would not have thought to use the button state.

PR #156, opened 2017-07-27
Fixed #155: Redirect javarosa warnings from console to log file

### LINKED ISSUE #155: Redirect javarosa warnings from console to log file
Use the new functionality in javarosa (which is now available after integration of #154) to redirect its output to a log file rather than the console.

batkinson:
Finishes the console logging cleanup by redirecting javarosa output to a file in the root of the briefcase storage directory (or the user's home directory per the normal storage dir default).

### DIFF
--- a/src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
+++ b/src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
@@ -126,7 +126,7 @@ public class BaseFormParserForJavaRosa {
   private static void redirectOutput() {
     File jrLogFile = new File(FileSystemUtils.getBriefcaseFolder(), ".briefcase-javarosa.log");
     try {
-      final PrintStream jrOut = new PrintStream(jrLogFile);
+      PrintStream jrOut = new PrintStream(jrLogFile);
       Std.setOut(jrOut);
       Std.setErr(jrOut);
     } catch (FileNotFoundException e) {
--- a/src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
+++ b/src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
@@ -129,11 +129,6 @@ public class BaseFormParserForJavaRosa {
       final PrintStream jrOut = new PrintStream(jrLogFile);
       Std.setOut(jrOut);
       Std.setErr(jrOut);
-      Runtime.getRuntime().addShutdownHook(new Thread() {
-        public void run() {
-          jrOut.close();
-        }
-      });
     } catch (FileNotFoundException e) {
       log.warn("failed to redirect javarosa output to " + jrLogFile);
     }
--- a/src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
+++ b/src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
@@ -17,7 +17,10 @@
 
 package org.opendatakit.aggregate.parser;
 
+import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.PrintStream;
 import java.io.StringReader;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -29,6 +32,7 @@ import java.util.Map;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.javarosa.core.io.Std;
 import org.javarosa.core.model.DataBinding;
 import org.javarosa.core.model.FormDef;
 import org.javarosa.core.model.IDataReference;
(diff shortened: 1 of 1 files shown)


src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
```diff
@@ -111,12 +116,29 @@ private static final void initializeJavaRosa() {
...
     }
   }
 
+  private static void redirectOutput() {
+    File jrLogFile = new File(FileSystemUtils.getBriefcaseFolder(), ".briefcase-javarosa.log");
+    try {
+      final PrintStream jrOut = new PrintStream(jrLogFile);
+      Std.setOut(jrOut);
+      Std.setErr(jrOut);
+      Runtime.getRuntime().addShutdownHook(new Thread() {
```
dcbriccetti: I question the need for the shutdown hook. Is there reason to believe the PrintStream won’t be closed properly at application exit. When I removed the hook and ran the application, the log was created.

src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
```diff
@@ -111,12 +116,29 @@ private static final void initializeJavaRosa() {
              // new CoreModelModule().registerModule();
              // replace with direct call to PrototypeManager
              PrototypeManager.registerPrototypes(SERIALIABLE_CLASSES);
+             redirectOutput();
```
dcbriccetti: Since this call changes System.out and .err application wide, shouldn’t this call be made, in a very obvious way, at the top level of the application?

src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
```diff
@@ -111,12 +116,29 @@ private static final void initializeJavaRosa() {
...
+      final PrintStream jrOut = new PrintStream(jrLogFile);
+      Std.setOut(jrOut);
+      Std.setErr(jrOut);
+      Runtime.getRuntime().addShutdownHook(new Thread() {
+        public void run() {
+          jrOut.close();
+        }
+      });
+    } catch (FileNotFoundException e) {
+      log.warn("failed to redirect javarosa output to " + jrLogFile);
```
dcbriccetti: I would say “... redirect stdout and stderr” (or something to that effect) to reflect the fact that this impacts more than JavaRosa output.

src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
batkinson: I agree, it's not strictly necessary. 

> When I removed the hook and ran the application, the log was created.

That's weird. It should create the file anyway. I'm not sure why you mention these as a pair. The hook just ensures explicit close of the file (rather than relying on the OS to clean up the dangling file descriptors when killing the process).

src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
batkinson: > Since this call changes System.out and .err application wide...
It doesn't. It redirects output from the JavaRosa library only. It only needs to be redirected before the library is used (and generates output).

src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
batkinson: > reflect the fact that this impacts more than JavaRosa output.
...but it doesn't. This class is scoped to JavaRosa as are `setOut` and `setErr`. If we're using `Std` elsewhere, we shouldn't be. This is literally just to redirect the output that JavaRosa generates using the Std class, nothing more.

dcbriccetti: Oh, you’re so right! Sorry. Disregard.

src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
dcbriccetti: Is it really the OS that closes the file, rather than the JVM? In any case I leave it to you whether to keep the hook.

batkinson: > Is it really the OS that closes the file, rather than the JVM?
It depends on how the JVM exits (and exits successfully).
> In any case I leave it to you whether to keep the hook.
So, I think you're right. I was being a little pedantic here. Even if someone embedded the class, this would leak file descriptors since the shutdown hook would only run upon normal VM exit. It adds to the noise and isn't all that useful. I'll remove it.

dcbriccetti: Hurray for less noise!

batkinson:
Sorry @dcbriccetti , I apparently pushed a small cleanup fix that cleared the approval.

dcbriccetti:
>Sorry @dcbriccetti , I apparently pushed a small cleanup fix that cleared the approval.
I don’t think my approvals do anything, Github-wise. But they let @lognaturel and @yanokwa know I like the change.

PR #163, opened 2017-08-13
Create PULL_REQUEST_TEMPLATE.md

yanokwa:


lognaturel:
Naturally, I think this is brilliant. ;) @yanokwa I'm disappointed you didn't fill it out for this PR!

@daveycrockett @nribeka @batkinson @dcbriccetti @jpknox since you've been active on this repo recently, I want to give you a chance to comment. Or if it looks fine you can just thumbs up. I think you've all been providing this info already so it shouldn't be a big change. The goal is to add a little structure so the reviewer burden is a little lower and to reduce the chances of merging a bad change. Thanks!

daveycrockett:
I'd be interested in hearing people's opinions on PR'ing works-in-progress, especially at decision-making points, versus waiting to PR at "Closes #."  Depending on my availability I'll sometimes just take an action and call it out at final PR time, but are there times we feel like incremental progress is worth sharing?  If so, do we want a different format for those PR's that should be called out here?

dcbriccetti:
I like the template. If it doesn’t apply in a particular case (such as work in progress), I suppose one could delete the inapplicable parts.

lognaturel:
I really like work in progress (WIP) PRs. I would personally write my work in progress text above the template, then when I'm ready for a final review I'd move that text to the bottom for historical purposes and fill out the template.

jpknox:
Looks good to me, I wasn't sure about the most appropriate place to document my thoughts/findings during the change I've been working on; this will make it easier to communicate the key things relating to each change.

lognaturel:
Let's merge and keep the conversation going. If you get good ideas for improvements or how it can work better with WIP pull requests, please make the changes and submit a PR!

Issue #137, opened 2017-06-27
Add Google Analytics to Briefcase


yanokwa:
Better analytics will help with important design decisions.

In the settings, there should be a checkbox to disable analytics if desired. It should be checked by default.

jpknox:
Hello yanokwa, I'd like to have a go at this.
I've found a few user-made libraries that interact with Google's Analytics API.
May I have this task assigned to me?

yanokwa:
Welcome to ODK, @jpknox! Please have a go at this! 

What I'd like to get out of the analytics is what operating systems are in use, what version of ODK Briefcase is being run, and what users take (so maybe we track if they use the Pull/Push/Export) and what screens they switch to (Pull/Push/Export/Settings) in the GUI.

Bonus points if you can figure out command line usage as well.

lognaturel:
Hi @jpknox! How is this looking? If you want to explore different options, we're happy to give feedback and ideas here. Also consider joining the developer Slack [here](http://slack.opendatakit.org/)!

jpknox:
Hi Helene,


I've got the app to send my Google analytics account a message when it starts up, which includes the operating system in use.


I've been thinking about getting it to send a breakdown of the ratio of tab usage when the app shuts down, so we'll be able to track which tabs see the most usage - in terms of seconds active. This will require knowledge of Swing, of which I know very little; I've been working through a tutorial on Swing to help myself figure out the GUI.


Thank you for inviting me to the Slack channel, I'll hop in to it when I'm stuck or thinking about possibilities for the analytics.


Regards,

Joao Paulo Knox

________________________________

yanokwa:
Keep up the great work, @jpknox! It'd be awesome to get this into the next monthly release (code freeze on Aug 23rd). If you run into Swing issues, please ask on the Slack. Lots of friendly folks there and @dcbriccetti is our Swing expert.

downeym:
Newbie question here. I understand the value of having analytics be opt-out. But is there a chance for the user to opt-out at the time of initial installation/setup? Or is this feature something they'd have to stumble across by chance (or rather, through a thorough deep review of the settings)? I'd strongly encourage a UX-disruptive (interruption) chance for the user to opt out when they first use Briefcase. It would score big points for people sensitive about data privacy issues.

jpknox:
[...]

PR #164, opened 2017-08-13
#137 Add Google Analytics to Briefcase

jpknox:
Add a checkBox in the settings tab so the user can give or deny their consent to having their behaviour tracked. The setting is persisted.

Generate a UUID for each unique user. The UUID is persisted.

On startup track the following: App name (Briefcase), app version, operating system type, unique user ID.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,7 @@ dependencies {
     compile group: 'org.opendatakit', name: 'opendatakit-javarosa', version: '2.4.0'
     compile group: 'org.hsqldb', name: 'hsqldb', version: '2.4.0'
     compile files('lib/smallsql-0.21.jar')
+    compile group: 'com.brsanthu', name: 'google-analytics-java', version: '1.1.2'
 }
 
 // Required to use fileExtensions property in checkstyle file
@@ -75,6 +76,7 @@ buildConfig {
     version = getVersionName()
     clsName = 'BuildConfig'
     packageName = 'org.opendatakit.briefcase.buildconfig'
+    buildConfigField 'String', 'GOOGLE_TRACKING_ID', "UA-102618497-1"
 }
 
 mainClassName = 'org.opendatakit.briefcase.ui.MainBriefcaseWindow'
--- a/src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
+++ b/src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
@@ -1,0 +1,177 @@
+package org.opendatakit.briefcase.model;
+
+import com.brsanthu.googleanalytics.EventHit;
+import com.brsanthu.googleanalytics.GoogleAnalytics;
+import com.brsanthu.googleanalytics.GoogleAnalyticsConfig;
+import org.opendatakit.briefcase.buildconfig.BuildConfig;
+
+/**
+ *  This class is used to send usage information to ODK's Google Analytics account.
+ *
+ * <P>A library built by 'brsanthu' is utilized to streamline the
+ * process of contacting Google's Measurement Protocol API.</P>
+ * <P>This class also implements the Singleton pattern.</P>
+ *
+ * @author Joao Paulo Knox
+ * @see <a href="https://github.com/brsanthu/google-analytics-java">google-analytics-java</a>
+ */
+public class BriefcaseAnalytics {
+
+  private static final String GOOGLE_TRACKING_ID = BriefcasePreferences.GOOGLE_TRACKING_ID;
+  private static final String UNIQUE_USER_ID = BriefcasePreferences.getUniqueUserID();
+  private static final String EVENT_CATEGORY = "ODK Briefcase";
+  private static final String APPLICATION_NAME = BuildConfig.NAME;
+  private static final String APPLICATION_VERSION = BuildConfig.VERSION;
+
+  private static BriefcaseAnalytics instance;
+  private static final GoogleAnalyticsConfig config = new GoogleAnalyticsConfig();
+  private static final GoogleAnalytics googleAnalytics = new GoogleAnalytics(config, GOOGLE_TRACKING_ID);
+  private static boolean hasTrackedStartup;
+
+  /**
+   * Constructor
+   *
+   * <P>The constructor is private so as to adhere to the rules of the Singleton pattern.</P>
+   */
+  private BriefcaseAnalytics() {
+  }
+
+  /**
+   * Singleton creator.
+   *
+   * <P>This method (along with the private constructor) will ensure
+   * that only one instance of this class is instantiated.</P>
+   * @return the instantiated Singleton of {@link BriefcaseAnalytics}.
+   */
+  public static BriefcaseAnalytics getInstance() {
+    return instance == null ? (instance = new BriefcaseAnalytics()) : instance;
+  }
+
+  /**
+   *  Sends usage information to Google Analytics, only if the user has consented.
+   *  The information is modeled using an {@link EventHit} object.
+   *
+   * @param action (required) name of the event itself.
+   * @param label (optional) an open-ended label.
+   * @param value (optional) an open-ended value. Must be an integer.
+   * @return the {@link EventHit} object.
+   */
+  public EventHit trackEvent(String action, String label, Integer value) {
+    checkStateTrackingConsent();
+    EventHit eventHit = new EventHitBuilder()
+      .trackingID(GOOGLE_TRACKING_ID)
+      .clientID(UNIQUE_USER_ID)
+      .dataSource(System.getProperty("os.name"))
+      .applicationName(APPLICATION_NAME)
+      .applicationVersion(APPLICATION_VERSION)
+      .eventCategory(EVENT_CATEGORY)
+      .eventAction(action)
+      .eventLabel(label)
+      .eventValue(value)
+      .build();
+    googleAnalytics.postAsync(eventHit);
+    return eventHit;
+  }
+
+  /**
+   * Track this application's startup event.
+   *
+   * <P>This should only be run once per application execution.</P>
+   * <P>Note that the method will check to see if it has already tracked this session's startup.</P>
+   */
+  public void trackStartup() {
+    if (!hasTrackedStartup) {
+      trackEvent("Application-Startup", null, null);
+      hasTrackedStartup = true;
+    }
+  }
+
+  /**
+   * Check if the user has set or updated their consent
+   * to ODK tracking their application usage.
+   *
+   * <P>Note that this method should be executed before any event is tracked,
+   * in case the user has unticked the consent checkbox within the session.</P>
+   */
+  private void checkStateTrackingConsent() {
+    config.setGatherStats(BriefcasePreferences.getBriefcaseTrackingConsentProperty());
+  }
+
+  /**
+   * A helper class used to streamline the creation of EventHits.
+   *
+   * <P>An {@link EventHit} object is used to model the information being tracked.</P>
+   */
+  private static class EventHitBuilder {
+
+    private String trackingID;
+    private String clientID;
+    private String dataSource;
+    private String applicationName;
+    private String applicationVersion;
+    private String eventCategory;
+    private String eventAction;
+    private String eventLabel;
+    private Integer eventValue;
+
+    public EventHitBuilder trackingID(String trackingID) {
+      this.trackingID = trackingID;
+      return this;
+    }
+
+    public EventHitBuilder clientID(String clientID) {
+      this.clientID = clientID;
+      return this;
+    }
+
+    public EventHitBuilder dataSource(String dataSource) {
+      this.dataSource = dataSource;
+      return this;
+    }
+
+    public EventHitBuilder applicationName(String applicationName) {
+      this.applicationName = applicationName;
+      return this;
+    }
+
+    public EventHitBuilder applicationVersion(String applicationVersion) {
+      this.applicationVersion = applicationVersion;
+      return this;
+    }
+
+    public EventHitBuilder eventCategory(String eventCategory) {
+      this.eventCategory = eventCategory;
+      return this;
+    }
+
+    public EventHitBuilder eventAction(String eventAction) {
+      this.eventAction = eventAction;
+      return this;
+    }
+
+    public EventHitBuilder eventLabel(String eventLabel) {
+      this.eventLabel = eventLabel;
+      return this;
+    }
+
+    public EventHitBuilder eventValue(Integer eventValue) {
+      this.eventValue = eventValue;
+      return this;
+    }
+
+    public EventHit build() {
+      EventHit eventHit = new EventHit();
+      eventHit.trackingId(trackingID);
+      eventHit.clientId(clientID);
+      eventHit.dataSource(dataSource);
+      eventHit.applicationName(applicationName);
+      eventHit.applicationVersion(applicationVersion);
+      eventHit.eventCategory(eventCategory);
+      eventHit.eventAction(eventAction);
+      eventHit.eventLabel(eventLabel);
+      eventHit.eventValue(eventValue);
+      return eventHit;
+    }
+
+  }
+}
--- a/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
+++ b/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
@@ -16,6 +16,7 @@
 
 package org.opendatakit.briefcase.model;
 import java.security.Security;
+import java.util.UUID;
 import java.util.prefs.Preferences;
 
 import org.apache.http.HttpHost;
@@ -29,6 +30,7 @@ import org.opendatakit.briefcase.buildconfig.BuildConfig;
 public class BriefcasePreferences {
   
   public static final String VERSION = BuildConfig.VERSION;
+  public static final String GOOGLE_TRACKING_ID = BuildConfig.GOOGLE_TRACKING_ID;
   public static final String USERNAME = "username";
   public static final String TOKEN = "token";
   public static final String AGGREGATE_1_0_URL = "url_1_0";
@@ -38,6 +40,8 @@ public class BriefcasePreferences {
   private static final String BRIEFCASE_PROXY_HOST_PROPERTY = "briefcaseProxyHost";
   private static final String BRIEFCASE_PROXY_PORT_PROPERTY = "briefcaseProxyPort";
   private static final String BRIEFCASE_PARALLEL_PULLS_PROPERTY = "briefcaseParallelPulls";
+  private static final String BRIEFCASE_TRACKING_CONSENT_PROPERTY = "briefcaseTrackingConsent";
+  private static final String BRIEFCASE_UNIQUE_USER_ID_PROPERTY = "uniqueUserID";
   
   static {
     // load the security provider
@@ -57,7 +61,7 @@ public class BriefcasePreferences {
   public static BriefcasePreferences forClass(Class<?> node) {
     return new BriefcasePreferences(node, PreferenceScope.CLASS_NAME);
   }
-  
+
   /**
    * Returns the value associated with the specified key in this preference
    * node. Returns the specified default if there is no value associated with
@@ -180,5 +184,37 @@ public class BriefcasePreferences {
       return new HttpHost(host, port);
     }
     return null;
+  }
+
+  /**
+   * Persist the user's decision to allow/disallow their behaviour being tracked.
+   * @param value (required) the boolean value representing the user's decision.
+   */
+  public static void setBriefcaseTrackingConsentProperty(boolean value) {
+      Preference.APPLICATION_SCOPED.put(BriefcasePreferences.BRIEFCASE_TRACKING_CONSENT_PROPERTY, Boolean.valueOf(value).toString());
+  }
+
+  /**
+   * Get the user's persisted decision regarding their consent to being tracked.
+   * @return the boolean representation of the user's consent to being tracked.
+   */
+  public static boolean getBriefcaseTrackingConsentProperty() {
+    return Boolean.valueOf(
+            Preference.APPLICATION_SCOPED.get(BriefcasePreferences.BRIEFCASE_TRACKING_CONSENT_PROPERTY, Boolean.FALSE.toString())
+    ).booleanValue();
+  }
+
+  /**
+   * Get the persisted UUID for the current user.
+   * <P>Note that the method will generate and persist a UUID if the user doesn't have one.</P>
+   * @return the String representation of the user's UUID.
+   */
+  public static String getUniqueUserID() {
+    String defaultUuidValue = "UUID missing, defaulting to this message";
+    String uniqueUserID = Preference.APPLICATION_SCOPED.get(BriefcasePreferences.BRIEFCASE_UNIQUE_USER_ID_PROPERTY, defaultUuidValue);
+    if ( uniqueUserID.equals(defaultUuidValue)) {
+      Preference.APPLICATION_SCOPED.put(BriefcasePreferences.BRIEFCASE_UNIQUE_USER_ID_PROPERTY, UUID.randomUUID().toString());
+    }
+    return Preference.APPLICATION_SCOPED.get(BriefcasePreferences.BRIEFCASE_UNIQUE_USER_ID_PROPERTY, defaultUuidValue);
   }
 }
--- a/src/org/opendatakit/briefcase/ui/MessageStrings.java
+++ b/src/org/opendatakit/briefcase/ui/MessageStrings.java
@@ -81,5 +81,6 @@ public class MessageStrings {
       "If you are behind a proxy, try setting up your proxy details through 'Settings' tab.";
 
   public static final String PARALLEL_PULLS = "Pull submissions in parallel (experimental)";
+  public static final String TRACKING_CONSENT = "Allow tracking of non-personal usage data";
 
 }
--- a/src/org/opendatakit/briefcase/ui/SettingsPanel.java
+++ b/src/org/opendatakit/briefcase/ui/SettingsPanel.java
@@ -47,6 +47,8 @@ public class SettingsPanel extends JPanel {
     private JSpinner spinPort;
     private JLabel lblParallel;
     private JCheckBox chkParallel;
+    private JLabel lblTrackingConsent;
+    private JCheckBox chkTrackingConsent;
 
     public SettingsPanel(MainBriefcaseWindow parentWindow) {
         this.parentWindow = parentWindow;
@@ -83,6 +85,11 @@ public class SettingsPanel extends JPanel {
         chkParallel.setSelected(BriefcasePreferences.getBriefcaseParallelPullsProperty());
         chkParallel.addActionListener(new ParallelPullToggleListener());
 
+        lblTrackingConsent = new JLabel(MessageStrings.TRACKING_CONSENT);
+        chkTrackingConsent = new JCheckBox();
+        chkTrackingConsent.setSelected(BriefcasePreferences.getBriefcaseTrackingConsentProperty());
+        chkTrackingConsent.addActionListener(new TrackingConsentToggleListener());
+
         GroupLayout groupLayout = new GroupLayout(this);
         groupLayout.setHorizontalGroup(
           groupLayout.createSequentialGroup()
@@ -90,7 +97,8 @@ public class SettingsPanel extends JPanel {
             .addGroup(
               groupLayout.createParallelGroup(Alignment.TRAILING)
                 .addComponent(chkProxy)
-                .addComponent(chkParallel))
+                .addComponent(chkParallel)
+                .addComponent(chkTrackingConsent))
             .addGroup(
               groupLayout.createParallelGroup(Alignment.LEADING)
                 .addGroup(
@@ -109,7 +117,8 @@ public class SettingsPanel extends JPanel {
                       groupLayout.createParallelGroup(Alignment.LEADING)
                         .addComponent(txtHost, GroupLayout.PREFERRED_SIZE, GroupLayout.PREFERRED_SIZE, GroupLayout.PREFERRED_SIZE)
                         .addComponent(spinPort, GroupLayout.PREFERRED_SIZE, GroupLayout.PREFERRED_SIZE, GroupLayout.PREFERRED_SIZE)))
-                      .addComponent(lblParallel))
+                      .addComponent(lblParallel)
+                      .addComponent(lblTrackingConsent))
             .addContainerGap()
         );
         groupLayout.setVerticalGroup(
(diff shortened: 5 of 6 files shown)


build.gradle
```diff
@@ -75,6 +76,7 @@ buildConfig {
     version = getVersionName()
     clsName = 'BuildConfig'
     packageName = 'org.opendatakit.briefcase.buildconfig'
+    buildConfigField 'String', 'GOOGLE_TRACKING_ID', "UA-102618497-1"
```
jpknox: This ID is just one I've made for testing purposes.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,182 @@
...
+   * <P>Note that this method should be executed before any event is tracked,
+   * in case the user has unticked the consent checkbox within the session.</P>
+   */
+  private void checkTrackingConsentState() {
+    briefcaseTrackingConsent = BriefcasePreferences.getBriefcaseTrackingConsentProperty();
+    config.setGatherStats(briefcaseTrackingConsent);
+  }
+
+  /**
+   * A helper class used to streamline the creation Events.
```
jpknox: Typo: "creation Events" should be "creation of Events".

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
```diff
@@ -181,4 +185,40 @@ public static HttpHost getBriefCaseProxyConnection() {
...
+  public static void setBriefcaseTrackingConsentProperty(Boolean value) {
+    if (value == null) {
+      Preference.APPLICATION_SCOPED.remove(BRIEFCASE_TRACKING_CONSENT_PROPERTY);
+    } else {
+      Preference.APPLICATION_SCOPED.put(BriefcasePreferences.BRIEFCASE_TRACKING_CONSENT_PROPERTY, value.toString());
+    }
+  }
+
+  /**
+   * Get the user's persisted decision to being tracked.
```
jpknox: Grammar: should be more a long the lines of "Get the user's persisted decision regarding their consent to being tracked".

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
dcbriccetti: Since build() returns an ```EventHit```, maybe it should read “... creation of events” or “... creation of EventHits”.

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
```diff
@@ -181,4 +185,40 @@ public static HttpHost getBriefCaseProxyConnection() {
...
+  }
+
+  /**
+   * Get the persisted UUID for the current user.
+   * <P>Note that the method will generate and persist a UUID if the user doesn't have one.</P>
+   * @return the String representation of the user's UUID.
+   */
+  public static String getUniqueUserID() {
+    String userIdErrorMessage = "An error occurred generating the UUID.";
+    String uniqueUserID = Preference.APPLICATION_SCOPED.get(BriefcasePreferences.BRIEFCASE_UNIQUE_USER_ID_PROPERTY, userIdErrorMessage);
```
jpknox: This line along with 217 should be refactored to make it clearer that the Preferences will return the userIdErrorMessage value only if no UUID already exists. At the moment it looks like some kind of obscure critical error is being caught.

src/org/opendatakit/briefcase/ui/SettingsPanel.java
```diff
@@ -1,4 +1,4 @@
-package org.opendatakit.briefcase.ui;
+    package org.opendatakit.briefcase.ui;
```
jpknox: I'll undo this.

src/org/opendatakit/briefcase/ui/SettingsPanel.java
```diff
@@ -229,6 +241,22 @@ public void actionPerformed(ActionEvent e) {
...
-
 
+    /**
+     * This listener will pass the user's consent to being tracked onto the
+     * application's preferences so it can be persisted and used elsewhere.
+     */
+    public class TrackingConsentToggleListener implements ActionListener {
+        @Override
+        public void actionPerformed(ActionEvent e) {
+            System.out.println("Tracking consent checkbox toggled.");
```
jpknox: I'll remove this line.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,182 @@
...
+   *
+   * <P>An {@link EventHit} object is used to model the information being tracked.</P>
+   */
+  private static class EventBuilder {
```
dcbriccetti: It builds EventHit objects, so maybe call it EventHitBuilder?

src/org/opendatakit/briefcase/ui/MessageStrings.java
```diff
@@ -81,5 +81,6 @@
       "If you are behind a proxy, try setting up your proxy details through 'Settings' tab.";
 
   public static final String PARALLEL_PULLS = "Pull submissions in parallel (experimental)";
+  public static final String TRACKING_CONSENT = "Allow tracking of non-personal usage data.";
```
dcbriccetti: The other lines don’t end with a period.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,182 @@
...
+ * @author Joao Paulo Knox
+ * @see <a href="https://github.com/brsanthu/google-analytics-java">google-analytics-java</a>
+ */
+public class BriefcaseAnalytics {
+
+  private static final String GOOGLE_TRACKING_ID = BriefcasePreferences.GOOGLE_TRACKING_ID;
+  private static final String UNIQUE_USER_ID = BriefcasePreferences.getUniqueUserID();
+  private static final String EVENT_CATEGORY = "ODK Briefcase";
+
+  private static BriefcaseAnalytics instance = null;
```
dcbriccetti: The redundant initializers here are noise that slow down reading. I suggest:
```
  private static BriefcaseAnalytics instance;
  private static GoogleAnalytics googleAnalytics;
  private static final GoogleAnalyticsConfig config = new GoogleAnalyticsConfig();
  private static boolean hasTrackedStartup;
```
And ```config``` can be initialized here and made final.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,182 @@
...
+
+  /**
+   * Constructor
+   *
+   * <P>This constructor will instantiate 'brsanthu's {@link GoogleAnalytics} class and
+   * check if the user is okay with their behaviour being tracked.</P>
+   * <P>The constructor is private so as to adhere to the rules of the Singleton pattern.</P>
+   */
+  private BriefcaseAnalytics() {
+    config = new GoogleAnalyticsConfig();
```
dcbriccetti: This line no longer needed.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,182 @@
...
+    checkTrackingConsentState();
+    this.googleAnalytics = new GoogleAnalytics(config, GOOGLE_TRACKING_ID);
```
dcbriccetti: Remove ```this.```. Good style says to not access static elements via an instance. (Are you using IDEA? It highlights these things.)

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,182 @@
...
+
+  /**
+   * Singleton creator.
+   *
+   * <P>This method (along with the private constructor) will ensure
+   * that only one instance of this class is instantiated.</P>
+   * @return the instantiated Singleton of {@link BriefcaseAnalytics}.
+   */
+  public static BriefcaseAnalytics getInstance() {
+    return instance == null ? new BriefcaseAnalytics() : instance;
```
dcbriccetti: ```instance``` always remains null. I suggest
```
    if (instance == null) {
      instance = new BriefcaseAnalytics();
    }
    return instance;
```

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  /**
+   * Track this application's startup event.
+   *
+   * <P>This should only be run once per application execution.</P>
+   * <P>Note that the method will check to see if it has already tracked this session's startup.</P>
+   */
+  public void trackStartup() {
+    if (hasTrackedStartup == false) {
```
dcbriccetti: Reduce noise with ```!hasTrackedStartup```.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,182 @@
...
+public class BriefcaseAnalytics {
+
+  private static final String GOOGLE_TRACKING_ID = BriefcasePreferences.GOOGLE_TRACKING_ID;
+  private static final String UNIQUE_USER_ID = BriefcasePreferences.getUniqueUserID();
+  private static final String EVENT_CATEGORY = "ODK Briefcase";
+
+  private static BriefcaseAnalytics instance = null;
+  private static GoogleAnalytics googleAnalytics = null;
+  private static GoogleAnalyticsConfig config = null;
+  private static Boolean briefcaseTrackingConsent = null;
```
dcbriccetti: briefcaseTrackingConsent  doesn’t need to be a field.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,182 @@
...
+
+  /**
+   * Check if the user has set or updated their consent
+   * to ODK tracking their application usage.
+   *
+   * <P>Note that this method should be executed before any event is tracked,
+   * in case the user has unticked the consent checkbox within the session.</P>
+   */
+  private void checkTrackingConsentState() {
+    briefcaseTrackingConsent = BriefcasePreferences.getBriefcaseTrackingConsentProperty();
```
dcbriccetti: With the field removed, these two lines can become
```    config.setGatherStats(BriefcasePreferences.getBriefcaseTrackingConsentProperty());```

File: src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java [hunk 16]
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  /**
+   * A helper class used to streamline the creation Events.
+   *
+   * <P>An {@link EventHit} object is used to model the information being tracked.</P>
+   */
+  private static class EventBuilder {
+
+    private String trackingID = null;
```
dcbriccetti: All of these redundant ``` = null``` initializers slow reading and add no value.

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
```diff
@@ -181,4 +185,40 @@ public static HttpHost getBriefCaseProxyConnection() {
     }
     return null;
   }
+
+  /**
+   * Persist the user's decision to allow/disallow their behaviour being tracked.
+   * @param value (required) the boolean value representing the user's decision.
+   */
+  public static void setBriefcaseTrackingConsentProperty(Boolean value) {
+    if (value == null) {
```
dcbriccetti: Collapse four lines into:
```BriefcasePreferences.setBriefcaseTrackingConsentProperty(chkTrackingConsent.isSelected());```

File: src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java (hunk 14 above)
Comment (jpknox) [inline src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java]: I like how it improves readability especially within checkStateTrackingConsent().

Comment (dcbriccetti) [inline src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java]: OK, how about you use it there, but as a variable local to the method (not a field)?

File: src/org/opendatakit/briefcase/model/BriefcasePreferences.java (hunk 17 above)
Comment (jpknox) [inline src/org/opendatakit/briefcase/model/BriefcasePreferences.java]: Good point, will do.

File: src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java (hunk 16 above)
Comment (jpknox) [inline src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java]: I'll remove them. Once in the past I tried to initialise a variable within a try-catch black and it insisted that I declare the variable as null before the 'try' section, ever since then I've gotten in to the habit of declaring absolutely everything null - I may rethink this :).

File: src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java (hunk 2 above)
Comment (jpknox) [inline src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java]: Nicely spotted, I'll change it to "creation of EventHits".

File: src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java (hunk 7 above)
Comment (jpknox) [inline src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java]: Nice spot once again, I'll amend it.

File: src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java (hunk 16 above)
Comment (dcbriccetti) [inline src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java]: This may bolster your confidence:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-4.html#jls-4.12.5

File: src/org/opendatakit/briefcase/ui/SettingsPanel.java [hunk 18]
```diff
@@ -83,14 +85,20 @@ public SettingsPanel(MainBriefcaseWindow parentWindow) {
         chkParallel.setSelected(BriefcasePreferences.getBriefcaseParallelPullsProperty());
         chkParallel.addActionListener(new ParallelPullToggleListener());
 
+        lblTrackingConsent = new JLabel(MessageStrings.TRACKING_CONSENT);
```
jpknox: I'd hoped it would as straightforward as giving each label an ActionListener, but they don't have a method to set one as such.

src/org/opendatakit/briefcase/ui/MessageStrings.java
jpknox: I swore I'd double checked this but you're right, my mistake :p

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
dcbriccetti: Here’s how I discovered that ```instance``` was never set (other than being redundantly set to null via its initializer): I removed that initializer, then IDEA told me immediately that it was never set. So, thanks, IDEA!

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: Good point. I was hesitant to set `config` as final because I needed to make changes within it, but I later learned how final only stops the variable from being redeclared, it doesn't stop the developer from making changes within the object itself.

src/org/opendatakit/briefcase/ui/SettingsPanel.java
dcbriccetti: I would have used this constructor
https://docs.oracle.com/javase/7/docs/api/javax/swing/JCheckBox.html#JCheckBox(java.lang.String)
to provide the label. Then clicking on the label would toggle the checkbox. You did the right thing, probably, by continuing in the existing pattern.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
dcbriccetti: Also noticed I recommend you change that Boolean to boolean. No need to use a reference type when a primitive will do. This boolean only needs two states, not three.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: My goal is to persist its state across the whole life cycle of the application, so a accidental second call to it in the app won't double the number of startups logged. I believe that if it is declared in the method it won't persist to the next time the method is called - I may be mistaken on this.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: Will remove.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
dcbriccetti: Again, IDEA—alerted me that it doesn’t need to be a field. Meaning you weren’t doing anything with it that required its value being remembered outside that use in a single function.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: I must have missed them. You're right about how I shouldn't refer to class-wide variables using the context of the current instance - will fix.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: That's pretty cool, Eclipse was good albeit my time with it has helped me to really appreciate IntelliJ.

jpknox: Thanks for finding the bug; I've gone for `return instance == null ? (instance = new BriefcaseAnalytics()) : instance;`

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: Your suggestion is clearer, will do.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: I think your comment was meant for line 253 in 'SettingsPanel.java', good point though - will do.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: I'll remember that, thanks for the tip 👍

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    public EventBuilder eventValue(Integer eventValue) {
+      this.eventValue = eventValue;
+      return this;
+    }
+
+    public EventHit build() {
+      EventHit eventHit = new EventHit();
+      eventHit.trackingId(this.trackingID);
```
dcbriccetti: All the ```this.```s here add no value.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
dcbriccetti: No, I meant here.
```
  private void checkTrackingConsentState() {
    config.setGatherStats(BriefcasePreferences.getBriefcaseTrackingConsentProperty());
  }
```
You can always look at https://github.com/dcbriccetti/briefcase/commit/ac98cfe2fc31b080aabe30225b1688775af1866e to see all my recommended changes, made.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: You have a good point so I tried to switch them to boolean, where I found that BriefcasePreferences insists on storing things as Strings, which is much easier if I use Boolean instead, otherwise I'd be switching from Boolean to boolean which will make things complicated. It's getting a bit late here so I'm going to leave them as Boolean.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: I'll remove them.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
dcbriccetti: I was referring to ```hasTrackedStartup```, which is not used in the way you describe.

dcbriccetti: setBriefcaseTrackingConsentProperty(Boolean value) takes a Boolean, and if you pass it a boolean, “auto boxing” will do the right thing.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: I'm still not sure it will stop a second startup being tracked if I remove the field and call it twice. The Boolean variable 'hasTrackedStartup' must have its state persisted across each method call in order to stop a developer accidentally tracking unnecessary startup events.

dcbriccetti: In my modified version of your code (which you might look at), I have this:
```
  private static boolean hasTrackedStartup;
```
It does need to be a field. ```briefcaseTrackingConsent``` doesn’t.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: Oh, I see what you mean. I like how my way makes the method clearer to read but I prefer how your way makes the class look clearer overall by decreasing the number of fields - I'll switch to your way.

I'm grateful that you made the changes, but I find that I learn a lot better by doing the changes myself. I'm going to keep doing the changes manually, for now at least.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: My mistake, I just realised you were talking about `briefcaseTrackingConsent`.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
dcbriccetti: By all means, yes. I offer it to you if you want to see that the result does, in fact, work, and to make clearer anything I missed with the individual comments. 

I love Git and how I can get a local copy of a pull request and run and experiment with it myself.

jpknox:
I've updated my PR in response to @dcbriccetti's feedback.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,179 @@
...
+public class BriefcaseAnalytics {
+
+  private static final String GOOGLE_TRACKING_ID = BriefcasePreferences.GOOGLE_TRACKING_ID;
+  private static final String UNIQUE_USER_ID = BriefcasePreferences.getUniqueUserID();
+  private static final String EVENT_CATEGORY = "ODK Briefcase";
+  private static final String APPLICATION_NAME = BuildConfig.NAME;
+  private static final String APPLICATION_VERSION = BuildConfig.VERSION;
+
+  private static BriefcaseAnalytics instance;
+  private static final GoogleAnalyticsConfig CONFIG = new GoogleAnalyticsConfig();
```
dcbriccetti: The all-caps suggests that config is a constant, but it isn’t.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,179 @@
...
+  private static boolean hasTrackedStartup;
+
+  /**
+   * Constructor
+   *
+   * <P>This constructor will instantiate 'brsanthu's {@link GoogleAnalytics} class.</P>
+   * <P>The constructor is private so as to adhere to the rules of the Singleton pattern.</P>
+   */
+  private BriefcaseAnalytics() {
+    hasTrackedStartup = false;
```
dcbriccetti: This line serves no purpose. It’s already false.

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
```diff
@@ -181,4 +185,40 @@ public static HttpHost getBriefCaseProxyConnection() {
...
+    );
+  }
+
+  /**
+   * Get the persisted UUID for the current user.
+   * <P>Note that the method will generate and persist a UUID if the user doesn't have one.</P>
+   * @return the String representation of the user's UUID.
+   */
+  public static String getUniqueUserID() {
+    String defaultUuidValue = "UUID missing, defaulting to this value";
```
dcbriccetti: The important noun in this noun phrase is message, not value. It’s a message about a missing UUID.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -0,0 +1,179 @@
...
+
+  private static BriefcaseAnalytics instance;
+  private static final GoogleAnalyticsConfig CONFIG = new GoogleAnalyticsConfig();
+  private static GoogleAnalytics googleAnalytics = new GoogleAnalytics(CONFIG, GOOGLE_TRACKING_ID);
+  private static boolean hasTrackedStartup;
+
+  /**
+   * Constructor
+   *
+   * <P>This constructor will instantiate 'brsanthu's {@link GoogleAnalytics} class.</P>
```
dcbriccetti: The comment is no longer correct.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: I got confused by the difference between constants and final references.
`static final int YEAR = 2017` would be a constant because of how (I think) it's a primitive type.
'static final MyClass myClass' would just be a final reference because of how it's a complex type which cannot be reassigned to some other instance/object.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: I was confusing boolean for Boolean here in terms of behaviour. Boolean defaults to null whereas boolean defaults to false.

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
jpknox: Good point, I'll amend it.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: Technically it is, when the class is instantiated it will also instantiate the `GoogleAnalytics` field.

jpknox:
PR Updated.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
dcbriccetti: Not so. Since googleAnalytics is a class variable (static), it is initialized before any instances of the class are created.

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
dcbriccetti: I don’t see this change yet. Perhaps you’re still making changes.

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
```diff
@@ -181,4 +185,40 @@ public static HttpHost getBriefCaseProxyConnection() {
...
+    } else {
+      Preference.APPLICATION_SCOPED.put(BriefcasePreferences.BRIEFCASE_TRACKING_CONSENT_PROPERTY, value.toString());
+    }
+  }
+
+  /**
+   * Get the user's persisted decision regarding their consent to being tracked.
+   * @return the boolean representation of the user's consent to being tracked.
+   */
+  public static Boolean getBriefcaseTrackingConsentProperty() {
```
dcbriccetti: This should return ```boolean``` (primitive, not reference). No need to complicate things with the return type suggesting the value could be null.

jpknox: I was against doing this until I found both JCheckedBox.isSelected() and GoogleAnalyticsConfig.setGatherStats() use boolean instead of Boolean. My choice to go with Boolean was because it made persisting it easier because of how it _has_ to be stored as a String, ergo Boolean.toString(). I've read that it's bad to jump from similar type to similar type (Boolean to boolean and vice versa), which I couldn't avoid even though I thought I was doing so. Upon learning about the two previously mentioned signatures I decided it's probably best to keep the complexity of storing it inside the class responsible for storing it, and to keep using boolean everywhere else.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: I disagree, because googleAnalytics is a static field, it is initialised when the first instance of BriefcaseAnalytics is initialised, BriefcaseAnalytics is a Singleton so the comment on the constructor is technically correct, albeit it's not necessary and it adds noise. 

Your comment was thought provoking and it led me to learn more about Java, keep them coming 👍 .

jpknox:
Updated the PR. I've converted the getter/setter of the tracking-consent boolean value to use a primitive boolean instead of a Boolean, to contain the complexity of how BriefcasePreferences needs to store a String, it will continue to convert it to a Boolean before ultimately converting it to a String; the reverse is true when getting it.

I removed a line from the comment for BriefcaseAnalytics' constructor.

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
dcbriccetti: Glad we are coming together. A note on autoboxing: performance might be an issue if it happens frequently. Otherwise it’s secondary to type correctness and clarity.

https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
dcbriccetti: You are mistaken, my new friend, as I demonstrate in the debugger here.

screen shot 2017-08-15 at 12 00 45 pm

It is created when this line is executed from ```MainBriefcaseWindow```:
```  private static BriefcaseAnalytics briefcaseAnalytics = BriefcaseAnalytics.getInstance();```

But it’s moot now because you have removed the comment.

dcbriccetti:
OK, @lognaturel, I’m happy with this and believe it can be merged. I want to come back and fix the JCheckBoxes (problem that preexisted this change), and once Briefcase is set up for testing, add some tests.

lognaturel:
Hurray! Thanks for working closely on this @jpknox and @dcbriccetti! And a huge thanks to @jpknox for taking this on, it's a feature that has been on the wishlist for a very long time.

@yanokwa do you have a tracking ID for this? If you share it in the Slack with @jpknox he can change it when he has a moment before we merge.

yanokwa:
I'll update the tracking ID in master as soon as I merge.

PR #134, opened 2017-06-15
Replace tabs with spaces

yanokwa:
This is a whitespace-only change that fixes #131.

The command used was from https://stackoverflow.com/a/11094620/152938
`find . -name '*.java' ! -type d -exec bash -c 'expand -t 4 "$0" > /tmp/e && mv /tmp/e "$0"' {} \;`

PR #167, opened 2017-08-15
Replace tabs with spaces

### LINKED ISSUE #131: Replace tabs with spaces
The following classes have tabs that should be replaced with spaces. 
```
AggregateUtils.java
BadFormDefinition.java
BriefcaseFolderChooser.java
BriefcaseStorageLocationDialog.java
DatabaseUtils.java
ExportFolderChooser.java
ExportType.java
FormDefinitionFileChooser.java
FormTransferTable.java
HtmlConsts.java
HtmlStrUtil.java
LegacyServerConnectionDialog.java
MainBriefcaseWindow.java
ODKCollectFileChooser.java
PrivateKeyFileChooser.java
PushTransferPanel.java
ServerConnectionDialog.java
ServerFetcher.java
StringUtils.java
TransferFromODK.java
TransferInProgressDialog.java
WebUtils.java
XFormParameters.java
```

Use IntelliJ's conversion tool to convert the tabs with spaces, then uncomment out the following in checkstyle.xml.
```
    <module name="FileTabCharacter">
        <property name="eachLine" value="true"/>
    </module>
```

Run `./gradlew checkstyle` and fix any remaining issues.

To make code review easy, please make sure the PR has mostly whitespace changes.

yanokwa:
Closes #131

Ran `./gradlew check` and `./gradlew run` to make sure there are no errors and the application runs. https://github.com/opendatakit/briefcase/pull/134 was an old PR that I closed because of inactivity with the same approach.

I couldn't find a nice way to do this in IntelliJ, so the command I used was from https://stackoverflow.com/a/11094620/152938.
`find . -name '*.java' ! -type d -exec bash -c 'expand -t 4 "$0" > /tmp/e && mv /tmp/e "$0"' {} \;`

There might be some place where a tab is inside a string, but that's why I used [expand](https://man.cx/expand), not sed.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -16,9 +16,9 @@
 
     <property name="fileExtensions" value="java, properties, xml"/>
 
-    <!-- <module name="FileTabCharacter">
+    <module name="FileTabCharacter">
         <property name="eachLine" value="true"/>
-    </module> -->
+    </module>
 
     <!-- ODK-specific additions -->
     <module name="RegexpMultiline">
--- a/src/org/opendatakit/aggregate/form/XFormParameters.java
+++ b/src/org/opendatakit/aggregate/form/XFormParameters.java
@@ -25,18 +25,18 @@ package org.opendatakit.aggregate.form;
  */
 public final class XFormParameters implements Comparable<Object> {
 
-	public final String formId;
-	public final String versionString;
-	public final Long modelVersion;
+    public final String formId;
+    public final String versionString;
+    public final Long modelVersion;
 
-	public XFormParameters(String formId, String versionString) {
-		if ( formId == null ) {
-			throw new IllegalArgumentException("formId cannot be null");
-		}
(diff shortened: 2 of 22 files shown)


lognaturel:
@dcbriccetti @daveycrockett @nribeka How do you all feel about this being merged now? We have relatively few opened PRs so it might be a good time to get it done. Or perhaps we can finish and merge @daveycrockett's #139 first? I think he might be gone on 🏝vacation🏝 but perhaps they're small enough things that I can push to his branch.

dcbriccetti:
It would be nice to get #139 put away. My only concerns are avoiding a lot of work that these should-be-trivial changes can cause. It’s nice when they happen when there are no unmerged branches active.

lognaturel:
Yeah, you're totally right. Let's see if we can get both the open ones in before we do this.

yanokwa:
Closing because I'd rather https://github.com/opendatakit/briefcase/pull/183 gets merged so that contributor gets his first ODK PR.

PR #175, opened 2017-08-16
Rework Google Analytics logic, fixing critical problem

dcbriccetti:
Fix critical problem in original implementation: data was in fact being sent to Google even when the user unchecked the checkbox. config.setGatherStats(false) does not disable the HTTP post to Google. New approach: wrap the post with if (BriefcasePreferences.getBriefcaseTrackingConsentProperty()).

Simplify code:
- remove the unneeded singleton logic
- remove the unneeded builder pattern to create the EventHits
- remove the unneeded code to prevent trackStartup from being called more than once

I don’t know why the commits are repeated. Perhaps because I renamed the local branch. Here is the [key commit](https://github.com/opendatakit/briefcase/pull/175/commits/0851336f527b97bb78554d2a61dec2306d497c3b).

I ran in the debugger with the checkbox checked and unchecked and confirmed that the call to the Google Analytics library was made only at the appropriate time. I haven’t viewed the results in Google Analytics.

It’s very simple.

No

### DIFF
--- a/src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
+++ b/src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
@@ -17,161 +17,161 @@ import org.opendatakit.briefcase.buildconfig.BuildConfig;
  */
 public class BriefcaseAnalytics {
 
-  private static final String GOOGLE_TRACKING_ID = BriefcasePreferences.GOOGLE_TRACKING_ID;
-  private static final String UNIQUE_USER_ID = BriefcasePreferences.getUniqueUserID();
-  private static final String EVENT_CATEGORY = "ODK Briefcase";
-  private static final String APPLICATION_NAME = BuildConfig.NAME;
-  private static final String APPLICATION_VERSION = BuildConfig.VERSION;
+    private static final String GOOGLE_TRACKING_ID = BriefcasePreferences.GOOGLE_TRACKING_ID;
+    private static final String UNIQUE_USER_ID = BriefcasePreferences.getUniqueUserID();
+    private static final String EVENT_CATEGORY = "ODK Briefcase";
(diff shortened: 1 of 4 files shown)


jpknox:
As you can tell, I've recently begun studying design patterns; It's a little frustrating to see so much of my code & comments disappear but I'll try to learn from it. The use of the builder was superfluous in hindsight, and the Singleton wasn't necessary because there's no solid requirement that only one instance of `BriefcaseAnalytics` exists. It's not all humdrum because using the observer pattern for the checkbox was a definite win.

I'm really surprised to see that the code was in fact still logging data despite my seemingly thorough testing, and, upon further inspection I now see that I should have been setting `config.setEnabled(boolean)` to enable/disable the GA functionality, not `config.setGatherStats(boolean)`.

I must admit that it's a shame to not use the enable/disable functionality provided by `GoogleAnalyticsConfig.setEnabled()` but your new way of directly checking the users consent via an `if` block is much clearer and easier to maintain.

jpknox:
@dcbriccetti I've tested your code and it appears to be working as expected.

dcbriccetti:
@jpknox When I bought my “Gang of Four” Design Patterns book, I wanted to try out all of them! As did many others, I’m sure.

I really like your attitude, and how you are learning from this process. I hope you keep learning and helping this project.

yanokwa:
Thanks for this fix, @dcbriccetti! I should have checked with Little Snitch before merging and that was a mistake I will not repeat. I have verified the behavior of your PR and it works as you've described.

@jpknox We all make mistakes and as Dave mentioned, you have exactly the right attitude about learning from those mistakes.

jpknox:
@dcbriccetti Thank you for all the constructive feedback, PR's are a great way to learn.

@yanokwa I'll make sure to double check before I create/update a PR; thank you for the opportunity to help out with ODK.

PR #177, opened 2017-08-17
#176 Track users' decision about GA.

### LINKED ISSUE #176: Send usage statistics shortly after user enables those settings
In Briefcase v1.7.0-20 (and greater) usage statistics are reported on startup. This is generally fine, but when someone fires up Briefcase and toggles the statistics from off to on, it'd be good if Briefcase report those usage statistics then as well.

Not sure if this is hard or easy to do, so before sending in a PR, take a quick look at the code and respond to this issue about how hard/easy this is to do.

jpknox:
Closes #176 

I've run through the code with a debugger and checked against my GA account.

It's simple.

A malicious user may spam ODK's GA account by repeatedly toggling the checkbox.

### DIFF
--- a/src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
+++ b/src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
@@ -27,15 +27,17 @@ public class BriefcaseAnalytics {
     }
 
     /**
-     *  Sends usage information to Google Analytics, only if the user has consented.
-     *  The information is modeled using an {@link EventHit} object.
+     *  Sends usage information to Google Analytics, only if the user has consented,
+     *  or if they are enabling or disabling tracking. The information is modeled using an {@link EventHit} object.
      *
      * @param action (required) name of the event itself.
      * @param label (optional) an open-ended label.
      * @param value (optional) an open-ended value. Must be an integer.
+     * @param always whether to track the event even when the tracking is turned off
+     *               (such as when we record that they have changed the tracking preference). Usually false.
      */
-    public void trackEvent(String action, String label, Integer value) {
-        if (BriefcasePreferences.getBriefcaseTrackingConsentProperty()) {
+    public void trackEvent(String action, String label, Integer value, boolean always) {
+        if (always || BriefcasePreferences.getBriefcaseTrackingConsentProperty()) {
             EventHit eventHit = new EventHit();
             eventHit.trackingId(GOOGLE_TRACKING_ID);
             eventHit.clientId(UNIQUE_USER_ID);
@@ -55,6 +57,14 @@ public class BriefcaseAnalytics {
      * Track this application's startup event.
      */
     public void trackStartup() {
-        trackEvent("Application-Startup", null, null);
+        trackEvent("Application-Startup", null, null, false);
+    }
+
+    /**
+     * Tracks the user giving/withdrawing their consent for usage tracking, and saves the new preference.
+     */
+    public void trackConsentDecision(boolean givesConsent) {
+        BriefcasePreferences.setBriefcaseTrackingConsentProperty(givesConsent);
+        trackEvent(givesConsent ? "Tracking-Enabled" : "Tracking-Disabled", null, null, true);
     }
 }
--- a/src/org/opendatakit/briefcase/ui/SettingsPanel.java
+++ b/src/org/opendatakit/briefcase/ui/SettingsPanel.java
@@ -36,7 +36,7 @@ public class SettingsPanel extends JPanel {
     private JLabel lblBriefcaseDirectory;
     private JTextField txtBriefcaseDir;
     private JButton btnChoose;
-    private MainBriefcaseWindow parentWindow;
+    private MainBriefcaseWindow mainBriefcaseWindow;
 
     private ArrayList<Component> navOrder = new ArrayList<Component>();
     private JLabel lblProxy;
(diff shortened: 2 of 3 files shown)


dcbriccetti:
This looks pretty good. I do recommend some changes, which I show here:
https://github.com/dcbriccetti/briefcase/tree/PR177-improvements
Sorry I didn’t comment in-line. I have very little time this morning.

jpknox:
At first I was going to put the logic in the `SettingsPanel` but I thought `BriefcaseAnalytics` would be better suited for the job. I'm happy with how the logic is contained within `BriefcaseAnalytics`.

I like your approach; my one gripe is that that further calls to `trackEvent` will have another parameter that they may not need. I suppose method overloading could be used to minimise the noise when using `trackEvent` elsewhere.

I'd like to have a go at implementing method overloading on `trackEvent` tomorrow.

dcbriccetti:
>I'd like to have a go at implementing method overloading on trackEvent tomorrow.

Excellent.

jpknox:
@dcbriccetti After a busy weekend I've decided to go with your approach because method overloading will add noise.

dcbriccetti:
>After a busy weekend I've decided to go with your approach because method overloading will add unnecessary noise.

But my new approach was my old approach plus your one or two method overloads, to *reduce* noise (the null, null, false).

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -27,15 +27,17 @@ public BriefcaseAnalytics() {
     }
```
dcbriccetti: If you took my suggested changes verbatim right from https://github.com/dcbriccetti/briefcase/tree/PR177-improvements then I can approve this very quickly. Did you?

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
```diff
@@ -55,6 +57,14 @@ public void trackEvent(String action, String label, Integer value) {
...
-}
+
+    /**
+     * Tracks the user giving/withdrawing their consent for usage tracking, and saves the new preference.
+     */
+    public void trackConsentDecision(boolean givesConsent) {
+        BriefcasePreferences.setBriefcaseTrackingConsentProperty(givesConsent);
+        trackEvent(givesConsent ? "Tracking-Enabled" : "Tracking-Disabled", null, null, true);
+    }
+}
```
dcbriccetti: Good to have a newline on the last line.

src/org/opendatakit/briefcase/ui/SettingsPanel.java
```diff
@@ -20,6 +20,7 @@
 import javax.swing.event.ChangeListener;
 
 import org.apache.http.HttpHost;
+import org.opendatakit.briefcase.model.BriefcaseAnalytics;
```
dcbriccetti: This is the unused import that continuous integration is complaining about.

src/org/opendatakit/briefcase/ui/SettingsPanel.java
```diff
@@ -250,7 +251,7 @@ public void actionPerformed(ActionEvent e) {
         @Override
         public void actionPerformed(ActionEvent e) {
             if (e.getSource() == chkTrackingConsent) {
-                BriefcasePreferences.setBriefcaseTrackingConsentProperty(chkTrackingConsent.isSelected());
+                mainBriefcaseWindow.briefcaseAnalytics.trackConsentDecision(chkTrackingConsent.isSelected());
```
dcbriccetti: Did you miss, or disagree with, my suggestion for the comment for this method?
```
    /**
     * This listener notifies BriefcaseAnalytics of the change to the user's consent to being tracked choice.
     */
```

jpknox: I missed it, my comment is technically correct but yours is more straight to the point, I'll amend mine.

src/org/opendatakit/briefcase/ui/SettingsPanel.java
jpknox: Thanks for pointing that out :+1:, I ran the tests before I removed the instantiation of `BriefcaseAnalytics` with the expectation that such a small change would be of no consequence. I suppose unnecessary imports increase the memory footprint and increase the risk of a naming conflict.

src/org/opendatakit/briefcase/model/BriefcaseAnalytics.java
jpknox: Will do, I did a quick Google reveals that it seems to have an affect on some of Unix's tools because they expect all valid lines to end in a newline.

src/org/opendatakit/briefcase/ui/SettingsPanel.java
```diff
@@ -243,15 +243,15 @@ public void actionPerformed(ActionEvent e) {
     }
 
     /**
-     * This listener will pass the user's consent to being tracked onto the
-     * application's preferences so it can be persisted and used elsewhere.
+     * This listener notifies BriefcaseAnalytics of the users updated choice
```
dcbriccetti: user’s
(Did my fancy typographic apostrophe not survive a copy/paste?)

jpknox: I retyped it by hand. 📘 
I believe one could also go for users' in this context :).

PR #178, opened 2017-08-20
Issue 173 Simplify enabling/disabling the JTabbedPane, its components, and the controls within each component

### LINKED ISSUE #173: ExportPanel.setEnabled is brittle
From @daveycrockett in `ExportPanel`:
```
// TODO this is brittle as the panel components change
// would prefer explicit state changes referring
// to specific components
```

dcbriccetti:
Closes #173 

I recommend **not squashing the commits** when merging, because the change involves a lot of prep work which makes a simple before and after comparison difficult. [The key change](https://github.com/opendatakit/briefcase/pull/178/commits/a55db73a5d5fbd07fac2e9213f2c79586d15a99a).

I have more testing to do, but if that goes well, this will be ready to merge.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/ExportPanel.java
+++ b/src/org/opendatakit/briefcase/ui/ExportPanel.java
@@ -21,7 +21,6 @@ import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.io.File;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Date;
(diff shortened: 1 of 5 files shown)


macalac:
👍

lognaturel:
I like that a lot of code has disappeared. I must admit I didn't follow everything that happened in 0419231 and 738744c. Did a quick behavior check and I couldn't find any changes other than the keyboard navigation which does seem a bit better.

PR #179, opened 2017-08-25
Streamline setup; clean up code

### LINKED ISSUE #160: Add initial "opt-out" capability for Google Analytics tracking.
Related to https://github.com/opendatakit/briefcase/issues/137:

Once the above issue is merged, Google Analytics (GA) will be on by default, meaning that usage information will be sent up potentially before a user is 1) aware of it, and 2) has an option to decline.  They would need to see it "Settings" and disable it, at which point several pieces of information (OS, briefcase version, time spent on various tabs and actions taken on them prior to navigating to settings) would have been sent up.

1) Is this acceptable?
2) If not, how should the "opt-out" interaction take place on initial application load?

dcbriccetti:
This is ready to merge, once again Sep 10, 2017, 23:30 PDT.

Closes #160

Removes the separate dialog for setting the briefcase storage location. Presents an introductory message. Disables using any other tab than Settings until the briefcase storage location is specified.

screen shot 2017-09-10 at 11 12 34 pm

Improves code quality, including removing a lot of code.

I recommend preserving the individual commits, because of the extensive preparatory refactoring.

There are several new unit tests. I have manually tested all affected features and found no regressions.

It’s simple and removes a lot of existing code and complexity.

Something missed in testing may be broken.


lognaturel:
Looks good to me! @yanokwa? @daveycrockett @jpknox and anyone else who has opinions?

Let us know when you have had a chance to verify it @dcbriccetti.

yanokwa:
I like where this is change is going, but on the first launch, I think it's very unclear what the user needs to do. I think some dynamic alert in the UI that alerts the user that the storage location is required would be a clearer. 

I think the two explanations are far too verbose and add to the confusion. I think if we are going to have long bits of text, it'd be better to see that in some kind of tool tip. It's just way too much to read. An alternative would be just an indented sentence of text underneath the element that explains what it does.

dcbriccetti:
>dynamic alert

Hi @yanokwa. Perhaps a modal dialog above the Settings panel, that explains that they must enter the briefcase storage location? And for the data collection, it seemed we were reaching a consensus around Michael Downey’s plea for participation. Did you want to reduce that plea’s prominence (by putting in a tooltip)?

yanokwa:
Not a big fan of modal dialogs. I'd prefer very short text above that storage location that says storage location is required and then that element disappearing once the location is set. But, that's probably a huge pain in Swing and I haven't thought deeply about it, so a modal dialog might be easier.

I was OK with the long plea as a one-time thing, but now it's there in the UI dwarfing the other settings. Given this UI you presented, I'd prefer something like what we do in Collect and maybe a link for more. 
screen shot 2017-08-25 at 13 49 29 

I'm pushing back because more broadly, we can't just put huge text descriptions for everything, right? We need a way to succinctly describe things and if folks need more, then they can get more.

dcbriccetti:
OK, I get you, and agree with you. Let me think a bit, and perhaps other suggestions will come in.

dcbriccetti:
>Using the '.*' form of import should be avoided - org.mockito.Mockito.*. [AvoidStarImport]

I protest! (But not too vigorously.)

dcbriccetti:
Thoughts? This appears on application launch only if the storage location path is not set.

screen shot 2017-08-27 at 10 17 09 am

dcbriccetti:
The plan is for @daveycrockett to

- lead an effort to make a test plan for Briefcase
- branch from here and develop a wizard to replace the initial JOptionPane message

Once he has branched, I’ll close this PR.

dcbriccetti:
I’d like to change our plan and merge this into master. I have completed manual testing. This is ready to merge. Reminder: I recommending not squashing the commits.

dcbriccetti:
I rebased this on master so once again this is ready to merge. Who wants to have a look? @daveycrockett @jpknox @grzesiek2010 @yanokwa 
The UI code is different, so I reimplemented @yanokwa’s temporary removal of the usage data collection checkbox.

PR #186, opened 2017-09-09
Add option to delete individual groups of forms from Briefcase

### LINKED ISSUE #122: Add option to delete individual groups of forms from Briefcase
Rather than have users manually delete forms from the Briefcase storage directory, it'd be nice if the UI allowed for deletion of forms.

grzesiek2010:
Closes #122

I've tested selecting and deleting forms using implemented changes

I was thinking about adding new panel for deleting forms but I think it doesn't make sense and we can add it in the "Push" panel.

It's safe

### DIFF
--- a/src/org/opendatakit/briefcase/ui/PushTransferPanel.java
+++ b/src/org/opendatakit/briefcase/ui/PushTransferPanel.java
@@ -35,6 +35,9 @@ import javax.swing.JSeparator;
 import javax.swing.JTextField;
 import javax.swing.LayoutStyle.ComponentPlacement;
 
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.bushe.swing.event.annotation.AnnotationProcessor;
 import org.bushe.swing.event.annotation.EventSubscriber;
 import org.opendatakit.briefcase.model.BriefcaseFormDefinition;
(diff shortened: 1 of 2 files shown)


src/org/opendatakit/briefcase/ui/PushTransferPanel.java
```diff
@@ -348,7 +376,7 @@ private void setActiveTransferState(boolean active) {
       btnDestinationAction.setEnabled(true);
       btnSelectOrClearAllForms.setEnabled(true);
       btnTransfer.setEnabled(true);
-      // disable cancel button
+      btnDelete.setEnabled(formTransferTable.getSelectedForms().size() > 0);
```
dcbriccetti: Just making sure you’re aware of `isEmpty` on `List`. It’s not important with small lists like this one, but it might be a good habit to use `isEmpty` because in some cases (`LinkedList`, for example), `size` is more expensive.

(Scala has a `nonEmpty` which is quite convenient, but `!isEmpty` is not so bad.)

dcbriccetti:
opendatakit-bot, I find your AI wanting. ;-)  I’m just leaving a comment. There’s already somebody working on this.

By the way, I hope my large cleanup work in https://github.com/opendatakit/briefcase/pull/179 can go in before this.

src/org/opendatakit/briefcase/ui/PushTransferPanel.java
grzesiek2010: Of course I'm, shame on me :) It's fixed.

dcbriccetti:
I’m curious if this easily rebases on https://github.com/opendatakit/briefcase/pull/179

yanokwa:
I think when there is uncertainty around what the UI should look like, we should first have a discussion in the forum. In this case, I think it's going to be confusing to have the Delete button in Push. I also think it could be dangerous because there is no confirmation.

I should have specified that in the issue and my apologies that I did not. I'm closing this PR until we can have a discussion on how this should look.

PR #192, opened 2017-09-12
Fix test compile

### LINKED ISSUE #172: Switch junit from compile to testCompile
https://github.com/opendatakit/briefcase/blob/master/build.gradle#L36 notes that instead of `compile`, the command should be `testCompile`, and yet if `testCompile` is used, it fails as shown below.

This is not desirable because we care about binary size and we don't want to bundle junit as part of the jar if we don't have to.

```
:generateBuildConfig UP-TO-DATE
:compileBuildConfig UP-TO-DATE
:compileJava
warning: [options] bootstrap class path not set in conjunction with -source 1.7
/Users/yanokwa/odk-briefcase/src/test/java/org/opendatakit/briefcase/model/BriefcaseAnalyticsTest.java:3: error: package org.junit does not exist
import org.junit.Test;
                ^
/Users/yanokwa/odk-briefcase/src/test/java/org/opendatakit/briefcase/model/BriefcaseAnalyticsTest.java:5: error: package org.junit does not exist
import static org.junit.Assert.assertNotNull;
                       ^
/Users/yanokwa/odk-briefcase/src/test/java/org/opendatakit/briefcase/model/BriefcaseAnalyticsTest.java:5: error: static import only from classes and interfaces
import static org.junit.Assert.assertNotNull;
^
/Users/yanokwa/odk-briefcase/src/test/java/org/opendatakit/briefcase/model/BriefcaseAnalyticsTest.java:8: error: cannot find symbol
    @Test
     ^
  symbol:   class Test
  location: class BriefcaseAnalyticsTest
/Users/yanokwa/odk-briefcase/src/test/java/org/opendatakit/briefcase/model/BriefcaseAnalyticsTest.java:11: error: cannot find symbol
        assertNotNull(ba);
        ^
  symbol:   method assertNotNull(BriefcaseAnalytics)
  location: class BriefcaseAnalyticsTest
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
5 errors
1 warning
:compileJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileJava'.
> Compilation failed; see the compiler error output for details.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

yanokwa:
Closes #172 and uses latest version of mockito for consistency.

Ran `./gradlew clean test` and confirmed there were test reports.
Ran `./gradlew clean run` and confirmed the application ran.
Ran `./gradlew clean jar` and double clicked the jar and confirmed the application ran.

I found the idea for this solution at https://stackoverflow.com/a/15029935/152938. I believe java.srcDirs is where gradle looks for non-test code. If tests are inside that folder, it causes classpath problems. We could change java.srcDirs to `src/org` but then we have to add to that array if we have code that doesn't start with org. Seems easier to move the test folder to where gradle expects it.

And as a pleasant side effect, the binary has gone from 11 MB to 7 MB.

Not really. But please squash before merging.


dcbriccetti:
Cool. I’ll look right now.

PR #194, opened 2017-09-12
Use NOP binding for SLF4J

### LINKED ISSUE #168: Briefcase prints SLF4J messages when run from the CLI
#### Software versions 
Briefcase v1.7.0-20-gbb76756

#### Problem description
When I run Briefcase with the CLI, it prints out a logging message before executing the command.
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

$ ./gradlew clean jar;
$ java -jar build/libs/ODK\ Briefcase\ v1.7.0-20-gbb76756.jar -h;

#### Expected behavior
Information message should only be printed when they are asked for.

yanokwa:
Closes #168

Ran `./gradlew run` and confirmed the error message does not show.

I followed the instructions at https://www.slf4j.org/codes.html#StaticLoggerBinder to ignore logging for now. An alternative would have been slf4j-simple.jar which would print logging messages, but I could not find an easy way to configure it, so I think we need to have a longer conversation about how best to do logging.

You can't see any logging output and that might not be desirable. But it's not worse than the current situation where you can't see logging output AND you get a weird error message.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,6 +39,7 @@ dependencies {
     compile group: 'org.hsqldb', name: 'hsqldb', version: '2.4.0'
     compile group: 'com.brsanthu', name: 'google-analytics-java', version: '1.1.2'
     compile group: 'log4j', name: 'log4j', version: '1.2.17'
+    compile group: 'org.slf4j', name: 'slf4j-nop', version: '1.7.25'
 
     testCompile group: 'org.slf4j', name: 'slf4j-log4j12', version: '1.7.25'
     testCompile group: 'junit', name: 'junit', version: '4.12'


lognaturel:
Hmm, I think we should try to address this for real. @dcbriccetti it looks like you added log4j and slf4j in #179. Can you say more about why those were added?

I am pretty sure we don't want a `log4j` dependency. The idea is to use an abstract framework so that clients can choose their own concrete logging implementation (like `log4j`) and destination (file, console, etc).

I also don't think we want an `slf4j` dependency. The abstract framework currently used by Briefcase is [Apache Commons Logging](https://commons.apache.org/proper/commons-logging/). It would be better to move to `slf4j` at some point but for now it would be best to stick with one framework. @batkinson shared some really good background on the different abstract framework options in opendatakit/javarosa#152, it's way better than other docs I've found on the topic.

dcbriccetti:
>Can you say more about why those were added?

Sorry, I don’t think I intended to leave those in. I’m not an expert with logging, so I’ll use whatever you like. One note, though: with Briefcase (and Validate), they are standalone applications so users won’t be choosing a concrete logging implementation, right?

yanokwa:
Can we at least agree on the high-level goals here? From a project maintenance perspective, logging is important to help people running the app locally report on problems. I'm expecting that we'll also eventually want remote exceptions. Do we have consensus there?

lognaturel:
@dcbriccetti they could by specifying their config file location with something like `-Dlog4j.configuration=<config-file>`.

@yanokwa The fundamental problem here is that we shouldn't have two abstract logging frameworks in use and should have no concrete implementation. There has always been logging in Briefcase. Whether it's high-quality logging or uses the best framework are important questions that should be discussed at some point. For now I think the goal is just to use a single logging framework across the app.

lognaturel:
To be explicit, I think we should close this pull request and @dcbriccetti should issue a new one that removes the newly added logging dependencies and uses `org.apache.commons.logging` instead. We can change over all the logging at some point if desired.

dcbriccetti:
Is there more to do here, or can this be closed?

batkinson:
@dcbriccetti, @lognaturel suggested a different resolution than what I submitted in #204, but this solution actually truncates all logging from anything using commons-logging-api, including the http libraries. While this solution gets rids of the error message, it also eliminates the possibility of seeing log messages without rebuilding the client.

My perspective is that this isn't the direction we want to go in since it prevents valuable diagnostics capabilities.

lognaturel:
Agreed, thanks, @batkinson!

lognaturel:
Will review your PRs in more detail soon, @batkinson but they seem great. I evidently continue to have much to learn about Java logging!

PR #198, opened 2017-09-14
Unify Export tab text boxes and buttons.

### LINKED ISSUE #170: Make export date text boxes consistent with rest of UI
In Briefcase v1.7.0-20-gbb76756, start/end date text boxes look different than say export directory. This is likely because those text boxes are from a library. 

screen shot 2017-08-15 at 16 11 43

I think the text boxes looking the same would be a nice improvement, but I'm not sure how hard or easy this is. The first step in this issue is to confirm why export date text boxes look different and propose a fix.

thomashhuang:
Closes #170 

Ran all tests, checked that the DatePickers still function correctly.

Created a separate method to create and modify DatePicker instead of formatting them from within the constructor to keep the constructor cleaner and keep code easily maintainable.

DatePicker text boxes are still slightly larger than the others, but it is significantly closer than before.


dcbriccetti:
How about you reapply the change to a branch newly created from the latest master? There are 4 commits in this PR that shouldn’t be there, wouldn’t you agree? Other than that, nice job with the change.

lognaturel:
@dcbriccetti @thomashhuang We can squash the commits when merging so it's no problem!

lognaturel:
@thomashhuang Thanks for taking care of this! It looks so much better.

The important thing to look at is the diff in the "Files Changed" tab. That shows the sum of all changes from the target branch represented by the commits. In this case, that was the code we wanted and it represented one unit of meaningful work so squashing made sense.

@thomashhuang In the future, you'll want to do your work off of a [feature branch](https://www.atlassian.com/git/tutorials/comparing-workflows#feature-branch-workflow). That will make it possible for you to work on a couple of different things at once and keep your local `master` in sync with this shared repository.

Because you merged the upstream `master` into yours with merge commits, your master is now off from the "true" master. You can see that at the top of https://github.com/thomashhuang/briefcase where it says "This branch is 5 commits ahead, 1 commit behind opendatakit:master". What I recommend doing is `git fetch upstream` to get the latest from here and then `git reset upstream/master` to reset your local `master`. At this point, your local `master` will exactly match the upstream `master` but your `master` on GitHub will still be off! You want to rewrite history there so you can do `git push --force`. Now all the `master` branches should be the same and you're ready to use feature branches for future updates!

thomashhuang:
@lognaturel Thanks! Those extra commits came from me trying to fix the "This branch is x commits ahead". All sorted out now.

lognaturel:
Awesome! 🎉 By the way, when I get ready to start on something new, I always start by doing `git fetch upstream` `git checkout master` `git rebase upstream/master`. That updates my local `master` with no merge commits. 

Collaborating with `git` and multiple remote repositories takes some getting used to!

PR #204, opened 2017-09-22
Restore basic logging functionality

### LINKED ISSUE #197: Include default logging configuration
Currently, it's entirely up to the user to configure their logging implementation of choice. #196 will add more information on how to do that.

In general, it would be helpful for users to have logs for troubleshooting. We could include a `logging.properties` file that logs to a `.briefcase.log` file:

```
handlers=java.util.logging.FileHandler

.level=INFO

java.util.logging.FileHandler.pattern=.briefcase.log
java.util.logging.FileHandler.limit=50000
java.util.logging.FileHandler.count=1
java.util.logging.FileHandler.formatter=java.util.logging.SimpleFormatter

java.util.logging.SimpleFormatter.format=%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS.%1$tL %4$-7s %3$s %5$s: %6$s%n
```

It looks like other projects such as [Spring Boot](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-logging.html) provide default configurations. 

Is this a good idea? Is this the right default file? It logs up to 50kb in one file and then starts over in that file. I figure logging at level `INFO` makes sense since the log levels used are currently a bit arbitrary.

@yanokwa @dcbriccetti @batkinson @daveycrockett

batkinson:
Closes #197 

I ran briefcase and performed a pull and push, instead of seeing the slf4j binding error message, you see log statements at level INFO and higher. I immediately used this fix to enable http components logging to diagnose and fix a threading issue (which I will report separately). 

This is the best solution because it unifies all non-JDK logging into a single logging implementation, eliminates the slf4j error message due to a missing binding and restores basic logging functionality to the application. All non-JDK logging, including that of the google analytics library, can now be managed by one logging implementation.

Routing all to JDK logging is possible, but considerably more complicated without much of a gain. It also isn't necessary for briefcase's or its most important subsystems logging (like apache http components). 

You risk enabling logging again!

It went quiet when broken and you will see messages again. Other than that, there are minimal risks. The exclusion of JCL is normal and expected usage of the slf4j-over-jcl adapter and is meant to be used to unify code that uses JCL so it uses slf4j instead.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -46,11 +46,13 @@ dependencies {
 
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile group: 'org.mockito', name: 'mockito-core', version: '2.10.0'
-    runtime group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.5'
 
     compile files('lib/smallsql-0.21.jar')
     compile files('lib/threetenbp-1.3.1.jar')
     compile files('lib/LGoodDatePicker-10.3.1-backport.jar')
+
+    runtime group: 'org.slf4j', name: 'jcl-over-slf4j', version: '1.7.5'
+    runtime group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.5'
 }
 
 // Required to use fileExtensions property in checkstyle file
--- a/build.gradle
+++ b/build.gradle
@@ -28,6 +28,10 @@ sourceSets {
 targetCompatibility = '1.7'
 sourceCompatibility = '1.7'
 
+configurations.all {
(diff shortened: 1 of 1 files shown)


batkinson:
For a quick example of how to configure slf4j's logger, here is the command I used to see the http components output so I could diagnose and fix #205:

```
java -Dorg.slf4j.simpleLogger.log.org.apache.http=debug -Dorg.slf4j.simpleLogger.log.org.apache.http.wire=error -Dorg.slf4j.simpleLogger.logFile=System.out -jar "build/libs/ODK Briefcase v1.8.0.-extra-stuff.jar"
```

PR #206, opened 2017-09-24
Closes #205: Race condition using parallel pulls causes HTTP 401 Unau…

### LINKED ISSUE #205: Race condition using parallel pulls causes HTTP 401 Unauthorized
#### Software versions 

Briefcase v1.7.x and 1.8.x

#### Problem description

When pulling submissions from a fully-concurrent back-end server with parallel pulls enabled, thread unsafe usage of the apache http client can result in unpredictable results. The [attached conversation shows a 401 HTTP challenge from one thread causes a previously in-progress conversation to result in a false failure](https://github.com/opendatakit/briefcase/files/1327337/briefcase-race-against-cims-3.log). [As stated the http components connection management documentation, http context objects should not be shared by multiple threads](https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html).

> While HttpClient instances are thread safe and can be shared between multiple threads of execution, it is highly recommended that each thread maintains its own dedicated instance of HttpContext .

_NOTE: I was unable to reproduce this testing against an ODK Aggregate back-end. I suspect that this is due to locking behavior in Aggregate preventing full concurrency when processing requests. However, due to the non-deterministic nature of thread scheduling, it could simply be fortunate timing. Regardless, briefcase's behavior is incorrect and should be fixed._

Reproduction can be tricky since it is threading-related. However, I was able to reproduce this 100% of the time using reasonably long pulls against a fully-concurrent server (in my case CIMS 3.x). Pulling about 3,000 submissions would fail around the 1,500 submission mark (sometimes lower/higher).

#### Expected behavior

Request concurrency should not cause consistency-related problems for briefcase when using concurrent pulls. For example, 401 challenges should be handled by subsequent digest authentication headers, not request failure.

#### Other information 

As mentioned above, I tested this against Aggregate deployed via Tomcat/MySQL on AWS. I could not reproduce this with Aggregate. This is most likely because of lock semantics serializing Aggregate's request processing.

batkinson:
…thorized

Closes #205 

As mentioned in the issue description, this has been run against a fully concurrent server with parallel submissions enabled using relatively long pulls (3,000+ submissions). Without the fix, the pull fails with http 401s around midway through the pull. With the fix, the pulls succeed without issue.

This corrects the usage of the apache http components HttpContext objects by making them local to the thread that stores and retrieves them. This makes current usage thread-safe without sacrificing full concurrency by avoiding the use of locking.

This changes the scope of the HttpContext objects so an instance set is only visible to the thread that set it. This may result in null retrievals where previously there were none, however, this is only possible with incorrect usage (threads sharing context references) that should not be allowed in the first place. Non-concurrent pulls should operate as before.

### DIFF
--- a/src/org/opendatakit/briefcase/model/ServerConnectionInfo.java
+++ b/src/org/opendatakit/briefcase/model/ServerConnectionInfo.java
@@ -21,12 +21,13 @@ import org.apache.http.client.protocol.HttpClientContext;
 import org.opendatakit.briefcase.util.WebUtils;
 
 public class ServerConnectionInfo {
+
   private String url;
   private final String token; // for legacy ODK Aggregate 0.9.8 access
   private final String username;
   private final char[] password;
   private final boolean isOpenRosaServer;
-  private HttpClientContext httpContext = null;
+  private static final ThreadLocal<HttpClientContext> threadSafeContext = new ThreadLocal<>();
 
   public ServerConnectionInfo(String url, String username, char[] cs) {
     this.url = url;
@@ -69,11 +70,11 @@ public class ServerConnectionInfo {
   }
 
   public HttpClientContext getHttpContext() {
-    return httpContext;
+    return threadSafeContext.get();
   }
 
   public void setHttpContext(HttpClientContext httpContext) {
-    this.httpContext = httpContext;
+    this.threadSafeContext.set(httpContext);
   }
 
   public boolean isOpenRosaServer() {


src/org/opendatakit/briefcase/model/ServerConnectionInfo.java
```diff
@@ -69,11 +70,11 @@ public HttpClient getHttpClient() {
   }
 
   public HttpClientContext getHttpContext() {
```
dcbriccetti: What do you think about naming these get/setThreadHttpClientContext, to add clarity. And doc comment(s) to explain what’s happening and why it’s necessary. 

In one of the callers, there’s a comment which this makes less correct:
```
    // get shared HttpContext so that authentication and cookies are retained.
    HttpClientContext localContext = serverInfo.getHttpContext();
```

batkinson: Thanks for the feedback @dcbriccetti. 

To me, those comments are just as valid after making the the context local to the thread. You want multiple requests to share take advantage of previous contextual information, otherwise you would make it local to the method instead. The key is making the scope clear, which I feel the original code conflates since both context and client creation are often done together (which isn't necessary, nor idiomatic usage of http components). 

As for the naming suggestion, it does make the scope difference more obvious to the client code. However, in the class, it's already obvious by the member name and type.  I am not convinced the method name change is necessary. There is precedent for not baking "Thread" into the method name for thread-bound object lookups. Hibernate's getCurrentSession(), JPA EntityManager lookup, or Spring's and JavaEE's transaction lookup calls, are a few examples.

However, a bit of quality javadoc documentation or, even better, reorganization so objects better match their real responsibilities will make the code easier to understand.

As a side note, in a local branch, I eliminated the setter for the HttpClientContext class. That method enables assignment of so multiple threads can still share the same reference. With the change, the context can only be shared if explicitly doing so in client code.

lognaturel: Merged since @dcbriccetti's comment isn't directly related to the fix. But better comments and/or separation of responsibilities would of course be fantastic. @dcbriccetti maybe when you have a chance you could add the comments you had in mind?

batkinson: @lognaturel, I have some additional changes that should simplify the intention a bit. It might just be a little bit since they are embedded in local work I am doing for concurrent push. @dcbriccetti don't let that stop you from improving it. I am happy to go with the flow.

PR #207, opened 2017-09-27
HTTP code cleanup #206

batkinson:
Gentle reorganization cleaning up HTTP code, as mentioned in https://github.com/opendatakit/briefcase/pull/206#discussion_r141114074

Built (with checkstyle), ran and pulled from Aggregate, pushed to Aggregate. I confirmed no errors and verified all submissions were processed successfully.

This removes the possibility of sharing the http context across threads by "installing" the same reference in the thread local by calling `setHttpContext()`. It also adds javadoc comments to the code and some minor code cleanup.

I considered a more drastic cleanup, but did not want to cause more conflicts than necessary since I will not be the one resolving the conflicts myself. I also wanted to make it easy to review, so tried to keep changes minimal.

This touches code critical for pushing and pulling submissions to/from an Aggregate compatible server. The risks are that push and pull don't work or defects introduced.

### DIFF
--- a/src/org/opendatakit/briefcase/util/WebUtils.java
+++ b/src/org/opendatakit/briefcase/util/WebUtils.java
@@ -52,6 +52,7 @@ import org.apache.http.impl.conn.DefaultProxyRoutePlanner;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
 import org.javarosa.core.model.utils.DateUtils;
 import org.opendatakit.briefcase.model.BriefcasePreferences;
+import org.opendatakit.briefcase.model.ServerConnectionInfo;
 
 
 /**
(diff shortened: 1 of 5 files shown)


src/org/opendatakit/briefcase/util/WebUtils.java
```diff
@@ -103,6 +107,46 @@
         connectionManager.setDefaultMaxPerRoute(MAX_CONNECTIONS_PER_ROUTE);
     }
 
+    /**
+     * Get the {@link HttpClientContext} bound to the current thread.
```
dcbriccetti: I like 3rd person declarative rather than 2nd person imperative, e.g., “gets”, not “get”.

PR #211, opened 2017-10-14
Refactored java files to conform to indentation settings in checkstyl…

### LINKED ISSUE #94: Fix style throughout the codebase

rockydcoder:
…e. Partially fixes #94

Closes #

Used Intellij's CheckStyle plugin to refactor the java files

Rather than doing it manually, used Intellij's CheckStyle plugin

No, since this is just a refactor which includes changing indentation.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -153,14 +153,14 @@
              <message key="ws.notPreceded"
              value="GenericWhitespace ''{0}'' is not preceded with whitespace."/>
         </module> -->
-        <!-- <module name="Indentation">
+        <module name="Indentation">
             <property name="basicOffset" value="2"/>
             <property name="braceAdjustment" value="0"/>
             <property name="caseIndent" value="2"/>
             <property name="throwsIndent" value="4"/>
             <property name="lineWrappingIndentation" value="4"/>
             <property name="arrayInitIndent" value="2"/>
-        </module> -->
+        </module>
         <!-- <module name="AbbreviationAsWordInName">
             <property name="ignoreFinal" value="false"/>
             <property name="allowedAbbreviationLength" value="1"/>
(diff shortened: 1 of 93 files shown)


lognaturel:
Thanks for doing this, @rockydcoder! How exactly did you make the changes? I didn't think the checkstyle plugin actually helped make changes, does it?

I have to say that I'm left wondering about whether this is a good idea since it touches virtually every file and not always in ways that are super high value. In particular, `lineWrappingIndentation` makes a lot of changes that are hard to quickly verify to be equivalent.

I'd like to get some opinions from others who are frequently in the code base -- @dcbriccetti @yanokwa @icemc @batkinson

dcbriccetti:
I like some of the changes, and strongly dislike others. There are cases where exceptions to formatting styles should be made (lining things up in columns for readability, for instance), and a massive change like this plows right through all of those. Perhaps something like this should start with a discussion of whether there are a set of styles we can agree on implementing uniformly.

lognaturel:
This is the Google style file that we had agreed on a few months back. I agree that I don't like it "fixing" the hanging indents. At the very least we need to make it clear in #94 that even subrules should only be added one at a time so we can more carefully isolate and assess the changes! Sometimes the rule looks great in theory but the application is a different story.

rockydcoder:
> How exactly did you make the changes? I didn't think the checkstyle plugin actually helped make changes, does it?
No, it does not. I used the `Reformat Code` option of Intellij on the whole project. Then I ran `gradle check` to see if it all conforms to the checkstyle. But if changing so many files is the problem, what I can do is uncomment one subrule and do the changes for that in one PR. 

And I think we need to follow this rule only for the _indentation_ section. The other sections should not affect so many files.

lognaturel:
Awesome, thanks for that extra info. Let's close this one for now. I agree with @dcbriccetti that the hanging indents rule should go. Let me come up with some wording to clarify issue #94 and then we can figure out how to proceed on indentation specifically.

PR #212, opened 2017-10-15
Refactored java files to conform to import settings in checkstyle. Pa…

### LINKED ISSUE #94: Fix style throughout the codebase

rockydcoder:
…rtially fixes #94

Closes #

Used Intellij's CheckStyle plugin to refactor the java files

Rather than doing it manually, used Intellij's CheckStyle plugin

No, since this is just a refactor which includes changing imports layout and order.


lognaturel:
It looks like this also includes the changes from #211 which are controversial. Reordering the imports should be an easy thing to merge but the PR should only include that change!

rockydcoder:
Yes, @lognaturel . Saw that now. My bad. I am closing this PR. I will reopen a fresh one with only import reordering changes.

PR #217, opened 2017-10-18
bugfix #159: Added including date option to export form

### LINKED ISSUE #159: Export date range should allow choosing inclusive or exclusive from UI.
From @yanokwa in https://github.com/opendatakit/briefcase/pull/139:

We should make the briefcase export UI work similar to the build UI for selecting date ranges:

28392841-a630da24-6c98-11e7-9144-24357ae11632

This should be as simple as adding the checkboxes to the UI, and based on isSelected() or not, adding one day to the selected date (the time defaults to 00:00:00 for the chooser).

rockydcoder:
Closes #159 

Have not tested it explicitly but this seems pretty straightforward. 

This is in sync with build UI for selecting date ranges

No risks.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/ExportPanel.java
+++ b/src/org/opendatakit/briefcase/ui/ExportPanel.java
@@ -25,9 +25,20 @@ import java.util.Date;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
-import javax.swing.*;
+import javax.swing.BorderFactory;
+import javax.swing.DefaultComboBoxModel;
+import javax.swing.GroupLayout;
 import javax.swing.GroupLayout.Alignment;
+import javax.swing.JButton;
+import javax.swing.JComboBox;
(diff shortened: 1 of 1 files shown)


lognaturel:
Awesome, thanks!

It's not quite working as I expected. I pulled the All Widgets form from the default https://opendatakit.appspot.com server. I set the end date to October 14 and unchecked the inclusive box. I got records through Oct 10. Then I checked the inclusive box and exported again. I got records through Oct 13. If I set the end date to Oct 16 then I see that there is an Oct 14 record which I would have expected to see when the end date is Oct 14 and the inclusive box is checked. This could be a but with the date selection feature we hadn't previously noticed, I'm not sure.

UI-wise, I initially had a little trouble knowing which checkbox belongs to which date since the spacing is identical.

How about grouping them a bit more as in:
screen shot 2017-10-18 at 10 26 09 am
We should discuss this before changes are made -- maybe someone else has a better idea!

Issue #209, opened 2017-09-29
Provide command-line mode feedback to System.out


lognaturel:
As highlighted by @batkinson in https://github.com/opendatakit/briefcase/issues/197#issuecomment-332254472, messages in the command line interface are provided through logging. This is problematic because it means that including a default logging configuration that goes to a file would make the CLI more difficult to use.

#### Steps to reproduce the problem
Configure logging to go to a file and use Briefcase in command-line mode. For example:
`java -jar -Dorg.slf4j.simpleLogger.logFile=.briefcase.log ODK\ Briefcase\ v1.8.0.jar --form_idxxx --aggregate_url xxx --odk_username xxx --odk_password xxx --storage_directory xxx`

There will be no feedback at the command line because all interactions are done through logging calls.

#### Expected behavior
Feedback is always provided at the command line when using the command line interface.

PR #218, opened 2017-10-18
BRIEFCASE-210: Adding console and file appender for slf4j.

### LINKED ISSUE #210: Log to a briefcase.log file by default
**#209 needs to be done first** so command-line mode isn't broken.

Provide a default configuration to the SLF4J simple binding so errors are logged to a `briefcase.log` file. This will make it easier to troubleshoot user problems. I propose making the filename `briefcase.log` rather than `.briefcase.log` so it's easy for users to look at and share if needed.

In #197 I had proposed logging at level `INFO` but I think that will be too much. I think `WARN` should provide sufficient information.

This will involve placing a file named `simplelogger.properties` in the `res` directory.
```
# Log to a file at log level WARN
org.slf4j.simpleLogger.logFile=briefcase.log
org.slf4j.simpleLogger.defaultLogLevel=warn
```

nribeka:
Closes #210 

I run briefcase and look at the generated console log and file log.

Since the google analytics require slf4j to work, we need to re-route all of our logging through slf4j and remove the dependency to commons-logging.

This could create a pretty big log file.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,7 @@ dependencies {
     compile files('lib/LGoodDatePicker-10.3.1-backport.jar')
 
     runtime group: 'org.slf4j', name: 'jcl-over-slf4j', version: '1.7.5'
-    runtime group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.5'
+    runtime group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
 }
 
 // Required to use fileExtensions property in checkstyle file
--- a/res/logback.xml
+++ b/res/logback.xml
@@ -1,0 +1,29 @@
+<configuration>
+    <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <file>briefcase.log</file>
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+            <fileNamePattern>briefcase.%d{yyyy-MM-dd}.%i.log</fileNamePattern>
+            <!-- keep 10 log history capped at 100MB total size and 10 MB each -->
+            <maxHistory>10</maxHistory>
+            <maxFileSize>10MB</maxFileSize>
+            <totalSizeCap>100MB</totalSizeCap>
+        </rollingPolicy>
+        <encoder>
+            <pattern>%date %level [%thread] %logger{10} %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+            <level>INFO</level>
+        </filter>
+        <encoder>
+            <pattern>%date %level [%thread] %logger{10} %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="debug">
+        <appender-ref ref="FILE" />
+        <appender-ref ref="STDOUT" />
+    </root>
+</configuration>


res/logback.xml
```diff
@@ -0,0 +1,26 @@
+<configuration>
+    <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <file>briefcase.log</file>
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+            <fileNamePattern>briefcase.%d{yyyy-MM-dd}.%i.log</fileNamePattern>
+            <!-- keep 14 days' worth of history capped at 1GB total size -->
```
lognaturel: Looks like it's actually 10 days for a total of 100MB?

res/logback.xml
```diff
@@ -0,0 +1,26 @@
...
+        </encoder>
+    </appender>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>[%date] [%level] [%thread] [%logger{10}] %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="debug">
```
lognaturel: Are we sure we want the default to be debug? That's pretty chatty! Maybe warn for the console and debug for the log file?

dcbriccetti:
Yes, @batkinson should approve.
Not debug for the console.

nribeka:
For some reason, I can't add @batkinson in the reviewer list. Maybe setting in the repo?

res/logback.xml
nribeka: Ah, I put 14 days first and then I changed it to 10. Will update 👯‍♂️ .

res/logback.xml
nribeka: I changed it to info for the console since I think it's good to see some informational message in the console, no? :)

batkinson: This discussion highlights why I said that getting clever with the logging config was the wrong solution. The reality is that when using CLI-mode you would want different console output than when using it in GUI-mode. You can accomplish this in a few ways, like programmatically manipulating the logging config at runtime. However, IMO it is better to simplify the design so it actually makes sense; use the console output streams System.out and err to provide CLI-specific messages. Logging should be used for other diagnostics, and sparingly (at least in terms of I/O - logging frameworks like this exist to make it possible to add good logging code that can later be turned on without causing performance issues... until the deeper diagnostics are necessary).

By trying to reduce this to a single problem, you conflate the meaning of the configs. You are going to have to carefully manage them in order to prevent inadvertently squelching and/or polluting the CLI output by removing/adding logging code (or changing levels) in other classes. You're increasing the complexity and cognitive load for a low-quality utility. That is simply not good design.

Ideally, in all logging modes you limit logging output to only relevant information. In practice, determining relevance can be difficult due to the fact you don't know perfectly in advance what the problems will be. However, to actually output routine information just adds unnecessary I/O that will slow the application down (which can be severe). You can (and should) log that type of information at debug or trace levels and squelch them by default. Logging routine information like this also makes diagnosing issues more difficult because you have to sift through massive volumes of "normal" output to find interesting information. This is true even if you can squelch them, since when you need to enable debug (or other levels) logging, you're still going to want quality messages. The moral is to carefully design the logging mode as a user experience in itself. Here we're doing the opposite - merging multiple experiences creating a single (lousy) one. The Apache HTTP Components logging is a great example of logging used well. It was clear they were purposeful in their use of logging, briefcase's use is not (yet).

As much as possible, we should instead try to provide direct feedback to users (through console output on CLI and graphical controls in GUI) rather than forcing users to sift through log output to figure out what's going on (and devs to keep both modes operating well). You will have full control over the user experience by not merging CLI and GUI output. You also can take full advantage of the strengths of the respective application modes and intent is clear from the code.

build.gradle
```diff
@@ -51,6 +51,7 @@ dependencies {
     compile files('lib/LGoodDatePicker-10.3.1-backport.jar')
 
     compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
+    compile group: 'org.slf4j', name: 'jcl-over-slf4j', version: '1.7.5'
```
batkinson: As I said previously, this shouldn't be a necessary change. The original code was working (and was more accurate in its scope). It also pollutes the stream of changes with what amounts to a no-op and changes the owner of the change so blame doesn't work as intended. Instead, tidy up the commits to the minimal set of changes.

build.gradle
```diff
@@ -51,8 +50,7 @@ dependencies {
     compile files('lib/threetenbp-1.3.1.jar')
     compile files('lib/LGoodDatePicker-10.3.1-backport.jar')
 
-    runtime group: 'org.slf4j', name: 'jcl-over-slf4j', version: '1.7.5'
-    runtime group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.5'
```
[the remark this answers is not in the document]
nribeka: I don't get any error when I removed the jcl-over-slf4j or both of this dependencies. I added later because I talked in slack and was deemed needed to have explicit dependency to jcl-over-slf4j instead of relying on the one comes from google-analytics-java dependency.

build.gradle
```diff
@@ -51,8 +50,7 @@ dependencies {
...
+    compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
```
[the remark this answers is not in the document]
nribeka: I agree this should be runtime instead of compile. Will change it.

build.gradle
```diff
@@ -36,7 +36,6 @@ dependencies {
     compile group: 'org.bouncycastle', name: 'bcprov-jdk16', version: '1.46'
     compile group: 'commons-cli', name: 'commons-cli', version: '1.3.1'
     compile group: 'commons-io', name:'commons-io', version: '2.5'
-    compile group: 'commons-logging', name: 'commons-logging', version: '1.2'
```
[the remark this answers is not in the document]
nribeka: I agree on your points:
* This should've been removed when jcl-over-slf4j introduced / when common-logging exclusion was added.
* This shouldn't have been part of this PR on the first place.

build.gradle
nribeka: This was added because we want to say explicitly we're using this library instead of just transitively getting it from the google analytics library. Removing this line will do no harm.

nribeka:
@batkinson 

Based on your comments, should I just recreate the PR with the needed changes? That way the blame will reflect only my changes?

res/logback.xml
nribeka: I'm not sure I catch everything here.

For me personally, I like to see some informational messages getting printed in my console when I run Java application. To me seeing those messages imply progress, there's something that the system is doing currently.

For CLI, I agree that the reply the user is getting should be in the System.out or System.err instead of getting log messages printed in front of them. Isn't that what #209 should be addressing? And I still see the benefit of CLI application create log file with more information that what the user is getting through System.out and System.err.

build.gradle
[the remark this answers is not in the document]
batkinson: @nribeka your comment actually highlights my point. You just created changes and make it look like original work. This was already fixed in 9c8cbd0ddd43.

batkinson:
@nribeka, no need to resubmit the PR. You just need to recraft the minimal changes to address the issue and push --force to the branch you submitted as a PR (nribeka:briefcase-210).

build.gradle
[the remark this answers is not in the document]
batkinson: By the way, I don't disagree with the explicit dependency (as you can see, I actually added it). I'm not sure what was said on slack about it, but when your code actually depends on transitives it adds risk that changes made to downstream libraries breaks the build with no apparent change (the build file won't change, just the artifact metadata, which is only visible in the artifact repository). Also, by modeling the *actual* dependencies the build file tells you a story. Lots of deps, your project will be more susceptible to decay over time and may suffer from library mismatching or conflicts.

lognaturel:
@nribeka @batkinson I'm sorry I haven't had a chance to react to all the great conversation here. I have high hopes for tomorrow! 😊

res/logback.xml
batkinson: @nribeka My overall feedback is that we should consider the resource's purpose to the user, in this case the console, and make usage the most effective considering the purpose. I don't think this is considerate of purpose. The console's utility to CLI-mode is different than in GUI-mode when launched from the command line. It changes again when launched from click-mode, where the console is entirely hidden.

I understand that a moving console can be comforting for many devs while developing. I am arguing your feeling of comfort may be misplaced: it's simply not effective use of the medium. An info message or two at startup is fine, but ideally you want feedback to be presented where the user will expect it and can use it: in the primary user interface. The current interface reduces the value of the console by raising the noise threshold past the point where normal users will venture to comprehend it. Add to consideration the performance degradation (for quite possibly invisible, messages) logging at inappropriate log levels and it enters the realm of "this project doesn't know what it is doing". We also would double-log these, once in the console and again in a persistent log. That is uninspiring and honestly, silly. 

A better alternative is showing only a single message at ERROR with the form uuid and plain-language message (user-actionable ideally) rather than 1000s of status messages with one error message somewhere in the noise.

This shouldn't be contentious. The reason for the framework is to allow for power-users or devs to override the defaults, but the default should be intelligent for common users.

* Fix the log levels to be more appropriate. For example, INFO shouldn't be active unless it's actually informative. Logging the form fetch for 100s or 1000s of forms isn't informative.
* Fix the default logging config to use the console sparingly (perhaps at WARN, or INFO once the logging level usages are fixed)
* Decouple CLI-mode output from the logging. Whether or not you want to log at INFO or not should not remove the primary output from CLI-mode. It is a misuse of the tool.

lognaturel:
Thank you @batkinson for all the useful feedback! I apologize -- @nribeka, @dcbriccetti and I had a chance to discuss this in Slack last week and I forgot to summarize here. We are all in agreement that sysout/syserr should be the primary feedback mechanisms for the CLI. #234 goes in that direction and we wanted to get that completed before coming back here and seeing what makes sense after that change. 

After trying out what is in #234, we found it provided too much output and are thinking of not putting anything on sysout unless a `--verbose` flag is included. Alternately, we could show only a message on successful completion.

You're absolutely right that we'll need to clean up log levels. I think we're all in agreement with what the final state should be, it's just a matter of finding the most effective sequencing given that what we expect to be small fixes tend to uncover a "rich vein" (as @dcbriccetti says) of other improvements that could be made. For example, the CLI surprisingly relies on dialogs for some error states. Or someone did make an attempt to use the same output as is shown in the GUI in the CLI but didn't quite put all the pieces together right.

kkrawczyk123:
Tested with success:
Verified on Ubuntu 17 and Windows 7

Three new files are created:
briefcase.log in /briefcase/
briefcase.log in /briefcase/build/libs
logback.xml in /briefcase/res

Logs while using the app are collected in:
briefcase.log in /briefcase/build/libs

Is it result that you wanted to obtain in these pull request? I am not sure what is the function of briefcase.log file in /briefcase/ if it is not logging while briefcase works?

lognaturel:
Closing in favor of #356, #368

PR #219, opened 2017-10-18
Uncomment Checkstyle rules which produce no issues

zstojanovic:
Partially addresses #94 
- uncommented 25 Checkstyle rules to which the code already conforms
- ran Checkstyle to check that there are no issues


lognaturel:
Thanks!

PR #224, opened 2017-10-20
BRIEFCASE-223: Stripping invalid chars from title.

### LINKED ISSUE #223: Briefcase successfully pull forms but not creating the correct folder structure
#### Software versions 
Briefcase v1.6

#### Problem description
When form title have period at the end, pull will indicate success but form name will not go to the exportable form list.

* Create with period at the end.
* Pull from aggregate or odk folder
* Briefcase will pull it successfully but if you go to export, you won't see it in the list.

#### Expected behavior
After successful pull, the form will be in the export list.

#### Other information 
There is already sanitize function call, but it's not exhaustive enough.

nribeka:
Closes #223

Tested with form with invalid character in the title.

We need ensure that no invalid path will be created because of the title of the form contains invalid characters.

Folder name will be different with the title of the form. Before this patch, they will always be the same (for non Windows machine).


src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
```diff
@@ -672,7 +672,7 @@ protected BaseFormParserForJavaRosa(String existingXml, String existingTitle, bo
       }
     }
     // clean illegal characters from title
-    title = formTitle.replace(BasicConsts.FORWARDSLASH, BasicConsts.EMPTY_STRING);
+    title = formTitle.replaceAll(BasicConsts.INVALID_CHARS, BasicConsts.EMPTY_STRING);
```
nribeka: There's already effort to sanitize the title, so I'm just updating the implementation which will replace more chars from the title. The title is what being used to create the folder name and the file name for the form.

src/org/opendatakit/briefcase/model/RemoteFormDefinition.java
```diff
@@ -26,7 +28,7 @@
 
   public RemoteFormDefinition(String formName, String formId,
       String versionString, String downloadUrl, String manifestUrl) {
-    this.formName = formName;
+    this.formName = formName == null ? null : formName.replaceAll(BasicConsts.INVALID_CHARS, BasicConsts.EMPTY_STRING);
```
nribeka: I'm in between for this one. We could change the getFormName() implementation.

This one is used to display remote form in the pull transfer panel and won't affect the creation of the folder name. I think it would be nice to make sure it looks consistent to the user.

src/org/opendatakit/common/web/constants/BasicConsts.java
```diff
@@ -54,6 +54,7 @@
   public static final String RIGHT_BRACKET = "]";
   public static final String LEFT_PARENTHESIS = "(";
   public static final String RIGHT_PARENTHESIS = ")";
+  public static final String INVALID_CHARS = "[^a-zA-Z0-9\\s]";
```
nribeka: Replace all non alphanumeric character or non space character.

src/org/opendatakit/briefcase/model/RemoteFormDefinition.java
yanokwa: I'm a no on this one. I think we should only strip things when they are needed. The form title will have a period everywhere else (e.g., Collect, Aggregate) so you aren't really making things that consistent.

nribeka: Reverting.

src/org/opendatakit/common/web/constants/BasicConsts.java
yanokwa: We might want to make this more robust, because I think this will allow Unicode, which will break on some filesystems. I'm thinking we should only allow a-z, A-Z, 0-9, and hyphen, underscore, and space. And we should probably do the same thing when we write out the CSV. @dcbriccetti what do you think?

nribeka: This will only allow a-z, A-Z, 0-9 and space character in the title. You want to include hyphen and underscore in the list?

yanokwa: Regex fail by @yanokwa 😭

Yeah, let's add hyphen and underscore!

src/org/opendatakit/common/web/constants/BasicConsts.java
```diff
@@ -54,6 +54,7 @@
   public static final String RIGHT_BRACKET = "]";
   public static final String LEFT_PARENTHESIS = "(";
   public static final String RIGHT_PARENTHESIS = ")";
+  public static final String INVALID_CHARS = "[^a-zA-Z0-9\\ \\-\\_]";
```
nribeka: @yanokwa,

I added the _ and - and I'm using explicit space character instead of \s now.

dcbriccetti:
I’m looking now.

src/org/opendatakit/briefcase/model/RemoteFormDefinition.java
nribeka: Will revert once @dcbriccetti is cool with the changes.

dcbriccetti: How about reverting now so I can see exactly what you are proposing to do?

dcbriccetti: I definitely agree that the name should be modified only when needed, i.e., when mapping it to a file path.

dcbriccetti:
Is there a need to implement this before solving https://github.com/opendatakit/briefcase/issues/230? Are they both addressing the same problem?

yanokwa:
@dcbriccetti Yes, it's basically the same problem.

kkrawczyk123:
I wasn’t able to reproduce issue [#223](https://github.com/opendatakit/briefcase/issues/223) I tried on both Briefcase 1.6 and 1.8 versions and on both systems Windows 7 and Ubuntu 17. I created two forms, one with multiple space after name and second with multiple _ in the end of name. I tried to pull them from aggregate and odk collect. Pulling finished successfully. Both forms were present in pull table and then in push table and in export field. Both forms exported into .csv file with success. Additionally aggregate and collect removed multiple spaces from the name automatically during pull process. Did I do something wrong when reproduce the problem?

lognaturel:
@yanokwa @nribeka you were able to reproduce the issue, right? Any chance you can help @kkrawczyk123 reproduce and verify the fix?

This doesn't interfere with larger refactors ongoing so if it does fix a real problem it would be good to finish it up!

yanokwa:
@kkrawczyk123 You said, `I created two forms, one with multiple space after name and second with multiple _ in the end of name.' As @nribeka said at https://github.com/opendatakit/briefcase/issues/223, the form title has to have a period at the end to reproduce. Can you try that?

kkrawczyk123:
Thanks @yanokwa, I checked it again, this time with form with period in the end of name means multiple dot which I guess I didn’t understand before.

Ubuntu:
On the master branch, I was able to pull from aggregate and collect that form without any changes in the name. I was also able to export it but period was replaced by empty space which looked like ______.csv. Then I tried the same activities on the branch with the same form. When I pulled it from aggregate and collect it seems to don’t involve any period of dots in name any more. When I try to export result was the same. So automatically removing period from form’s name by briefcase did succeed.  

Windows:
I tried the same as on Ubuntu and I have problem with pulling from aggregate form with period on master, “parsing error” did show up to me. On the branch I had the same results as on linux, that dot period was removed  automatically from form’s name. I had also problem when I tried to copy “odk” folder directly from collect, some unexpected error occurred everytime I tried. So I couldn’t check what happen when pull from collect.

Mac:
Results was the same for master and for the branch with pull from aggregate process and export as on ubuntu. On Mac also occured the same problem with copy “odk” folder with that invalid form directly from collect.

lognaturel:
Ok, this is turning into the most complicated little issue ever! Would like to try to get this finalized soon.

To recap, the new regex ensures that filenames will only contain alphanumeric chars, -, _ and space. This will be accomplished by removing any other characters.

For risks, @nribeka wrote:
> Folder name will be different with the title of the form

But I believe there are no real implications to that -- the filename and title aren't compared programmatically at any point, for example.

What is problematic is that a form name such as "シャンプー" which used to work just fine now can't be pulled through Briefcase. Can we just address the period case outlined in #223? Or can we do this more aggressive change only on Windows? @yanokwa why did you want to be so restrictive?

CSV export does `name.replaceAll("\\p{Punct}", "_").replace("\\p{Space}", " ")`. Could that be sufficient?

yanokwa:
@lognaturel I wanted to reduce the characters to a fully portable filename. I don't feel strongly about this, and upon reflection, a narrower fix and supporting Unicode seems like a good idea.

lognaturel:
I think Windows now handles unicode filenames. My preference would be to match what's done on the CSV export and preferably to reuse the same code. @nribeka what do you think? Do you see any problems with that approach?

nribeka:
Looking at the definition from Pattern.java:

```java
\p{Punct} | Punctuation: One of !"#$%&'()*+,-./:;<=>?@[\]^_`{\|}~
```

I think the one from the csv should work.

lognaturel:
@nribeka ✨ If you change it real quick we can get this in. 😊

PR #226, opened 2017-10-25
Continue (complete?) the task of removing Aggregate 0.9.x support. 

### LINKED ISSUE #225: Remove ServerConnectionInfo.isOpenRosaServer
> If you eliminate this field, isOpenRosaServer, which is now always true, you will discover another rich vein of code to delete. Keep going! :-)  You’ll also discover an error in some related error-detecting logic (that was there before your change).
-- https://github.com/opendatakit/briefcase/pull/208

dcbriccetti:
ServerConnectionInfo always sets the final boolean isOpenRosaServer to true. This change consists of the following:

Change isOpenRosaServer to always return true
Inline that method, creating always-true or always-false code.
Use automatic refactorings to remove code and if statements.
In the case of AggregateUtils, there was an incorrect error message produced when there is no WebUtils.OPEN_ROSA_VERSION_HEADER. Correct this.

Closes #225 

All changes, except the error message correction were made with IDEA automatic corrections and refactorings. I ran Briefcase and did a push and a pull. Perhaps @batkinson can review the error message change in AggregateUtils.

There’s only one way to do this.

No

### DIFF
--- a/src/org/opendatakit/briefcase/model/ServerConnectionInfo.java
+++ b/src/org/opendatakit/briefcase/model/ServerConnectionInfo.java
@@ -21,13 +21,11 @@ public class ServerConnectionInfo {
   private String url;
   private final String username;
   private final char[] password;
-  private final boolean isOpenRosaServer;
 
   public ServerConnectionInfo(String url, String username, char[] cs) {
     this.url = url;
     this.username = username;
     this.password = cs;
-    this.isOpenRosaServer = true;
   }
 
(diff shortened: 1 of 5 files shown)


dcbriccetti:
Hello @opendatakit-bot, please bark up @batkinson’s or @lognaturel’s trees. My tree is the wrong tree up which to bark.

dcbriccetti:
HAHAHAHAHAHAHA! Funny bot.

PR #231, opened 2017-10-26
Added single line declaration check

### LINKED ISSUE #94: Fix style throughout the codebase

icemc:
Closes #94 (partially)

It has been verified using the checkstyle plugin.No more multiple variable declaration on the same line

No risks


icemc:
@yanokwa @lognaturel this is up for review

PR #234, opened 2017-11-03
Improve command line user interface

### LINKED ISSUE #209: Provide command-line mode feedback to System.out

lognaturel:
Closes #209 

Tried the following CLI actions:
- `java -jar build/libs/ODK\ Briefcase\ v1.8.0-50-gd7a45a42.jar --form_id all-widgets --storage_directory ../ODK\ Briefcase\ Storage/ --aggregate_url https://nafundi-test.appspot.com -u x -p x`
- `java -jar build/libs/ODK\ Briefcase\ v1.8.0-50-gd7a45a42.jar --form_id all-widgets --storage_directory ./ --export_directory ~/Desktop --export_filename foo.csv`

This is a step towards a better interface. We know we shouldn't be relying on logging and should instead use System.out/System.err. We know events that go to the GUI should also go to this UI. There will still be things to clean up like not using dialogs.

This changes the CLI output because no events were previously logged. There's a chance of harming someone's workflow. Also, since logging still goes to the console (see #210), there are some log messages interspersed with actual user interaction messages.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/BriefcaseCLI.java
+++ b/src/org/opendatakit/briefcase/ui/BriefcaseCLI.java
@@ -36,6 +36,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.bouncycastle.openssl.PEMReader;
 import org.bushe.swing.event.EventBus;
+import org.bushe.swing.event.annotation.AnnotationProcessor;
 import org.bushe.swing.event.annotation.EventSubscriber;
 import org.opendatakit.aggregate.parser.BaseFormParserForJavaRosa;
 import org.opendatakit.briefcase.model.BriefcaseFormDefinition;
@@ -74,6 +75,7 @@ public class BriefcaseCLI {
     private static final Log log = LogFactory.getLog(BaseFormParserForJavaRosa.class);
 
     public BriefcaseCLI(CommandLine cl) {
+        AnnotationProcessor.process(this);
         mCommandline = cl;
     }
 
--- a/src/org/opendatakit/briefcase/ui/BriefcaseCLI.java
+++ b/src/org/opendatakit/briefcase/ui/BriefcaseCLI.java
@@ -102,16 +102,16 @@ public class BriefcaseCLI {
         if (!f.exists()) {
             boolean success = f.mkdirs();
             if (success) {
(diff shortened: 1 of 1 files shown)


lognaturel:
@dcbriccetti at the bottom of the class there are several methods annotated with `@EventSubscriber`. Those define actions upon reception of `EventBus` events. That's the mechanism used to update the GUI, usually in "details..." windows. The methods were never called because the annotations were never being ignored. Now they're actually called so the same info that goes to the GUI goes to the CLI.

We had some good discussion in Slack about how much output should be given. I'm convinced by the arguments that it should be silent unless things go wrong or a `--verbose` flag is included. Will update to reflect that soon.

src/org/opendatakit/briefcase/ui/BriefcaseCLI.java
```diff
@@ -248,7 +251,8 @@ public void run() {
                         break;
                     } catch (IOException e) {
                         String msg = "The supplied PEM file could not be parsed.";
-                        log.error(msg, e);
+                        System.err.println("msg");
```
rockydcoder: This will print _msg_ literally instead of the `msg` variable

rockydcoder: And I think `msg` can be inlined.

lognaturel: Thank you! This is why code review is so important.

lognaturel:
Closing in favor of deeper cleanup work started by @ggalmazor to establish a dedicated interface for the command line version and improve logging.

PR #238, opened 2017-11-06
Downgraded hsqldb to 2.3.5 to work with Java 7. Fixes #203

### LINKED ISSUE #203: No Suitable driver found for jdbc:hsqldb in (push and pull) using java 1.7.79
### Notes :                        in java jre version 1.8 (no issue).

rockydcoder:
Closes #203 

Application was running with both Java 7 and Java 8.

The latest version of hsqldb - 2.4.0 is not compatible with Java 7. Since we are currently supporting Java 7, we can use hsqldb - 2.3.5. This can be upgraded once we move to Java 8.

No.


lognaturel:
Thank you, @rockydcoder!

PR #243, opened 2017-11-17
Fixes #237. Replaces dialogs with command line messages in CLI

### LINKED ISSUE #237: Replace dialogs from CLI with command line messages
It's pretty unusual for a command line interface to have dialogs...

rockydcoder:
Closes #237 

Ran this command `java -jar ODK\ Briefcase\ v1.8.0-50-gb2f3f12-dirty.jar -id encrypted-form -sd ~/Desktop -ed ~/Desktop -f test1.csv -pf ~/Downloads/test.pem` in terminal with a valid and invalid pem file and the response was as expected.

CLI should not have dialogs as a way to throw exception. Hence, getting rid of dialogs and introducing CL messages is the logical solution.

No, since this is related to error logging only.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/BriefcaseCLI.java
+++ b/src/org/opendatakit/briefcase/ui/BriefcaseCLI.java
@@ -223,9 +223,8 @@ public class BriefcaseCLI {
                           // ignore.
                         }
                         if (o == null) {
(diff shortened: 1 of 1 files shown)


lognaturel:
Thank you, @rockydcoder! That makes a lot more sense! 😊

PR #252, opened 2017-11-24
Add a cli adapter

ggalmazor:
This is the first of a series of PRs that will try to solve current coupling problems in Briefcase between logging, command line args parsing, console command execution and console UI concerns.

Currently, there is no clear separation between what can be executed from the command line, required and optional args parsing and the rest of Briefcase's logic. The current code relies heavily on duplication, conditional blocks and comments to give readers hints of what's going on and it's intertwined in code that prepares and launches the GUI.

Although the command line args parsing is not the most pressing issue to be solved, it is the easiest to start with and it will make easier the task of decoupling the rest.

**Warning: this PR bumps the project to Java8** 

#### What has been done to verify that this works as intended?
Run the "pull from Aggregate" and "export form" operations from the new command line interface and from the GUI.

I haven't been able yet to test the "import from sd card" and the "export encrypted form" operations. Will work on this but I wanted to create the PR to start discussion.

#### Why is this the best possible solution? Were any other approaches considered?
By creating a command line adapter in `Cli`, we can define a clear constraint about where to do command line stuff and where not to. 

This change lets us:
- be explicit about what operations can be executed from the command line
- clearly define their required and optional params
- a way to obtain argument values passed from the command line (and even map them from `String` to other types)

Also:
- it doesn't introduce any new library to the project.
- it could be extracted to a lib and be reused in other parts of ODK.

#### Are there any risks to merging this code? If so, what are they?
Any risk related to Java8 migration. I think that, as discussed in #158 and #203, this should be a safe move

PR #253, opened 2017-11-27
Decouple UI from logic

ggalmazor:
This PR is a continuation of #252 and **it's not ready to be merged**.

I'm creating this PR just to show and discuss what I'm thinking about separation of concerns in Briefcase.

TL;DR: I'd like to propose to coordinate through an event bus all UI outcomes to whatever Briefcase operation has been executed. There is an example of this in the Export operation, which is reused by both the console and the GUI.

Currently, Briefcase has some pure business logic code that gets invoked by the UI and the console. `ExportAction` and `TransferAction` are two examples of this. These classes are coupled to the UI in two ways, mainly:
 - By using the static `EventBus` Swing API
 - By being aware of the concurrency details of whatever is going to invoke them: the `TerminationFuture` instance that's passed over from one place to the next and the fact that there are static instances of `Executor` implementation.

These coupling points make hard to reuse these pure logic classes, and the proof of that is that this causes lots of code duplication wherever the console needs to invoke them.

I have removed this coupling points by separating the context of execution of the `ExportAction` class from the actual arguments for the invocation of its `.execute(...)` method. Now, in order to obtain an instance, you must provide a `PubSub` (the event bus) instance and an `Executor` instance.

In the Swing GUI concurrent context, you use a thread-safe `PubSub` implementation and the default Java8 thread pool as the `Executor`. In the simpler console context everything can be executed in the main thread, therefore, you can use a simple (single-threaded) `PubSub` and `Runnable::run` as your `Executor` instance (it will run everything in the main thread).

This opens the door to test these action classes as well.

In this rationale, pure logic classes must not interact with the user directly and any communication must be done publishing events. They can, though, write logs, but only for tracing the program's execution.

The next steps would be to continue replacing EventBus and ad-hoc `Executor` instances with the same `PubSub` and `Executor` instance created at bootstrap time to every class that needs it.

Links to highlighted parts of the PR:
- [pubsub package](https://github.com/ggalmazor/briefcase/tree/ui_separation/src/org/opendatakit/common/pubsub)
- [Instantiation of the Swing GUI](https://github.com/ggalmazor/briefcase/blob/ui_separation/src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java#L70)


ggalmazor:
I'm closing this PR to create it again with better git history

PR #262, opened 2018-01-06
Issue 261 file folder dialog

### LINKED ISSUE #261: Folder selection behavior is odd
In the master build of Briefcase, here's the behavior I'm seeing in Export Directory on macOS. 

If I'm "in" the Desktop folder, Open should select that folder. Instead, I have to be in my Home folder and before I can select the Desktop folder. This only happens when Export Directory is blank. When it's not, I can be in a folder and select it.

select-folder

@ggalmazor reports that this doesn't happen on Linux, which suggests it might be a platform-specific bug.

@kkrawczyk123 everywhere in Briefcase where is a selection for a folder or file (Export DIrectory, PEM Private Key File, Storage Location), can you try the master build on Ubuntu, macOS, and Windows and report back what you find? Both when the selection is a folder and when it is a file.

ggalmazor:
Closes #261 

I've been able to reproduce the error on a virtualized Mac.

I've been able to fix the problem, just by forcing Briefcase to use Swing's file dialog.

After confirming that this is, indeed, a fix, we should migrate other file dialogs to use the new FileChooser.

Manually tested the changes on Linux and a virtualized Mac (High Sierra)

First I thought of making a workaround by debugging what the previous file dialog returned when clicked on "Open" without selecting any item but then I tried to just force Swing's dialog and it worked. ¯\\\_(ツ)\_/¯ 

I don't think so.

### DIFF
--- a/src/org/opendatakit/briefcase/util/FindDirectoryStructure.java
+++ b/src/org/opendatakit/briefcase/util/FindDirectoryStructure.java
@@ -1,12 +1,12 @@
 /*
  * Copyright (C) 2012 University of Washington.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
  * the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
  * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
@@ -26,7 +26,7 @@ import org.apache.commons.logging.LogFactory;
 
 /**
  * Originally written by Dylan.  Determines the mounts that have SD Cards attached.
- * 
+ *
  * @author the.dylan.price@gmail.com
  * @author mitchellsundt@gmail.com
  *
@@ -44,12 +44,17 @@ public class FindDirectoryStructure {
     String os = System.getProperty(PROPERTY_OS);
     return os.contains(OS_MAC);
   }
-  
+
+  public static boolean isUnix() {
+    String os = System.getProperty(PROPERTY_OS).toLowerCase();
+    return (os.contains("nix") || os.contains("nux") || os.contains("aix"));
+  }
+
   /**
    * Searches mounted drives for /odk/instances and returns a list of
    * positive results. Works for Windows, Mac, and Linux operating
    * systems.
-   * 
+   *
    * @return a {@link List} of {@link File}  containing matches of currently mounted file systems
    *         which contain the directoryStructureToSearchFor
    */
@@ -72,7 +77,7 @@ public class FindDirectoryStructure {
       File f = new File("/media", username);
       if (f.exists() && f.isDirectory()) {
         mountslist.add(f);
-      } 
+      }
 
       f = new File("/run/media", username);
       if (f.exists() && f.isDirectory()) {
@@ -92,11 +97,11 @@ public class FindDirectoryStructure {
     }
     return false;
   }
-  
+
   /**
    * Checks each given potential directory for existence of odk/instances
    * under it and returns a list of positive matches.
-   * 
+   *
    * @param mounts
    *          the potential mount points to check.
    * @param isDirectChild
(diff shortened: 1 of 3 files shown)


yanokwa:
This PR works, but I think we should use native widgets whenever possible. Is there a reason why a native file chooser wouldn't be preferred?

ggalmazor:
I don't really know... The original code had 2 implementations: 

1. One based on Swing and used by default
2. One based on AWT and used only when Briefcase runs on a Mac with Java7 because there was some kind of incompatibility issues

Since we're not using Java7 anymore, I've just removed implementation n. 2.

When you say "use native widgets" are you referring to the AWT implementation that's been removed? Take into account that on Windows and Linux hosts, Swing is what Briefcase has always used and only Mac users would see that something has changed.

lognaturel:
From master, when I open a file chooser from export or settings using Java 8 on a Mac, I get the native file choosers. With this change, from the export tab I get this:
screen shot 2018-01-08 at 12 01 32 pm

And from the settings tab I get the native chooser:
screen shot 2018-01-08 at 12 02 03 pm

I agree the native chooser feels better. I don't understand why the behavior has changed, though. I would have expected the AWT code path not to run previously since I'm on java 8.

ggalmazor:
Damn booleans! I think I got it wrong from the beginning:

Please help me extract some names from this:

```
boolean useAwt = (!((javaVersion.compareTo("1.7") >= 0) && javaVersion.compareTo("1.8") < 0)) && FindDirectoryStructure.isMac();
```

```java
boolean greterOrEqualThan7 = javaVersion.compareTo("1.7") >= 0;
boolean lessThan8 = javaVersion.compareTo("1.8") < 0;
boolean useAwt = (!(greterOrEqualThan7 && lessThan8)) && FindDirectoryStructure.isMac();
```

```java
boolean greterOrEqualThan7 = javaVersion.compareTo("1.7") >= 0;
boolean lessThan8 = javaVersion.compareTo("1.8") < 0;
boolean isJava7 = greterOrEqualThan7 && lessThan8;
boolean useAwt = (!isJava7) && FindDirectoryStructure.isMac();
```

```java
boolean greterOrEqualThan7 = javaVersion.compareTo("1.7") >= 0;
boolean lessThan8 = javaVersion.compareTo("1.8") < 0;
boolean isJava7 = greterOrEqualThan7 && lessThan8;
boolean notJava7 = !isJava7;
useAwt = notJava7 && FindDirectoryStructure.isMac();
```
So use Swing only if it's Java7 or if it's not a Mac. I should have removed option n. 1 and left AWT as the default file chooser :(

lognaturel:
Arg, I got tricked too. 😩 Goes to show the value of useful abstractions and naming.

ggalmazor:
I'm dropping these commits and doing the change again... Give me some minutes :P

ggalmazor:
Here is the thing: AWT's FileDialog on Linux doesn't let you select a directory we're forced to use another file chooser if we want this to work.

If we want to support just one implementation of a file chooser, the most reasonable solution IMO would be to use Swing's non-native file dialog for all platforms which lets use customize its behavior just the way we need.

Other solution would be to maintain the current two implementations. I'd fix the criterion to decide between Swing and AWT to something like "use always AWT unless you're in a Linux machine".

Some references:

lognaturel:
I don't mind keeping the two implementations. I only suggested removing it because I misunderstood why they were needed! "Always use AWT/native unless on Linux" seems great to me as long as comments/variable names are meaningful.

ggalmazor:
It seems that CircleCI has failed on the cloning phase... :/ 

Is there something I can do to help?

yanokwa:
This seems to be some subtle git/CircleCI bug because it works great when I push to my CircleCI account. Let's not worry about it after we merge.

ggalmazor:
Should I push a small change just to trigger a build?

src/org/opendatakit/briefcase/util/FindDirectoryStructure.java
```diff
@@ -44,12 +44,18 @@ public static boolean isMac() {
     String os = System.getProperty(PROPERTY_OS);
     return os.contains(OS_MAC);
   }
-  
+
+  public static boolean isUnix() {
+    // Copied from https://www.mkyong.com/java/how-to-detect-os-in-java-systemgetpropertyosname/
```
yanokwa: Maybe it was inspired by the mkyong site, but I would remove this comment because it's not a copy. 

As an aside, unless the code on the site you look at is explicitly licensed, we can't use it. That is public code != open source code. We have this policy detailed in https://github.com/opendatakit/briefcase/blob/master/CONTRIBUTING.md#code-from-external-sources.

PR #263, opened 2018-01-08
Add a simple Swing test

ggalmazor:
This PR is just to explore solutions to the CircleCI build that let us run Swing UI tests.


yanokwa:
You'll need to set up a windowing environment on CircleCI in `.circleci/config.yml` after the checkout command.

```yaml
      - run:
          name: Set up windowing environment
          command: |
            export DISPLAY=:99.0
            Xvfb :99 -screen 0 1024x768x24 >/dev/null 2>&1 &
```

yanokwa:
Ugh. I should have tested the PR. This full thing below should work. The key is to put `DISPLAY` under environment and then to add `background:true`

```yaml
version: 2
jobs:
  build:
    working_directory: ~/src
    docker:
      - image: circleci/openjdk:latest
    environment:
      - DISPLAY: :99.0
    steps:
      - checkout
      - run:
          name: Set up windowing environment
          command: Xvfb :99 -screen 0 1024x768x24 >/dev/null 2>&1
          background: true
      - restore_cache:
          key: jars-{{ checksum "build.gradle" }}
      - run: ./gradlew check test jar
      - save_cache:
          paths:
            - ~/.gradle
            - ~/src/.gradle
          key: jars-{{ checksum "build.gradle" }}
      - store_artifacts:
          path: build/reports
          destination: reports
      - store_artifacts:
          path: build/libs
          destination: libs
      - store_test_results:
          path: build/test-results
```

ggalmazor:
When a test fails, CircleCI gives enough feedback but could we have more information about the successful or ignored tests as well? Now it just says:

    Your build ran 7 tests in test with 0 failures
    Slowest test: org.opendatakit.briefcase.ui.export.ExportPanelTest export_button_should_be_enabled_when_there_are_no_configuration_errors (took 1.51 seconds).

yanokwa:
Go to Artifacts, then drill down: Container 0, reports, tests, test, index.html. That should give you all the test results.  It'll look like this: https://291-40322074-gh.circle-artifacts.com/0/reports/tests/test/index.html

ggalmazor:
Looking good!

dcbriccetti:
@yanokwa I assume you meant I should review this, but leave it assigned to @ggalmazor. He and I are looking at it together right now.

test/java/org/opendatakit/briefcase/ui/SwingTestRig.java
```diff
@@ -0,0 +1,74 @@
...
+      else
+        throw new RuntimeException("Can't create test Briefcase storage folder");
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  public static void installFormsFrom(Path path) {
+    Path briefcaseStorageFolder = Paths.get(new StorageLocation().getBriefcaseFolder().toURI());
+    uncheckedWalk(path).forEach(sourcePath -> uncheckedCopy(
```
dcbriccetti: I can see why you need uncheckedCopy (because of the stream), but not the other two unchecked..., given that they are each called only once.

screen shot 2018-01-11 at 1 28 02 pm

test/java/org/opendatakit/briefcase/ui/export/ExportPanelPageObject.java
```diff
@@ -0,0 +1,51 @@
...
+class ExportPanelPageObject {
+  private final ExportPanel exportPanel;
+  private FrameFixture window;
+
+  private ExportPanelPageObject(ExportPanel exportPanel, FrameFixture window) {
+    this.exportPanel = exportPanel;
+    this.window = window;
+  }
+
+  static ExportPanelPageObject setUp(Robot robot) {
```
dcbriccetti: Nice correct capitalization of the phrasal verb `set up` (as opposed to the noun or adjective `setup`). We often don’t see that.

test/java/org/opendatakit/briefcase/ui/SwingTestRig.java
```diff
@@ -0,0 +1,74 @@
...
+import org.opendatakit.briefcase.util.FileSystemUtils;
+
+public class SwingTestRig {
+  private static final Log log = LogFactory.getLog(SwingTestRig.class);
+
+  public static void prepareBriefcaseStorageFolder() {
+    try {
+      BriefcasePreferences.setBriefcaseDirectoryProperty(Files.createTempDirectory("briefcase_test").toString());
+      File briefcaseFolder = new StorageLocation().getBriefcaseFolder();
+      boolean mkdirs = briefcaseFolder.mkdirs();
```
dcbriccetti: Consider a more meaningful name, like `dirsMade` or `made` or `created` or something?
(15 seconds later)  Or inline the variable.

test/java/org/opendatakit/briefcase/ui/export/ExportPanelTest.java
```diff
@@ -0,0 +1,44 @@
...
+  @Test
+  public void export_button_should_be_disabled_by_default() {
+    page.exportButton().requireDisabled();
+  }
+
+  @Test
+  public void export_button_should_be_enabled_when_there_are_no_configuration_errors() throws IOException {
+    // Export dir must exist to be valid
+    page.setExportDirectory(Files.createTempDirectory("test_briefcase_export").toString());
+    page.selectFormATRow(0);
```
dcbriccetti: This is cool. I can run this under the debugger, and watch the row getting selected.

test/java/org/opendatakit/briefcase/ui/SwingTestRig.java
ggalmazor: it's just to avoid bloating the code with useless try/catch blocks. An alternative would be to declare that this method throws this and that exceptions but my **secret plan** (don't tell anyone) is to use the `Try` monad from Vavr.io to "uncheck" those calls when we add it to the project. Can we agree on leaving it like that in the meantime?

ggalmazor:
Hi @yanokwa! This PR is flagged with requested changes but I think we covered all the feedback already... didn't we? Did I miss anything?

Issue #258, opened 2018-01-04
Support for bulk form exports


ggalmazor:
This is a new feature proposal originally posted by @yanokwa and discussed in [this post on the forum](https://forum.opendatakit.org/t/adding-bulk-export-of-data-to-briefcase-gui/10868):

> I’d like to update the Export user interface to allow the selection of multiple forms, each one with a separate export directory, export type, private key, etc.
> 
> This is an important feature because some implementers (hi noel) have campaigns that need the flexibility to put different exports in different folders (just like the CLI supports) with different permissions, but they need to do it without requiring the complexities of writing a batch/shell script.
> 
> The changes I’m proposing are:
> 
> - Update Export interface to mirror the Pull/Push interface.
>   - Allow users to select some or all forms and then export the selection with a single button press (similar to Pull/Push)
>   - Interface will have a list of Forms to Export with column headers for Selected, Form Name, and Configure
>   - Interface will have other useful column headers (e.g., Last Exported Date, Details) from community feedback
>   - Configure will allow user to select export directory, private key, and start/end date (similar to the current Export tab) for the specific form
>   - Configure will allow user to enable pull before exporting and tries to use persisted pull credentials
> - Update application so all parameters are persisted between applications launches
>   - Sensitive parameters (e.g., passwords) will have a checkbox to save
> I’ll do all the things to get it to ship in Briefcase, but I’d love feedback from Briefcase users if what I’m proposing sounds reasonable.

I've extracted the following list of new small features (in no particular order) to track work:

- [x] New UI mirroring the Pull/Push interface. (#259)
- [x] New form table column: Configure (to set the export dir, pem file and date ranges for a specific form)
- [x] New form table column: Last export date
- [x] Settings (excluding passwords and any other sensitive parameters) must be persisted between app launches
- [ ] New setting: pull before export
- [x] Setting that the user can check to also save passwords and any other sensitive parameters and acknowledge that this is not safe.

Other related stuff:
- [x] #265 
- [x] #271

ggalmazor:
@opendatakit-bot label enhancement

PR #264, opened 2018-01-08
Issue 258 save settings in preferences

ggalmazor:
Partially addresses issue #258

In this PR we set a very basic use of persisted preferences to remember the export configuration between Briefcase launches.

The affected fields are the export dir and the pem file location.

Manually tested on my machine.

I've just mirrored what's done in other tabs

None regarding the app's stability but the lateral effect of this change is that, after setting a pem file location, you can never clear that field. This is due to currently not having a mechanism for clearing these kind of filesystem location fields in Briefcase. Other fields like the date ranges have a button to clear them.

### DIFF
--- a/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
+++ b/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
@@ -81,13 +81,18 @@ public class BriefcasePreferences {
      * Associates the specified value with the specified key in this preference
      * node.
      *
+     * If the value is null, then the key is removed
+     *
      * @param key
      *          key with which the specified value is to be associated.
      * @param value
-     *          value to be associated with the specified key.
+     *          value to be associated with the specified key or null.
      */
     public void put(String key, String value) {
-        preferences.put(key, value);
+        if (value != null)
+            preferences.put(key, value);
+        else
+            remove(key);
     }
 
     /**
--- a/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
+++ b/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
@@ -16,7 +16,9 @@
 
 package org.opendatakit.briefcase.model;
 import java.security.Security;
+import java.util.Optional;
 import java.util.UUID;
+import java.util.function.Function;
 import java.util.prefs.Preferences;
 
 import org.apache.http.HttpHost;
@@ -75,6 +77,28 @@ public class BriefcasePreferences {
      */
     public String get(String key, String defaultValue) {
         return preferences.get(key, defaultValue);
+    }
+
+    /**
+     * Returns the value associated with the specified key in this preference
+     * node mapped by a function that takes a String and returns a value of type T.
+     * Returns the specified default if there is no value associated with
+     * the key, or the backing store is inaccessible.
+     *
+     * @param key
+     *          key whose associated value is to be returned.
+     * @param mapper
+     *          function that takes a String and returns a value of type T
+     * @param defaultValue
+     *          the value to be returned in the event that this preference node
+     *          has no value associated with key.
+     * @param <T>
+     *          type of the output of this method
+     * @return the value associated with key, or defaultValue if no value is associated
+     *         with key, mapped with the mapper function, or the backing store is inaccessible.
+     */
+    public <T> T get(String key, Function<String,T> mapper, T defaultValue) {
+      return Optional.ofNullable(preferences.get(key, null)).map(mapper).orElse(defaultValue);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ggalmazor:
> In previous preferences work, we made the keys private and camel case ...

You're right. I based mines in `BriefcasePreferences.AGGREGATE_1_0_URL` which holds `url_1_0` but I didn't check the others... :P I'll change that... Also, I forgot to extract to constants those magic strings...

> ... and wrapped them with private methods. Is there a good reason not to do the same here?

These preferences will only be used by the `ExportPanel`, therefore, I don't see any benefit on extracting the access to them outside of it. Even more, I'd see that as a flaw in the design because it breaks encapsulation of the domain of ExportPanel.

One design improvement would be, in fact, to inline everything on Briefcase into their respective verticals unless it's reused from two or more verticals. Those reused preferences would be the only ones to stay there.

These little details become more and more important as we go further separating each functional part of Briefcase in their respective vertical package.

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
```diff
@@ -77,17 +79,44 @@ public String get(String key, String defaultValue) {
...
      * @param value
-     *          value to be associated with the specified key.
+     *          value to be associated with the specified key or null.
      */
     public void put(String key, String value) {
-        preferences.put(key, value);
+        if (value != null)
+            preferences.put(key, value);
+        else
+            remove(key);
```
yanokwa: I haven't seen this type of usage before where `put(key, null)` removes a key. Is this typical?

yanokwa:
This looks pretty good to me, @lognaturel. Can you please do a quick sanity check?

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
ggalmazor: It's just a convenient API for users of `BriefcasePreferences`. Take a look at how this is used in [line 148 of ExportPanel](https://github.com/opendatakit/briefcase/pull/264/commits/108397f99047f15fea7577f224b464bdafa9b75f#diff-53cd7276f9b5fe0579443fdf5f255e79R148).

The fact that the key gets removed from the Preferences is just an implementation detail that users of this class don't have to know. Also, removing the key is just what I'd have to do in [the same line](https://github.com/opendatakit/briefcase/pull/264/commits/108397f99047f15fea7577f224b464bdafa9b75f#diff-53cd7276f9b5fe0579443fdf5f255e79R148) inside an ugly if block inside the listener function.

The next step would be to accept `Optional<String>` values in `BriefcasePreferences.put(...)` to avoid handling nulls and simplifying even more the code at use site.

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -106,37 +114,43 @@ public ExportPanel(TerminationFuture terminationFuture) {
...
+    }));
 
     JLabel lblDateFrom = new JLabel("Start Date (inclusive):");
     JLabel lblDateTo = new JLabel("End Date (exclusive):");
 
     pickStartDate = createDatePicker();
+    pickStartDate.setDate(PREFERENCES.get(EXPORT_START_DATE_PREF, LocalDate::parse, null));
     pickStartDate.addDateChangeListener(__ -> updateExportButton());
     pickStartDate.addDateChangeListener(__ -> validateDate(pickStartDate));
+    pickStartDate.addDateChangeListener(event -> PREFERENCES.put(EXPORT_START_DATE_PREF, Optional.ofNullable(event.getNewDate()).map(ld -> ld.format(ISO_DATE)).orElse(null)));
```
lognaturel: How about some wrapping here? These really wide lines are hard to read on smaller screens (e.g. my laptop) or when arranging multiple windows horizontally. Even with a wide screen I think it's difficult to have to scan horizontally for a long time.

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -305,10 +335,11 @@ public void updateForms() {
         .collect(toList()));
   }
 
-  private void chooseLocation(FileChooser fileChooser, JTextField locationField) {
+  private void chooseLocation(FileChooser fileChooser, JTextField locationField, String preferenceKey) {
```
lognaturel: I don't quite understand what this is for. Is it maybe for when each form will be able to have its own configuration? Looks like it's not used now.

ggalmazor:
Closing this PR because I've realized that all the code that handles persisted preferences through the `BriefcasePreferences` class is removed as a result of #268 and we're going to have to rework it.

PR #267, opened 2018-01-11
Issue 258 configuration per form

ggalmazor:
Partially addresses #258

This PR adds the possibility to set a custom export configuration per selected form.

*Warning*: I'm creating the PR just to start the discussion about this feature. I have yet to solve some layout issues (darn Swing!!) and the git commit history has to be pimped up.

Now, the export button will be enabled whenever there is at least one selected form and:
 - The user has set up a default configuration using the controls on the Export tab
OR
 - The user has set up a custom configuration for **every** selected form

Setting a default configuration and custom configurations for specific forms is also valid and Briefcase will use the custom confs wherever they're defined and the default conf otherwise.

This PR also comes with improvements and corrections on the enable/disable logic of buttons:
 - The detail button is only enabled if there is a status history to show
 - The row's selection checkbox and the custom conf button are enabled only if the form is selected

## Testing and design

I have explored the idea of creating Swing UI components and isolating controller classes from view classes that have the swingy thingy stuff.

The idea is to improve the design to be able to unit test the business logic in isolation of its UI layer. I have written some tests following this idea.

On the other hand, the UI layer could be tested driving the UI in some end-to-end tests like the one proposed in #263

Although I'm not happy quite yet with the results, this is something that we're going to be evolving over weeks/months and I think that this work could be merged as is to open up discussion and reflection about testing strategies.

Run automated tests (yay!)
Manually tested

N/A

N/A


ggalmazor:
I'm closing this PR while I finish work as we have discussed on Slack...

PR #268, opened 2018-01-12
Issue 258 configuration per form

ggalmazor:
Partially addresses #258

This PR adds the possibility to set a custom export configuration per selected form.

Now, the export button will be enabled whenever there is at least one selected form and:
 - The user has set up a default configuration using the controls on the Export tab
OR
 - The user has set up a custom configuration for **every** selected form

Setting a default configuration and custom configurations for specific forms is also valid and Briefcase will use the custom confs wherever they're defined and the default conf otherwise.

This PR also comes with improvements and corrections on the enable/disable logic of buttons:
 - The detail button is only enabled if there is a status history to show
 - The row's selection checkbox and the custom conf button are enabled only if the form is selected

## Testing and design

I have explored the idea of creating Swing UI components and isolating controller classes from view classes that have the swingy thingy stuff.

The idea is to improve the design to be able to unit test the business logic in isolation of its UI layer. I have written some tests following this idea.

On the other hand, the UI layer could be tested driving the UI in some end-to-end tests like the one proposed in #263

Although I'm not happy quite yet with the results, this is something that we're going to be evolving over weeks/months and I think that this work could be merged as is to open up discussion and reflection about testing strategies.

## GUI `.form` files

This PR comes with some new `.form` files that define the Swing UI. These files are to be used with IntelliJ's GUI designer.

Run automated tests (yay!)
Manually tested

N/A

N/A


src/org/opendatakit/briefcase/ui/export/components/FormsTableViewModel.java
```diff
@@ -0,0 +1,163 @@
...
+        } finally {
+          button.setEnabled(true);
+        }
+      });
+    }
+    return button;
+  }
+
+  JButton buildOverrideConfButton(FormStatus form) {
+    JButton button = new JButton("⚙");
```
dcbriccetti: The ⚙ is not appearing for me.
screen shot 2018-01-13 at 3 41 15 pm

src/org/opendatakit/briefcase/ui/export/components/ConfigurationDialogForm.java
```diff
@@ -0,0 +1,189 @@
...
+
+    setDefaultCloseOperation(DO_NOTHING_ON_CLOSE);
+
+    addWindowListener(new WindowAdapterBuilder().onClosing(e -> dispose()).build());
+
+    dialog.registerKeyboardAction(e -> dispose(), getKeyStroke(VK_ESCAPE, 0), WHEN_ANCESTOR_OF_FOCUSED_COMPONENT);
+  }
+
+  public void open() {
+    setVisible(true);
```
dcbriccetti: For me, this shows the dialog far away from the main window.

not centered

If you’ll add this here,
`setLocationRelativeTo(null);`

You can have such beauty as this:
centered

And, I think good taste (if not some style guide somewhere) suggests having a title in all dialogs.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationDialogForm.java
```diff
@@ -0,0 +1,189 @@
...
+
+    okButton.addActionListener(e -> dispose());
+    removeButton.addActionListener(e -> dispose());
+    cancelButton.addActionListener(e -> dispose());
+
+    setDefaultCloseOperation(DO_NOTHING_ON_CLOSE);
+
+    addWindowListener(new WindowAdapterBuilder().onClosing(e -> dispose()).build());
+
+    dialog.registerKeyboardAction(e -> dispose(), getKeyStroke(VK_ESCAPE, 0), WHEN_ANCESTOR_OF_FOCUSED_COMPONENT);
```
dcbriccetti: I’m not too concerned about it, but in case you’re not aware, registerKeyboardAction has this comment:

**This method is now obsolete, please use a combination of getActionMap() and getInputMap() for similar behavior.**

src/org/opendatakit/briefcase/ui/export/ExportConfiguration.java
```diff
@@ -0,0 +1,177 @@
...
+
+  public void ifDateRangeEndPresent(Consumer<LocalDate> consumer) {
+    endDate.ifPresent(consumer);
+  }
+
+  private List<String> getErrors() {
+    List<String> errors = new ArrayList<>();
+
+    if (!exportDir.isPresent())
+      errors.add("Export directory was not specified.");
```
dcbriccetti: Must an export directory be supplied on the custom config dialog? What if the user just wants to override the date range?

src/org/opendatakit/briefcase/ui/export/ExportConfiguration.java
```diff
@@ -0,0 +1,177 @@
...
+  public void ifDateRangeStartPresent(Consumer<LocalDate> consumer) {
+    startDate.ifPresent(consumer);
+  }
+
+  public void ifDateRangeEndPresent(Consumer<LocalDate> consumer) {
+    endDate.ifPresent(consumer);
+  }
+
+  private List<String> getErrors() {
+    List<String> errors = new ArrayList<>();
```
dcbriccetti: This code may be in flux, but at present this method is only called from here, and the errors themselves are never used.
```
  public boolean isValid() {
    return getErrors().isEmpty();
  }
```

src/org/opendatakit/briefcase/ui/export/ExportForms.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  public ExportForms(List<FormStatus> forms) {
+    this.forms = forms;
+    refreshIndex();
+  }
+
+
+  public void merge(List<FormStatus> forms) {
+    this.forms.addAll(forms.stream().filter(form -> !this.formsIndex.containsKey(form.getFormDefinition().getFormId())).collect(toList()));
```
dcbriccetti: `this.formsIndex` could be `formsIndex`

src/org/opendatakit/briefcase/ui/export/ExportForms.java
```diff
@@ -0,0 +1,102 @@
...
+    configurations.remove(form);
+  }
+
+  public void setConfiguration(FormStatus form, ExportConfiguration configuration) {
+    configurations.put(form, configuration);
+  }
+
+  public void selectAll() {
+    forms.forEach(form -> form.setSelected(true));
+  }
```
dcbriccetti: Don’t you miss Scala, where you could `_.setSelected` instead of `form -> form.setSelected`? Java’s making some steps in the right direction, though.

src/org/opendatakit/briefcase/ui/export/ExportForms.java
```diff
@@ -0,0 +1,102 @@
...
+  public void appendStatus(BriefcaseFormDefinition formDefinition, String statusUpdate, boolean successful) {
+    getForm(formDefinition).setStatusString(statusUpdate, successful);
+  }
+
+  private FormStatus getForm(BriefcaseFormDefinition formDefinition) {
+    return Optional.ofNullable(formsIndex.get(formDefinition.getFormId()))
+        .orElseThrow(() -> new RuntimeException("Form " + formDefinition.getFormName() + " " + formDefinition.getFormId() + " not found"));
+  }
+
+  private void refreshIndex() {
```
dcbriccetti: A small point, but this is really a `rebuildIndex`, rather than refreshing an index in place.

dcbriccetti:
Push custom config button for a row. Cancel. Then export.
```
Exception in thread "Thread-6" java.lang.RuntimeException: Wrong export configuration
	at org.opendatakit.briefcase.ui.export.ExportPanel.lambda$export$63(ExportPanel.java:120)
	at org.opendatakit.briefcase.ui.export.ExportPanel$$Lambda$105/1941374365.get(Unknown Source)
	at java.util.Optional.orElseThrow(Optional.java:290)
	at org.opendatakit.briefcase.ui.export.ExportPanel.export(ExportPanel.java:120)
	at org.opendatakit.briefcase.ui.export.ExportPanel.lambda$export$62(ExportPanel.java:105)
	at org.opendatakit.briefcase.ui.export.ExportPanel$$Lambda$100/1504597683.apply(Unknown Source)
	at java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:267)
```

dcbriccetti:
On Push tab, select a form and push Push, with URL empty.
```
[pool-1-thread-1] ERROR org.opendatakit.briefcase.util.TransferAction - upload transfer action failed
java.lang.NullPointerException
	at org.opendatakit.briefcase.util.AggregateUtils.testServerConnectionWithHeadRequest(AggregateUtils.java:409)
	at org.opendatakit.briefcase.util.ServerUploader.uploadForm(ServerUploader.java:278)
	at org.opendatakit.briefcase.util.ServerUploader.uploadFormAndSubmissionFiles(ServerUploader.java:206)
	at org.opendatakit.briefcase.util.TransferToServer.doAction(TransferToServer.java:41)
	at org.opendatakit.briefcase.util.TransferAction$UploadTransferRunnable.run(TransferAction.java:58)
	at java.lang.Thread.run(Thread.java:745)
```

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -16,412 +16,125 @@
...
-            .addPreferredGap(ComponentPlacement.RELATED)
-            .addGroup(groupLayout.createParallelGroup(TRAILING)
-                .addComponent(btnSelectAll)
-                .addComponent(btnClearAll)
-                .addComponent(btnExport))
-            .addContainerGap());
+    form.onExport(() -> new Thread(() -> {
+      List<String> errors = export(confPanel.getConfiguration());
+      if (!errors.isEmpty()) {
+        String message = String.format("%s\n\n%s", "We have found some errors while performing the requested export actions:", errors.stream().map(e -> "- " + e).collect(joining("\n")));
```
dcbriccetti: Line is very long

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -16,412 +16,125 @@
...
-  private List<String> export() {
-    return tableModel.getSelectedForms().parallelStream()
+  private List<String> export(ExportConfiguration defaultConfiguration) {
+    form.disableUI();
+    terminationFuture.reset();
+    List<String> errors = forms.getSelectedForms().parallelStream()
         .peek(FormStatus::clearStatusHistory)
         .map(formStatus -> (BriefcaseFormDefinition) formStatus.getFormDefinition())
-        .flatMap(formDefinition -> this.export(formDefinition).stream())
+        .flatMap(formDefinition -> this.export(defaultConfiguration, formDefinition).stream())
```
dcbriccetti: Should work without `this.`, yes?

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -16,412 +16,125 @@
...
         .collect(toList());
+    form.enableUI();
+    return errors;
   }
 
-  private List<String> export(BriefcaseFormDefinition formDefinition) {
+  private List<String> export(ExportConfiguration defaultConfiguration, BriefcaseFormDefinition formDefinition) {
```
dcbriccetti: Should this method and the one above both have the name `export`? Perhaps either or both could have a more descriptive name. (This method is called only by the one above with the same name.)

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -16,412 +16,125 @@
...
     List<String> errors = new ArrayList<>();
-    Optional<File> pemFile = Optional.ofNullable(pemPrivateKeyFilePath.getText()).map(File::new).filter(File::exists);
+    ExportConfiguration conf = forms.getConfiguration(formDefinition).orElse(defaultConfiguration);
+    Optional<File> pemFile = conf.mapPemFile(Path::toFile).filter(File::exists);
     if ((formDefinition.isFileEncryptedForm() || formDefinition.isFieldEncryptedForm()) && !pemFile.isPresent())
       errors.add(formDefinition.getFormName() + " form requires is encrypted and you haven't defined a valid private key file location");
```
dcbriccetti: Can we improve this language, while we’re in here? “form requires is encrypted”. Maybe “form is encrypted”?

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -16,412 +16,125 @@
...
-            new File(txtExportDirectory.getText().trim()),
+            conf.mapExportDir(Path::toFile).orElseThrow(() -> new RuntimeException("Wrong export configuration")),
             ExportType.CSV,
             formDefinition,
             pemFile.orElse(null),
             terminationFuture,
-            pickStartDate.convert().getDateWithDefaultZone(),
-            pickEndDate.convert().getDateWithDefaultZone()
+            conf.mapStartDate((LocalDate ld) -> Date.from(ld.atStartOfDay(ZoneId.systemDefault()).toInstant())).orElse(null),
+            conf.mapEndDate((LocalDate ld) -> Date.from(ld.atStartOfDay(ZoneId.systemDefault()).toInstant())).orElse(null)
```
dcbriccetti: This code is complex enough, and duplicated, so how about a method?
```
conf.mapStartDate(this::dateFromLocal).orElse(null),
conf.mapEndDate  (this::dateFromLocal).orElse(null)
```
```
private Date dateFromLocal(LocalDate ld) {
  return Date.from(ld.atStartOfDay(ZoneId.systemDefault()).toInstant());
}
```

src/org/opendatakit/briefcase/ui/export/components/FormsTableViewModel.java
ggalmazor: It seems that Lucida Sans doesn't support `GEAR` unicode char... I will have to find an alternative...

ggalmazor: Any suggestions for buttons with icons in Swing? I'd prefer to use an icon for this to save some horizontal pixels on that column...

ggalmazor: Strangely, I can see the glyph on win10/java8...

src/org/opendatakit/briefcase/ui/export/components/ConfigurationDialogForm.java
ggalmazor: Thanks!

src/org/opendatakit/briefcase/ui/export/components/ConfigurationDialogForm.java
ggalmazor: Honestly, I'd wait for a `@Deprecated` annotation on that method to change it... 

Also, this line was generated by IntelliJ's GUI designer... I'm guessing that it's safe to use.

Let's keep it :)

src/org/opendatakit/briefcase/ui/export/ExportConfiguration.java
ggalmazor: That's a good point. We've tried to start simple and work on the smallest possible feature set and on this PR we're focusing on the simplest form of "custom configurations per form". 

I don't know if that's something that user's would want to do but it feels something that we could add on further PRs... what do you think, @yanokwa @ln?

src/org/opendatakit/briefcase/ui/export/ExportConfiguration.java
ggalmazor: That's true. It has got there as a result of refactoring the previous version. Since we're going to work on UI notifications and logging during the next weeks, let's leave it that way for the moment and come back to it after we discuss how we're going to solve those things across the entire app.

src/org/opendatakit/briefcase/ui/export/ExportForms.java
ggalmazor: So true!

In Java8 I like a more radical functional approach of currying a function: 

```java
Function2<Boolean,FormStatus,Void> setSelected = (value, form) -> form.setSelected(value)
```

and then using it on the `forEach` like so:

```java
forms.forEach(setSelected(true));
```

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
ggalmazor: So, here's what I've done: I've moved the low level export method (the one that does the exporting) to `ExportAction` class to be more coherent. This `ExportPanel` shouldn't have to know so many details about exporting forms and this way we can work the design around `ExportAction`, `ExportToCSV`, etc. classes

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
ggalmazor: Now that this is on `ExportAction` class, what I'd really want to do is to migrate the uses of `Date` to `LocalDate` to avoid that ugly conversion...

src/org/opendatakit/briefcase/ui/export/components/ConfigurationDialog.java
```diff
@@ -0,0 +1,46 @@
...
+        form.enableOK();
+      else
+        form.disableOK();
+    });
+  }
+
+  static ConfigurationDialog from(ExportConfiguration configuration) {
+    ConfigurationPanel confPanel = ConfigurationPanel.from(configuration);
+    ConfigurationDialogForm form = new ConfigurationDialogForm(confPanel.getForm());
+    form.setBounds(100, 100, 450, 250);
```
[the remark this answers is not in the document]
ggalmazor: Thanks!

ggalmazor:
Thanks for your thorough review, @dcbriccetti!

You reported a `RuntimeException` when "Push custom config button for a row. Cancel. Then export."... I've been unable to reproduce that... could you give more context, please?

Regarding the other crash on the Push tab, I think we should file it as an independent issue to avoid mixing commits for another tabs on this PR...

yanokwa:
On my mac (Java 1.8.0_152), I don't see a list of forms to export with commit 42be6e1.

ggalmazor:
You should see it now. I remove a `this` reference by mistake on my previous commit

src/org/opendatakit/briefcase/ui/export/ExportConfiguration.java
lognaturel: I agree that it should be possible to only override the date or the PEM file. Will file as a separate issue.

src/org/opendatakit/briefcase/ui/export/components/FormsTableViewModel.java
lognaturel: This seems to be a Mac issue. We can file separately.

PR #272, opened 2018-01-16
Issue 258 save preferences

ggalmazor:
Partially addresses #258 

Changes on this PR solve loading and saving export configurations (default and per-form) into user preferences using `BriefcasePreferences` class.

Manual tests

N/A

None I'm aware of.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/export/ExportForms.java
+++ b/src/org/opendatakit/briefcase/ui/export/ExportForms.java
@@ -3,28 +3,39 @@ package org.opendatakit.briefcase.ui.export;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
 
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import org.opendatakit.briefcase.model.BriefcaseFormDefinition;
+import org.opendatakit.briefcase.model.BriefcasePreferences;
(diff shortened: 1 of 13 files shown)


src/org/opendatakit/briefcase/ui/export/ExportForms.java
```diff
@@ -3,28 +3,39 @@
...
+    this.configurations = configurations;
     rebuildIndex();
   }
 
+  public static ExportForms load(List<FormStatus> forms, BriefcasePreferences preferences) {
+    // This should be a simple Map filtering block but we'll have to wait for Vavr.io
+
+    Map<FormStatus, ExportConfiguration> configurations = new HashMap<>();
+    forms.forEach(form -> {
+      ExportConfiguration configuration = ExportConfiguration.load(preferences, "custom_" + form.getFormName() + "_");
```
lognaturel: I hadn't considered before now that Briefcase uses form name a lot (directory naming, etc). That's potentially dangerous as there's no uniqueness guarantee but I suppose there isn't one for form_id across multiple servers, either. Using form name here is the consistent way to go but thought I'd mention this just in case we run into issues related to it at some point.

ggalmazor: Maybe we should use UUID values. I haven't reviewed which method Aggregate uses to generate its UIDs but they are quite like UUIDs... Those could be used safely across multiple servers.

PR #273, opened 2018-01-17
Issue 258 last export date

ggalmazor:
Partially addresses #258 

This PR is built on top of #272 because it uses the Preferences saving/loading infrastructure for the Export tab

briefcase_last_export_date

Now the forms table has a new column that will track the last export date between launches.

Export dates will be tracked while using both the GUI and CLI command.

The export date will be formatted according to the user's locale.

Manual tests

N/A

None I'm aware of


lognaturel:
Works well. 👍 Can you please rebase on master so we can keep history clean when merging?

ggalmazor:
Good to go :)

yanokwa:
@ggalmazor I'm happy to merge this after the conflicts are resolved!

ggalmazor:
Ready to go :)

PR #274, opened 2018-01-17
Issue 265 clear PEM file field

### LINKED ISSUE #265: Filesystem location text fields should be resetable
Currently, once you select a location for a file or a directory, the corresponding text field holds a value an can never go back to hold nothing.

This makes sense for required fields like the export dir in the export tab or the storage location dir in the configuration tab.

On the other hand, in other fields like the pem file location, the user should be able to clear that field to configure the export operation to not use any pem file.

#### Software versions 
N/A

#### Problem description
N/A

N/A

#### Expected behavior
N/A

#### Other information 
N/A

ggalmazor:
Closes #265

briefcase_clear_buttonb

After setting the location for a PEM file, the button switches to a 'Clear' button that will clear the field.

Manual tests

N/A 

None I'm aware of

### DIFF
--- a/src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.form
+++ b/src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.form
@@ -88,15 +88,6 @@
           <text value="Choose..."/>
         </properties>
       </component>
-      <component id="97f45" class="javax.swing.JButton" binding="pemFileButton">
-        <constraints>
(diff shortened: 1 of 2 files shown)


ggalmazor:
> Works great, but I think Clear should align to the left, not the center.

I've prepared some snapshots of different options we could go for:

Option 1) full width, text centered

briefcase_clear_option_1

Option 2) full width, text to the left

briefcase_clear_option_2

Option 3) adjusted width, button to the left

briefcase_clear_option_3

Which one would you want me to commit?

yanokwa:
Thanks for giving me the choices! Let's go with option 3 and get this merged.

PR #279, opened 2018-01-19
Use work instead of src to reduce confusion when SSHing into CircleCI

yanokwa:


ggalmazor:
If it's for the working directory, I'd like to suggest `work`

PR #280, opened 2018-01-19
CodeCov support

### LINKED ISSUE #277: Add coverage report
Now that we're starting to have some tests, it would be helpful to have some metrics about code coverage.

This information would be also helpful on CircleCI

yanokwa:
Closes #277

I confirmed that the reports post to CircleCI and to my CodeCov URL.

I tried Coveralls and CodeClimate and both were confusing and hard to set up. CodeCov was pretty straightforward and I believe it's more maintainable.

I have not verified this works on master, but I'm hoping to merge and iterate.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,8 @@ jobs:
           background: true
       - restore_cache:
           key: jars-{{ checksum "build.gradle" }}
-      - run: ./gradlew check jar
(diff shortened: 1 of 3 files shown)


ggalmazor:
Sounds good to me :)

yanokwa:
@lognaturel this is ready for your merge!

PR #281, opened 2018-01-19
Issue 271 custom conf export dir requirement

### LINKED ISSUE #271: Export directory should not be required for form-level configuration
#### Software versions 
Briefcase v1.8.0-113-g50d5b8d9

#### Problem description
Currently setting a custom configuration for a form export requires specifying an export directory. This means that only override the date range or the PEM file also requires setting the same export directory as in the main configuration. 

Go to the export tab, select a form. Click on the gear icon to change its export configuration. Only set a PEM file or a date range. The configuration can't be saved.

#### Expected behavior
It should be possible to override only some of the settings.

#### Other information 
I see two options:
- make all custom configuration options optional
- if no custom configuration is specified, pre-populate fields with values from the main configuration and allow the user to change the ones they want

CC @ggalmazor @dcbriccetti @yanokwa

### LINKED ISSUE #286: Global export preferences are not persisted on clear
In the global export settings of master (3b29d73ad7c688276eccb44f463fe0fb64395494), if PEM file location or start/end date is cleared, that clearing is not persisted. If I relaunch the app, the value  still exists. If instead I change the value, that value is persisted.

ggalmazor:
Closes #271
Closes #286 

On this PR, we have changed the export configuration to "fall back" to a default configuration, in order to make all the fields on the custom configuration dialog optional. All the logic behind this fall back system is behind `ExportConfiguration`.

I've also reviewed configuration handling everywhere and fixed some consistency issues with the UI and user preferences.

This PR also makes some minor aesthetic changes (dialog title, text on remove button, etc.)

Manual tests and some automated unit test

N/A 

None


yanokwa:
Gave the functionality a run through and it works well for me. One change I'd like to see is the ability to clear the custom Export directory. This is needed because unlike in the global export, you can have this be empty and the only way to do it now is to clear all and then come back and set PEM, start/end.

ggalmazor:
The only thing that doesn't make me completely happy is the behavior of the clear all and ok buttons when you come from a custom conf that wasn't empty from the beginning.

- The OK button will be enabled only if the form is not empty and any entered data is valid. 
- The clear all button will be enabled if the configuration wasn't empty when opening the dialog.

If, for example, I had set a PEM file and now I wan't to clear it:

1. I click on the popup button
1. Clear all is enabled, OK is enabled
1. I click on Clear on the PEM file field
1. Clear all is enabled **but** OK is disabled (empty form)

PR #284, opened 2018-01-22
Update jar-in-jar loader files to make Briefcase runnable with Java9

### LINKED ISSUE #220: Briefcase should run on Java 9 machines
Verify that Briefcase works with Java 9 (JDK and JRE). If it doesn't, then suggest a fix!

ggalmazor:
Closes #220 

This PR updates Briefcase dependency for Eclipse's JAR-in-JAR loader which has been updated to properly run under Java9.

Manual tests with both Java8 (8u161-oracle) and Java9.
Successfully exported an encrypted form in both JVMs.

Jshell and other packaging methods like creating an uberjar were considered. Although more work on this front can be done, this solution follows the steps of what we were previously doing and "it just works".

None I'm aware of


yanokwa:
NIcely done, @ggalmazor!

PR #285, opened 2018-01-22
Upgrade Gradle

yanokwa:
Upgrading because
* Always nice to use new versions
* Collect, via Android Studio, uses Gradle 4.x
* Gradle 3.x doesn't work with Java 9. Errors out with `Could not determine java version from '9.0.4'.`


ggalmazor:
My bad, sorry! I forgot to update this along with the new jar in jar zip file...

PR #288, opened 2018-01-23
Set an icon for the custom conf popup buttons

ggalmazor:
Closes #270

This PR adds a pair of new icons from Material Design to replace the text representation of a gear that wasn't working on Macs

#### What has been done to verify that this works as intended?
Manual tests

#### Why is this the best possible solution? Were any other approaches considered?
N/A

#### Are there any risks to merging this code? If so, what are they?
N/A

PR #294, opened 2018-01-24
Use font for gear, change colors, fix default color

yanokwa:
Closes #288 

Confirmed working on macOS and Windows as a packaged jar. 

Fonts are easier to scale and color than PNGs, so I tried Emoji fonts, but the files were big and the icons not that good looking. icomoon let me upload a SVG from the Android project and generate the font I used.

As far as FontUtils, I'd be open to move that functionality elsewhere.


ggalmazor:
Does this thing cache the fonts?
Since we're always going to use the same font, we could move the `Font` instance to a static property in `FontUtils`.
I guess the best place for `FontUtils` right now would be the `org.opendatakit.briefcase.ui.reused` package

PR #308, opened 2018-01-25
Stabilize UI tests

ggalmazor:
This PR brings headless testing for our Swing UI tests.

It does so by running the tests with [CacioCavallo](http://hg.openjdk.java.net/caciocavallo/ng/), which in turn needs the quite old [FEST-swing](https://github.com/alexruiz/fest-swing-1.x) library.

This is something that could solve the test hanging problems by itself but more work could be necessary 

Cacio uses a [GNU license](http://hg.openjdk.java.net/caciocavallo/ng/file/5f250d540211/LICENSE) and FEST uses [Apache 2 license](https://github.com/alexruiz/fest-swing-1.x/blob/master/fest-swing-junit/LICENSE.txt).

N/A

N/A

N/A


ggalmazor:
So, it seems that those 2 tests are still failing on macs. It must be because we're using different file dialogs on macs and windows hosts. 

Since CircleCI build is working, I'm thinking about filing an issue to fix this after the release. What do you think @dcbriccetti @yanokwa @lognaturel?

ggalmazor:
I'd merge this PR, though, as it makes the test suite harder to fail due to problems with interactions with the desktop and mouse

yanokwa:
We can't use GPL libraries because it puts the project at legal risk. See https://github.com/opendatakit/briefcase/blob/master/CONTRIBUTING.md#code-from-external-sources.

I propose we comment out the failing tests on Macs for now so we can make forward progress on the rest of our tests.

ggalmazor:
Ok!

PR #315, opened 2018-01-25
Fix tests that fail on windows

ggalmazor:
This PR fixes a test to avoid problems with windows's path separators

N/A

N/A

N/A


ggalmazor:
I tried `File.pathSeparator` without luck. Let me give your option a quick spin...

yanokwa:
pathSeparator is for PATH separation (: or ;) and separator is for directory separation (/ or \). It's super confusing.

ggalmazor:
Doh!

ggalmazor:
Ok, using `File.separator` now

PR #326, opened 2018-01-29
Fix #323 - Add date range limit to date pickers.

### LINKED ISSUE #323: Set proper date range values for start and end date
#### Software versions 
Briefcase vv1.8.0-227-gbebd3f5, Java v1.8, Ubuntu, Aggregate v1.x.x, Collect v1.x.x...

#### Problem description
There is no restriction of the date range values in the start date and end date datepickers.

Open any datepicker.

#### Expected behavior
The start date should not exceed current date and when a particular start date is selected, the end date values less than the start date should be disabled instead of giving an error message afterwards.

#### Other information 
Set appropriate date range in the DatePickerProperties.

Can I work on this issue?

mayank8318:
Closes #323 

Built and ran the program.

It is best to disable end date till you enter start date, thus eliminating one logical error.

I tested for all possible logical scenarios that I could think of and it works.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
+++ b/src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
@@ -15,14 +15,13 @@
  */
 package org.opendatakit.briefcase.ui.export.components;
 
-import static org.opendatakit.briefcase.ui.ODKOptionPane.showErrorDialog;
 import static org.opendatakit.briefcase.ui.StorageLocation.isUnderBriefcaseFolder;
 import static org.opendatakit.briefcase.ui.reused.FileChooser.directory;
 import static org.opendatakit.briefcase.ui.reused.FileChooser.file;
 import static org.opendatakit.briefcase.util.FileSystemUtils.isUnderODKFolder;
 
 import com.github.lgooddatepicker.components.DatePicker;
-import com.github.lgooddatepicker.zinternaltools.DateChangeEvent;
+
 import java.awt.Component;
 import java.awt.FlowLayout;
 import java.awt.GridBagConstraints;
@@ -42,7 +41,6 @@ import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.JTextField;
 
-import org.opendatakit.briefcase.ui.MessageStrings;
 import org.opendatakit.briefcase.ui.reused.FileChooser;
 import org.opendatakit.briefcase.util.StringUtils;
 
@@ -89,28 +87,16 @@ public class ConfigurationPanelForm extends JComponent {
         buildPemFileDialog().choose().ifPresent(file -> setPemFile(Paths.get(file.toURI())))
     );
     pemFileClearButton.addActionListener(__ -> clearPemFile());
-    startDatePicker.addDateChangeListener(event -> {
-      if (!isDateRangeValid()) {
-        startDatePicker.closePopup();
-        startDatePicker.clear();
-        showError(MessageStrings.INVALID_DATE_RANGE_MESSAGE, "Export configuration error");
-      } else
-        onSelectStartDateCallbacks.forEach(consumer -> consumer.accept(extractDate(event)));
-    });
-    endDatePicker.addDateChangeListener(event -> {
-      if (!isDateRangeValid()) {
-        endDatePicker.closePopup();
-        endDatePicker.clear();
-        showError(MessageStrings.INVALID_DATE_RANGE_MESSAGE, "Export configuration error");
-      } else
-        onSelectEndDateCallbacks.forEach(consumer -> consumer.accept(extractDate(event)));
-    });
-  }
 
-  private boolean isDateRangeValid() {
-    LocalDate startDate = startDatePicker.getDate();
-    LocalDate endDate = endDatePicker.getDate();
-    return startDate == null || endDate == null || startDate.isBefore(endDate);
+    startDatePicker.addDateChangeListener(event -> {
+      endDatePicker.getSettings().setDateRangeLimits(event.getNewDate(), null);
+      onSelectStartDateCallbacks.forEach(consumer -> consumer.accept(event.getNewDate()));
+    });
+
+    endDatePicker.addDateChangeListener(event -> {
+      startDatePicker.getSettings().setDateRangeLimits(null, event.getNewDate());
+      onSelectEndDateCallbacks.forEach(consumer -> consumer.accept(event.getNewDate()));
+    });
   }
 
   @Override
@@ -184,10 +170,6 @@ public class ConfigurationPanelForm extends JComponent {
     onSelectEndDateCallbacks.add(callback);
   }
 
-  protected void showError(String message, String title) {
-    showErrorDialog(container, message, title);
-  }
-
   private void createUIComponents() {
     // Custom creation of components occurs inside the constructor
   }
@@ -226,14 +208,6 @@ public class ConfigurationPanelForm extends JComponent {
     return Optional.ofNullable(textField.getText())
         .filter(StringUtils::nullOrEmpty)
         .map(path -> Paths.get(path).toFile());
-  }
-
-  private static LocalDate extractDate(DateChangeEvent event) {
-    return event.getNewDate() != null ? LocalDate.of(
-        event.getNewDate().getYear(),
-        event.getNewDate().getMonthValue(),
-        event.getNewDate().getDayOfMonth()
-    ) : null;
   }
 
   /**
--- a/test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
+++ b/test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
@@ -17,6 +17,7 @@ package org.opendatakit.briefcase.ui.export.components;
 
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThat;
 import static org.opendatakit.briefcase.ui.matchers.SwingMatchers.empty;
 import static org.opendatakit.briefcase.ui.matchers.SwingMatchers.visible;
@@ -68,20 +69,14 @@ public class ConfigurationPanelTest extends AssertJSwingJUnitTestCase {
   }
 
   @Test
-  public void shows_error_and_clears_field_after_inserting_an_invalid_start_date() {
-    component.setEndDate(LocalDate.of(2018, 1, 1));
-    component.setStartDate(LocalDate.of(2020, 1, 1));
-    assertThat(component.errorDialog(500), is(GenericUIMatchers.visible()));
-    component.acceptErrorDialog();
-    assertThat(component.startDateField(), is(empty()));
+  public void cannot_insert_an_end_date_before_the_set_start_date() {
+    component.setStartDate(LocalDate.of(2017, 1, 30));
+    assertFalse(component.endDatePicker().isDateAllowed(LocalDate.of(2017, 1, 28)));
   }
 
   @Test
-  public void shows_error_and_clears_field_after_inserting_an_invalid_end_date() {
-    component.setStartDate(LocalDate.of(2020, 1, 1));
-    component.setEndDate(LocalDate.of(2018, 1, 1));
-    assertThat(component.errorDialog(500), is(GenericUIMatchers.visible()));
-    component.acceptErrorDialog();
-    assertThat(component.endDateField(), is(empty()));
+  public void cannot_insert_a_start_date_after_the_set_end_date() {
+    component.setEndDate(LocalDate.of(2017,1,25));
+    assertFalse(component.startDatePicker().isDateAllowed(LocalDate.of(2017,1,27)));
   }
 }
--- a/test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelUnitTest.java
+++ b/test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelUnitTest.java
@@ -15,10 +15,8 @@
  */
 package org.opendatakit.briefcase.ui.export.components;
 
-import static com.github.npathai.hamcrestopt.OptionalMatchers.isEmpty;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.is;
-import static org.hamcrest.Matchers.nullValue;
 import static org.junit.Assert.assertThat;
 
 import java.io.IOException;
@@ -62,34 +60,6 @@ public class ConfigurationPanelUnitTest {
     assertThat(view.getPemFile(), is(initialConfiguration.getPemFile().get()));
     assertThat(view.getDateRangeStart(), is(initialConfiguration.getStartDate().get()));
     assertThat(view.getDateRangeEnd(), is(initialConfiguration.getEndDate().get()));
-  }
-
-  @Test
-  public void shows_an_error_and_resets_the_field_when_the_UI_tries_to_set_a_date_range_end_that_is_before_its_start() {
-    FakeConfigurationPanelForm view = new FakeConfigurationPanelForm(false);
-    ConfigurationPanel panel = new ConfigurationPanel(ExportConfiguration.empty(), view);
-    view.setStartDate(LocalDate.of(2019, 1, 1));
-
-    // Trigger the error
-    view.setEndDate(LocalDate.of(2018, 1, 1));
-
-    assertThat(view.errorShown, is(true));
-    assertThat(view.endDatePicker.getDate(), nullValue());
-    assertThat(panel.getConfiguration().getEndDate(), isEmpty());
-  }
-
-  @Test
-  public void shows_an_error_and_resets_the_field_when_the_UI_tries_to_set_a_date_range_start_that_is_after_its_end() {
-    FakeConfigurationPanelForm view = new FakeConfigurationPanelForm(false);
-    ConfigurationPanel configurationPanel = new ConfigurationPanel(ExportConfiguration.empty(), view);
-    view.setEndDate(LocalDate.of(2018, 1, 1));
-
-    // Trigger the error
-    view.setStartDate(LocalDate.of(2019, 1, 1));
-
-    assertThat(view.errorShown, is(true));
-    assertThat(view.startDatePicker.getDate(), nullValue());
-    assertThat(configurationPanel.getConfiguration().getStartDate(), isEmpty());
   }
 
   @Test
--- a/test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelPageObject.java
+++ b/test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelPageObject.java
@@ -19,6 +19,7 @@ import static javax.swing.SwingUtilities.invokeLater;
 import static org.assertj.swing.edt.GuiActionRunner.execute;
 import static org.assertj.swing.timing.Timeout.timeout;
 
+import com.github.lgooddatepicker.components.DatePicker;
 import java.awt.event.ActionEvent;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -89,12 +90,12 @@ class ConfigurationPanelPageObject {
     return component.form.pemFileField;
   }
 
-  public JTextComponent startDateField() {
-    return component.form.startDatePicker.getComponentDateTextField();
+  public DatePicker startDatePicker() {
+    return component.form.startDatePicker;
   }
 
-  public JTextComponent endDateField() {
-    return component.form.endDatePicker.getComponentDateTextField();
+  public DatePicker endDatePicker() {
+    return component.form.endDatePicker;
   }
 
   public void setSomePemFile() {
@@ -124,10 +125,6 @@ class ConfigurationPanelPageObject {
     uncheckedSleep(50);
   }
 
-  private DialogFixture errorDialog() {
-    return errorDialog(50);
-  }
-
   public DialogFixture errorDialog(int timeoutMillis) {
     // Similar to buttonByName(name) or textFieldByName(name), we won't
     // throw when exhausting the timeout for obtaining a dialog and we will
@@ -154,14 +151,6 @@ class ConfigurationPanelPageObject {
     } catch (WaitTimedOutError e) {
       return null;
     }
-  }
-
-  public void acceptErrorDialog() {
-    // We need to be sure that the dialog is rendered before trying to click its button
-    waitFor(() -> errorDialog(500) != null && errorDialog().button() != null);
-    errorDialog().button().click();
-    // We wait for the dialog to disappear before handing over the thread
-    waitFor(() -> errorDialog() == null);
   }
 
   public void cancelFileDialog() {
--- a/test/java/org/opendatakit/briefcase/ui/export/components/FakeConfigurationPanelForm.java
+++ b/test/java/org/opendatakit/briefcase/ui/export/components/FakeConfigurationPanelForm.java
@@ -28,11 +28,6 @@ class FakeConfigurationPanelForm extends ConfigurationPanelForm {
   }
 
   @Override
-  protected void showError(String message, String title) {
-    errorShown = true;
-  }
-
-  @Override
   public void setEnabled(boolean enabled) {
     this.enabled = enabled;
   }


src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -74,12 +73,15 @@
     startDatePicker = createDatePicker();
     endDatePicker = createDatePicker();
     $$$setupUI$$$();
+    startDatePicker.getSettings().setDateRangeLimits(null, LocalDate.now());
+    startDatePicker.setDateToToday();
```
ggalmazor: I think we shouldn't set the default start date to any value...

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -74,12 +73,15 @@
...
     startDatePicker.getSettings().setGapBeforeButtonPixels(0);
     startDatePicker.getComponentDateTextField().setPreferredSize(exportDirField.getPreferredSize());
     startDatePicker.getComponentToggleCalendarButton().setPreferredSize(exportDirChooseButton.getPreferredSize());
     endDatePicker.getSettings().setGapBeforeButtonPixels(0);
     endDatePicker.getComponentDateTextField().setPreferredSize(exportDirField.getPreferredSize());
     endDatePicker.getComponentToggleCalendarButton().setPreferredSize(exportDirChooseButton.getPreferredSize());
+    endDatePicker.setEnabled(false);
```
ggalmazor: This changes previous behavior, where we let the user select an end date **before** a start date.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -89,26 +91,14 @@
...
-    });
-    endDatePicker.addDateChangeListener(event -> {
-      if (!isDateRangeValid()) {
-        showError(MessageStrings.INVALID_DATE_RANGE_MESSAGE, "Export configuration error");
-        endDatePicker.clear();
-      } else
-        onSelectEndDateCallbacks.forEach(consumer -> consumer.accept(extractDate(event)));
+      endDatePicker.setEnabled(true);
+      endDatePicker.getSettings().setDateRangeLimits(extractDate(event), null);
+      setEndDate(extractDate(event));
```
ggalmazor: Updating the end date could pass unnoticed by the user and feels a little bit weird even when you are setting trying to set a date after the established end date

mayank8318:
@ggalmazor Sorry, realised that now. Will update the PR soon.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
mayank8318: ok but it has to be before the current date right?

ggalmazor:
Awesome, thanks! I'm on Slack should you need help on anything ;)

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
mayank8318: But then we were checking for empty start date during submission anyways, so why not disable the end date itself till the user sets a start date?

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
mayank8318: Ok then I will clear it? Is that fine?

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
ggalmazor: That makes sense but I wonder if we could have problems with forms pulled from different time zones... Maybe a form is from a timezone with a positive offset that effectively marks it as being created "in the future" respect to the host's timezone.

We can sit on this one and wait for feedback from @lognaturel or @yanokwa to confirm this bit.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
ggalmazor: I'm ok with disabling it, but I think it's worth to mention that this changes behavior, just to be sure that we're all ok with this change... 

I don't really know if users want to set the end before the start, but it's supported on the current version. It might be safe to change it, though.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
ggalmazor: Maybe the best option would be to clear the end date only if it's before the start date we're trying to set. What do you think?

ggalmazor:
Also, remove the unused  `ConfigurationPanelForm.showError()` method [commit line here](https://github.com/opendatakit/briefcase/pull/326/commits/7fd90248b9d07b49291b65ff8e6c35339f4cc0d6?diff=unified#diff-9014a675cb101d05fd895ad99eab07d5R175), please

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
mayank8318: That situation wont be possible as such dates would be disabled already.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
mayank8318: Okay

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
ggalmazor: Hmmm... If I set start to `2018-01-01`, the end date to `2018-01-20`, then I can set the start to `2018-01-30`, producing the situation I'm talking about. Then I would clear the end field.

If the second time I set the start date to `2018-01-10` instead, I wouldn't do anything, because the date range is valid.

mayank8318: Oh okay sorry didn't think of this. Yes, this I will handle and clear the end date.

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Nafundi
+ * Copyright (C) 2017 Nafundi
```
ggalmazor: This should read 2018

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
```diff
@@ -68,20 +70,32 @@ public void choose_pem_file_button_is_swapped_for_a_clean_button_that_cleans_the
...
+    component.setStartDate(LocalDate.of(2017,1,27));
+    assertEquals(component.endDatePicker().getDate(), null);
+  }
+
+  @Test
+  public void new_start_date_is_before_or_on_end_date_is_valid() {
+    component.setStartDate(LocalDate.of(2017,1,20));
+    component.setEndDate(LocalDate.of(2017,1,25));
+    component.setStartDate(LocalDate.of(2017,1,24));
+    assertEquals(component.endDatePicker().getDate(), LocalDate.of(2017,1,25));
```
ggalmazor: Please, change these `assertEquals` to `assertThat(actualValue, is(expectedValue))` like the on the rest of the test suite.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -156,6 +152,10 @@ private void clearPemFile() {
     pemFileClearButton.setVisible(false);
   }
 
+  private boolean isInvalidStartDate(LocalDate date) {
```
ggalmazor: I'd suggest renaming this to `isAfterEndDate(LocalDate date)`

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -89,26 +90,21 @@
...
-        showError(MessageStrings.INVALID_DATE_RANGE_MESSAGE, "Export configuration error");
-        startDatePicker.clear();
-      } else
-        onSelectStartDateCallbacks.forEach(consumer -> consumer.accept(extractDate(event)));
-    });
-    endDatePicker.addDateChangeListener(event -> {
-      if (!isDateRangeValid()) {
-        showError(MessageStrings.INVALID_DATE_RANGE_MESSAGE, "Export configuration error");
+
+      if (event.getNewDate() == null || isInvalidStartDate(extractDate(event)))
```
ggalmazor: Every call to `extractDate(event)` could be replaced by `event.getNewDate()` because now the datepicker returns a `java.time.LocalDate` directly (we were using an old version before that used a third party library for dates).

As a result, the `extractDate` method could be removed.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -89,26 +90,21 @@
...
         endDatePicker.clear();
-      } else
-        onSelectEndDateCallbacks.forEach(consumer -> consumer.accept(extractDate(event)));
+
+      endDatePicker.setEnabled(event.getNewDate() != null);
```
ggalmazor: If you're up for that, all these `null` checks could be replaced by using a little bit of `Optional<LocalDate> startDate = Optional.ofNullable(event.getNewDate())`, but we can do that on a future PR as well.

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
```diff
@@ -68,20 +70,32 @@ public void choose_pem_file_button_is_swapped_for_a_clean_button_that_cleans_the
...
   }
-}
\ No newline at end of file
+
+  @Test
+  public void clear_end_date_when_new_start_date_is_after_end_date() {
+    component.setStartDate(LocalDate.of(2017,1,22));
+    component.setEndDate(LocalDate.of(2017,1,25));
+    component.setStartDate(LocalDate.of(2017,1,27));
+    assertEquals(component.endDatePicker().getDate(), null);
```
ggalmazor: Please, change this with with `assertThat(value, is(nullValue()));`

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
```diff
@@ -68,20 +70,32 @@ public void choose_pem_file_button_is_swapped_for_a_clean_button_that_cleans_the
...
+  }
+
+  @Test
+  public void new_start_date_is_before_or_on_end_date_is_valid() {
+    component.setStartDate(LocalDate.of(2017,1,20));
+    component.setEndDate(LocalDate.of(2017,1,25));
+    component.setStartDate(LocalDate.of(2017,1,24));
+    assertEquals(component.endDatePicker().getDate(), LocalDate.of(2017,1,25));
+    component.setStartDate(LocalDate.of(2017,1,25));
+    assertEquals(component.endDatePicker().getDate(), LocalDate.of(2017,1,25));
```
ggalmazor: Same here: `assertThat(actualValue, is(expectedValue))`

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
```diff
@@ -68,20 +70,32 @@ public void choose_pem_file_button_is_swapped_for_a_clean_button_that_cleans_the
...
   }
 
   @Test
-  public void shows_error_and_clears_field_after_inserting_an_invalid_end_date() {
-    component.setStartDate(LocalDate.of(2020, 1, 1));
-    component.setEndDate(LocalDate.of(2018, 1, 1));
-    assertThat(component.errorDialog(500), is(GenericUIMatchers.visible()));
-    component.acceptErrorDialog();
-    assertThat(component.endDateField(), is(empty()));
+  public void empty_start_date_end_datepicker_is_disabled() {
```
ggalmazor: I'd suggest renaming the method to something more like an English phrase 
`end_datepicker_is_disabled_if_the_start_datepicker_is_empty`

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
```diff
@@ -68,20 +70,32 @@ public void choose_pem_file_button_is_swapped_for_a_clean_button_that_cleans_the
   }
 
   @Test
-  public void shows_error_and_clears_field_after_inserting_an_invalid_start_date() {
-    component.setEndDate(LocalDate.of(2018, 1, 1));
-    component.setStartDate(LocalDate.of(2020, 1, 1));
-    assertThat(component.errorDialog(500), is(GenericUIMatchers.visible()));
-    component.acceptErrorDialog();
-    assertThat(component.startDateField(), is(empty()));
+  public void cannot_insert_an_invalid_end_date() {
```
ggalmazor: I'd suggest renaming this to explicitly tell the condition which makes impossible to insert an invalid date: `cannot_insert_an_end_date_after_the_established_start_date` or something like that

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
```diff
@@ -68,20 +70,32 @@ public void choose_pem_file_button_is_swapped_for_a_clean_button_that_cleans_the
...
-    assertThat(component.endDateField(), is(empty()));
+  public void empty_start_date_end_datepicker_is_disabled() {
+    component.startDatePicker().clear();
+    assertFalse(component.endDatePicker().isEnabled());
   }
-}
\ No newline at end of file
+
+  @Test
+  public void clear_end_date_when_new_start_date_is_after_end_date() {
```
ggalmazor: Rename `end_date_is_cleared_when_the_new_start_date_is_after_it`

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
```diff
@@ -68,20 +70,32 @@ public void choose_pem_file_button_is_swapped_for_a_clean_button_that_cleans_the
...
+    component.setStartDate(LocalDate.of(2017,1,22));
+    component.setEndDate(LocalDate.of(2017,1,25));
+    component.setStartDate(LocalDate.of(2017,1,27));
+    assertEquals(component.endDatePicker().getDate(), null);
+  }
+
+  @Test
+  public void new_start_date_is_before_or_on_end_date_is_valid() {
```
ggalmazor: Rename: `end_date_remains_unchanged_if_it_is_after_the_new_start_date`

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelUnitTest.java
```diff
@@ -65,31 +66,37 @@ public void it_initializes_UI_fields_with_values_from_the_initial_configuration(
   }
 
   @Test
-  public void shows_an_error_and_resets_the_field_when_the_UI_tries_to_set_a_date_range_end_that_is_before_its_start() {
+  public void cannot_insert_an_invalid_end_date() {
```
ggalmazor: I think these tests are safe to remove because the UI tests on `ConfigurationPanelTest` are easier to understand and they're testing the same things...

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelUnitTest.java
```diff
@@ -65,31 +66,37 @@ public void it_initializes_UI_fields_with_values_from_the_initial_configuration(
...
-    ConfigurationPanel panel = new ConfigurationPanel(ExportConfiguration.empty(), view);
-    view.setStartDate(LocalDate.of(2019, 1, 1));
+    view.setStartDate(LocalDate.of(2017, 1, 30));
+    assertFalse(view.endDatePicker.isDateAllowed(LocalDate.of(2017, 1, 28)));
+  }
 
-    // Trigger the error
-    view.setEndDate(LocalDate.of(2018, 1, 1));
+  @Test
+  public void empty_start_date_end_datepicker_is_disabled() {
```
ggalmazor: Suggest removing test

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelUnitTest.java
```diff
@@ -65,31 +66,37 @@ public void it_initializes_UI_fields_with_values_from_the_initial_configuration(
...
+    FakeConfigurationPanelForm view = new FakeConfigurationPanelForm(false);
+    view.startDatePicker.clear();
+    assertFalse(view.endDatePicker.isEnabled());
+  }
 
-    assertThat(view.errorShown, is(true));
+  @Test
+  public void clear_end_date_when_new_start_date_is_after_end_date() {
```
ggalmazor: Suggest removing test

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelUnitTest.java
```diff
@@ -65,31 +66,37 @@ public void it_initializes_UI_fields_with_values_from_the_initial_configuration(
...
+    view.setStartDate(LocalDate.of(2017,1,22));
+    view.setEndDate(LocalDate.of(2017,1,25));
+    view.setStartDate(LocalDate.of(2017,1,27));
     assertThat(view.endDatePicker.getDate(), nullValue());
-    assertThat(panel.getConfiguration().getEndDate(), isEmpty());
   }
 
   @Test
-  public void shows_an_error_and_resets_the_field_when_the_UI_tries_to_set_a_date_range_start_that_is_after_its_end() {
+  public void new_start_date_is_before_or_on_end_date_is_valid() {
```
ggalmazor: Suggest removing test

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
mayank8318: Done

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
mayank8318: Done

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
mayank8318: Done

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
mayank8318: Renamed to `cannot_insert_an_end_date_before_the_set_start_date`

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
mayank8318: Renamed to `end_datepicker_is_disabled_if_the_start_datepicker_is_empty`

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
mayank8318: Done

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
mayank8318: Done

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
mayank8318: Done

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
mayank8318: Done

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
mayank8318: Done

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelUnitTest.java
mayank8318: Ok

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelUnitTest.java
mayank8318: Done

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelUnitTest.java
mayank8318: :+1:

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelUnitTest.java
mayank8318: :+1:

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
mayank8318: I guess a new PR for that would be apt since it is a different issue :)

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelUnitTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Nafundi
+ * Copyright (C) 2017 Nafundi
```
ggalmazor: Should read 2018

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -89,26 +88,19 @@
...
-    endDatePicker.addDateChangeListener(event -> {
-      if (!isDateRangeValid()) {
-        showError(MessageStrings.INVALID_DATE_RANGE_MESSAGE, "Export configuration error");
+
+      if (event.getNewDate() != null && isAfterEndDate(event.getNewDate()))
         endDatePicker.clear();
-      } else
-        onSelectEndDateCallbacks.forEach(consumer -> consumer.accept(extractDate(event)));
+
+      if (event.getNewDate() != null)
```
ggalmazor: You can remove the `if` block and directly set the date range limits with whatever comes in `event.getNewDate()` This way, when the user clears the field, the restriction is lifted.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -89,26 +88,19 @@
...
-        showError(MessageStrings.INVALID_DATE_RANGE_MESSAGE, "Export configuration error");
-        startDatePicker.clear();
-      } else
-        onSelectStartDateCallbacks.forEach(consumer -> consumer.accept(extractDate(event)));
-    });
-    endDatePicker.addDateChangeListener(event -> {
-      if (!isDateRangeValid()) {
-        showError(MessageStrings.INVALID_DATE_RANGE_MESSAGE, "Export configuration error");
+
+      if (event.getNewDate() != null && isAfterEndDate(event.getNewDate()))
```
ggalmazor: Don't clear the field. Let's go for only applying restrictions to datepickers :)

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -89,26 +88,19 @@
...
+
+      onSelectStartDateCallbacks.forEach(consumer -> consumer.accept(event.getNewDate()));
     });
-  }
 
-  private boolean isDateRangeValid() {
-    LocalDate startDate = startDatePicker.getDate();
-    LocalDate endDate = endDatePicker.getDate();
-    return startDate == null || endDate == null || startDate.isBefore(endDate);
+    endDatePicker.addDateChangeListener(event -> onSelectEndDateCallbacks.forEach(consumer -> consumer.accept(event.getNewDate())));
```
ggalmazor: On this change listener we're missing a `startDatePicker.getSettings().setDateRangeLimits(null, event.getNewDate());` to apply restrictions to the start datepicker.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -89,26 +88,19 @@
...
-      if (!isDateRangeValid()) {
-        showError(MessageStrings.INVALID_DATE_RANGE_MESSAGE, "Export configuration error");
+
+      if (event.getNewDate() != null && isAfterEndDate(event.getNewDate()))
         endDatePicker.clear();
-      } else
-        onSelectEndDateCallbacks.forEach(consumer -> consumer.accept(extractDate(event)));
+
+      if (event.getNewDate() != null)
+        endDatePicker.getSettings().setDateRangeLimits(event.getNewDate().plusDays(1), null);
```
ggalmazor: You don't have to add a day to the date because we need to be able to select the same day on both datepickers

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -156,6 +148,10 @@ private void clearPemFile() {
     pemFileClearButton.setVisible(false);
   }
 
+  private boolean isAfterEndDate(LocalDate date) {
```
ggalmazor: After the changes above, this method will not be used. Remove.

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
```diff
@@ -68,20 +70,26 @@ public void choose_pem_file_button_is_swapped_for_a_clean_button_that_cleans_the
...
+    component.setStartDate(LocalDate.of(2017,1,22));
+    component.setEndDate(LocalDate.of(2017,1,25));
+    component.setStartDate(LocalDate.of(2017,1,27));
+    assertThat(component.endDatePicker().getDate(), is(nullValue()));
   }
-}
\ No newline at end of file
+
+  @Test
+  public void end_date_remains_unchanged_if_it_is_after_the_new_start_date() {
```
ggalmazor: Remove this test

test/java/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelTest.java
```diff
@@ -68,20 +70,26 @@ public void choose_pem_file_button_is_swapped_for_a_clean_button_that_cleans_the
...
   }
 
   @Test
-  public void shows_error_and_clears_field_after_inserting_an_invalid_end_date() {
-    component.setStartDate(LocalDate.of(2020, 1, 1));
-    component.setEndDate(LocalDate.of(2018, 1, 1));
-    assertThat(component.errorDialog(500), is(GenericUIMatchers.visible()));
-    component.acceptErrorDialog();
-    assertThat(component.endDateField(), is(empty()));
+  public void end_date_is_cleared_when_the_new_start_date_is_after_it() {
```
ggalmazor: Remove this test

mayank8318:
image
@ggalmazor We don't need these pop us now right?

ggalmazor:
> @ggalmazor We don't need these pop-ups now right?
That's right :)

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -89,28 +88,23 @@
...
+
     startDatePicker.addDateChangeListener(event -> {
-      if (!isDateRangeValid()) {
-        showError(MessageStrings.INVALID_DATE_RANGE_MESSAGE, "Export configuration error");
-        startDatePicker.clear();
-      } else
-        onSelectStartDateCallbacks.forEach(consumer -> consumer.accept(extractDate(event)));
+
+      if (event.getNewDate() != null)
+        endDatePicker.getSettings().setDateRangeLimits(event.getNewDate().plusDays(1), null);
```
ggalmazor: This block should be the same as on the `endDatePicker`:
- `... setDateRangeLimits(event.getNewDate(), null)`
- No need for the `if/else` case, no need for adding 1 day to the date

mayank8318:
:) More to follow ;)

lognaturel:
@mayank8318, this is a nice usability improvement.

There's a strange glitch on macOS, java 1.8.0_112. If I pick a start date and then go back one month I end up with something like:
screen shot 2018-02-02 at 7 16 36 pm

I can reset the interface by hovering over the month or year at the top of the menu.

mayank8318:
@lognaturel Sorry I don't have access to a Mac here.

ggalmazor:
I've also reproduced that glitch on a mac with Java 1.8.0_161... 

The date picker comes from a third party library. I think we should at least file an issue. 

@yanokwa @lognaturel do you think we should wait to merge this PR until the glitch is solved on the date picker library?

ggalmazor:
I think this issue is the same one we're experiencing: https://github.com/LGoodDatePicker/LGoodDatePicker/issues/61

A user is proposing a hack to solve it (basically triggering a MouseEntered on the year field to force a repaint)

lognaturel:
Sounds fine to merge if it's Mac only! Let's wait for another round of manual verification and merge after 1.9.0 comes out.

kkrawczyk123:
Everything looks good on Ubuntu and Windows, but on MacOS I encountered the same problem as @lognaturel.

kkrawczyk123:

PR #336, opened 2018-02-02
Issue 258 save connection password

### LINKED ISSUE #258: Support for bulk form exports

ggalmazor:
Closes #258 

This PR adds a new checkbox on the Settings tab to get user's permission to save server connection passwords on preferences. This way, the password fields won't be empty between launcher.

This PR tries to solve this by modifying the least possible amount of code and without changing the design and it prepares our work to support the "pull before export" feature for the Export tab.

Note that the pull and push server connection settings are stored separately and setting them for one tab won't set them for the other.

Manual tests on Linux and Win10. Can't get my hands on a mac to test it yet, sorry!
@yanokwa, @kkrawczyk123 could you test this one on a Mac plz?

I didn't work on any automated tests because this part of Briefcase hasn't been refactored yet to be easily testable.

I considered refactoring the SettingsPanel to its own vertical and doing the same work we did for the ExportPanel (own `.form` file for IntelliJ's GUI Designer, separation of vertical/reused artifacts) but the code is really entangled with `BriefcasePreference`'s, `FileSystemUtils`'s and `StorageLocation`'s static API and I don't see an easy way to achieving this without a huge redesign with lots of possible side-effects.

N/A


kkrawczyk123:
@ggalmazor as you wish, I checked it on all three systems and it's look good.

ggalmazor:
Thanks!

src/org/opendatakit/briefcase/ui/SettingsPanel.java
```diff
@@ -24,83 +22,97 @@
...
+    chkProxy.setSelected(false);
+    chkProxy.addActionListener(new ProxyToggleListener());
+
+    chkParallel.setSelected(BriefcasePreferences.getBriefcaseParallelPullsProperty());
+    chkParallel.addActionListener(new ParallelPullToggleListener());
+
+    chkTrackingConsent.setSelected(BriefcasePreferences.getBriefcaseTrackingConsentProperty());
+    chkTrackingConsent.addActionListener(new TrackingConsentToggleListener(parentWindow.briefcaseAnalytics));
+
+    chkStorePasswordsConsent.setSelected(BriefcasePreferences.getStorePasswordsConsentProperty());
```
dcbriccetti: Maybe add a tooltip text? 
`chkStorePasswordsConsent.setToolTipText("Passwords are saved in plaintext in the local file system");`
I don’t know if we have a policy on using them.

ggalmazor:
Thanks for the review @dcbriccetti! I've updated the checkbox text to what @yanokwa suggested.

PR #349, opened 2018-02-05
Upgrade to the latest minor release of libraries

### LINKED ISSUE #348: Upgrade libraries to the latest minor release
#### Software versions 
Briefcase v1.9.0

#### Problem description
Briefcase uses libraries that are out of date. Out of date libraries can have bugs that have been resolved, so we should generally be on the latest stable release.

We have to do this incrementally to avoid breakage and minor upgrades to libraries very early in our release cycle with extensive testing is the best way to do this.

yanokwa:
Closes #348 

Tried a pull and an export.

These are minor upgrades to libraries, it should be safe.

The upgrades break something. This is why we want to do this early in the release cycle and get it tested.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,13 +39,13 @@ configurations.all {
 
 dependencies {
     compile group: 'org.bouncycastle', name: 'bcprov-jdk16', version: '1.46'
-    compile group: 'commons-cli', name: 'commons-cli', version: '1.3.1'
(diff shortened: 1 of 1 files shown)


yanokwa:
@ggalmazor Since this is a library change, can you do quick test of this to make sure logging, HTTP connections, mockito, etc, generally work?

PR #353, opened 2018-02-07
Issue 258 pull before export

ggalmazor:
Closes #258 (this is the last pending task)

This PR changes Briefcase to let the users automatically pull forms before exporting them. 

Provided that the user checks the "store passwords" setting on the Settings tab, Briefcase will store the Aggregate connection settings whenever a form is successfully pulled. On the Export tab, a new set of checkboxes let the users decide to pull the forms before exporting.

If a user chooses to pull before exporting a form, the status history of the form will have log lines for both operations (revealed with the Detail button).

briefcase_pull_before

Some specific details and use cases: 
- On the **default** export configuration, a user can select a checkbox to **pull where it is possible** (aka, the transfer settings have been stored)
- On the **override** export configuration panel, a user can tell to **pull or to inherit** whatever behavior is set on the default export configuration
- When the user hasn't checked the "store passwords" setting on the Settings tab, the checkboxes are disabled, and a message is shown telling the user that she won't be able to pull until she permits to store the passwords.
- When we haven't stored the transfer settings yet, the checkboxes are disabled, and a message is shown telling the user to pull once the form to be able to pull on further exports automatically.

### Scenarios

| Panel type | Save pass | Transfer settings | Checkbox | Combo box | Hint |
| --- | --- | --- | --- | --- | --- |
| Default | no | no | disabled | not visible | visible, text about remembering passwords |
| Default | no | yes | disabled | not visible | visible, text about remembering passwords |
| Default | yes | no | enabled | not visible | visible, text about pulling the form |
| Default | yes | yes | enabled | not visible | visible, text about pulling the form |
| Override | no | no | not visible | disabled | visible, text about remembering passwords |
| Override | no | yes | not visible | disabled | visible, text about remembering passwords |
| Override | yes | no | not visible | disabled | visible, text about pulling the form |
| Override | yes | yes | not visible | enabled | not visible |

#### What has been done to verify that this works as intended?
Manual tests on Linux, Mac, and Win10
Added some automated tests.

#### Why is this the best possible solution? Were any other approaches considered?
N/A

#### Are there any risks to merging this code? If so, what are they?
None that I can think of.

PR #354, opened 2018-02-07
Briefcase analytics

ggalmazor:
This PR comes with Google Analytics tracking. The list of things we're tracking is:
- Open/close Briefcase
- Enter/leave all four tabs
- Success/failure at exporting forms
- Success/failure at pulling forms
- Success/failure at pushing forms
- Grant/revoke permission to track activity

Every hit is submitted with the following information attached:
- Briefcase version
- Unique User ID
- Anonymized IP address
- Operating System
- Screen resolution
- Briefcase window resolution

**Testing guidelines**
I guess one way to test this is to create a GA account and configure the GA Tracking ID in `build.gradle`. Then, browse the app with and without having checked the tracking consent on the Settings panel. You should see a lot of activity on the real-time report whenever tracking is enabled.

**About the User ID**
Briefcase will generate a User ID (a UUID) and store it using Java user preferences. If these preferences get removed, a new (different) User ID will be created.

**About the Operating System**
Google Analytics doesn't let us specify an OS manually. Instead, we are using three default user agent strings (one for each OS we want to track). This way, Google Analytics (hopefully) will be able to add OS info to each hit.

**About the GA Tracking ID**
I have created a GA account using my google account to be able to work this feature. For obvious reasons, this GA Tracking ID must not be used in the release versions of Briefcase.

Manual tests on Linux, Mac, and Win10

We could have used the previous code but the GA library [was updated recently and had breaking changes](https://github.com/brsanthu/google-analytics-java).

Tracking uses network resources. Although all hits are sent asynchronously (meaning GA shouldn't affect negatively to the user experience), maybe on scenarios with low bandwidth or data support, users will be forced to remove their consents.


dcbriccetti:
Did we take out the informed consent/opt out? Many people don’t like applications that “phone home”.

ggalmazor:
There’s the a checkbox on the Settings panel to opt-in for tracking.

dcbriccetti:
I’ll have to look into why I don’t see that setting when I run the program.

ggalmazor:
My bad! I messed up while applying the commits for git's log...

dcbriccetti:
Just completing the record here—the opt out is, in fact, not present in master now, but is restored with this change.

src/org/opendatakit/briefcase/ui/reused/Analytics.java
```diff
@@ -0,0 +1,127 @@
...
+    EventHit request = buildEventHit()
+        .eventCategory(category)
+        .eventAction(action);
+    Optional.ofNullable(label).ifPresent(request::eventLabel);
+    Optional.ofNullable(value).ifPresent(request::eventValue);
+    request.sendAsync();
+  }
+
+  public ComponentListener buildComponentListener(String screenName) {
+    return new ComponentListener() {
```
dcbriccetti: Aren’t these adapters great‽ See, I still have a few AWT/Swing tricks up my sleeve.

screen shot 2018-02-07 at 9 22 58 pm

src/org/opendatakit/briefcase/ui/reused/Analytics.java
```diff
@@ -0,0 +1,127 @@
...
+        .userId(userId)
+        .screenResolution(screenSize)
+        .anonymizeIp(true)
+        .userAgent(USER_AGENT)
+        .dataSource("app")
+        .viewportSize(buildResolutionString(windowSizeGetter.get()));
+  }
+
+  private static String buildResolutionString(Dimension dimension) {
+    return String.format("%sx%s", dimension.width, dimension.height);
```
dcbriccetti: I wonder if whatever takes this n × n string would like a real times/by character (**×**) instead of a crude, primitive x. I like to think that some people somewhere appreciate these fine touches.

test/java/org/opendatakit/briefcase/ui/reused/NopAnalytics.java
```diff
@@ -0,0 +1,69 @@
+package org.opendatakit.briefcase.ui.reused;
+
+import java.awt.event.ComponentEvent;
+import java.awt.event.ComponentListener;
+import javax.swing.JComponent;
+
+public class NopAnalytics extends Analytics {
+
+  private static final ComponentListener NOP_COMPONENT_LISTENER = new ComponentListener() {
```
dcbriccetti: Maximum benefit!

`private static final ComponentListener NOP_COMPONENT_LISTENER = new ComponentAdapter() {};`

test/java/org/opendatakit/briefcase/ui/reused/NopAnalytics.java
```diff
@@ -0,0 +1,69 @@
+package org.opendatakit.briefcase.ui.reused;
+
+import java.awt.event.ComponentEvent;
+import java.awt.event.ComponentListener;
+import javax.swing.JComponent;
+
+public class NopAnalytics extends Analytics {
```
dcbriccetti: I wonder if this wouldn’t sound more clear as NoOpAnalytics. I’ve never heard No-Op or NOP pronounced as a single syllable before. (Maybe you’re thinking spell it NopAnalytics and pronounce it no-op-analytics.)

src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java
```diff
@@ -213,9 +214,18 @@ public void setFullUIEnabled(boolean enabled) {
...
+
+        Analytics analytics = Analytics.from(
+            BuildConfig.GOOGLE_TRACKING_ID,
+            BuildConfig.VERSION,
+            BriefcasePreferences.getUniqueUserID(),
+            Toolkit.getDefaultToolkit().getScreenSize(),
+            frame::getSize
+        );
+        analytics.enter("Briefcase");
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> analytics.leave("Briefcase")));
```
dcbriccetti: How does the thread help us here?

src/org/opendatakit/briefcase/ui/SettingsPanel.java
```diff
@@ -88,17 +89,26 @@ public void actionPerformed(ActionEvent arg0) {
...
     chkStorePasswordsConsent.addActionListener(e -> {
-      if (e.getSource() == chkStorePasswordsConsent) {
-        boolean isSelected = chkStorePasswordsConsent.isSelected();
-        BriefcasePreferences.setStorePasswordsConsentProperty(isSelected);
-        if (!isSelected) {
-          PullTransferPanel.PREFERENCES.remove(BriefcasePreferences.PASSWORD);
-          PushTransferPanel.PREFERENCES.remove(BriefcasePreferences.PASSWORD);
-        }
+      boolean isSelected = chkStorePasswordsConsent.isSelected();
+      if (isSelected)
```
dcbriccetti: Come on, brother, I know you prefer this shorter style:
`analytics.event("Settings", "Save passwords", isSelected ? "Enable" : "Disable", null);`

src/org/opendatakit/briefcase/ui/SettingsPanel.java
```diff
@@ -88,17 +89,26 @@ public void actionPerformed(ActionEvent arg0) {
...
+        analytics.event("Settings", "Save passwords", "Enable", null);
+      else
+        analytics.event("Settings", "Save passwords", "Disable", null);
+      BriefcasePreferences.setStorePasswordsConsentProperty(isSelected);
+      if (!isSelected) {
+        PullTransferPanel.PREFERENCES.remove(BriefcasePreferences.PASSWORD);
```
dcbriccetti: Not sure I can sell this, and I bet you already considered it:
```
Arrays.asList(PullTransferPanel.PREFERENCES, PushTransferPanel.PREFERENCES).forEach(p -> 
    p.remove(BriefcasePreferences.PASSWORD));
```
Maybe if we already had a collection of the two panels, shared with other code.

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -44,11 +45,12 @@
   private final ExportForms forms;
   private final ExportPanelForm form;
 
-  public ExportPanel(TerminationFuture terminationFuture, ExportForms forms, ExportPanelForm form, BriefcasePreferences preferences) {
+  public ExportPanel(TerminationFuture terminationFuture, ExportForms forms, ExportPanelForm form, BriefcasePreferences preferences, Analytics analytics) {
```
dcbriccetti: This line is 155 characters. Do you have a limit? 😀

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -91,17 +93,20 @@ public ExportPanel(TerminationFuture terminationFuture, ExportForms forms, Expor
 
     form.onExport(() -> new Thread(() -> {
       List<String> errors = export();
+      int exportedFormsAmount = forms.getSelectedForms().size();
       if (!errors.isEmpty()) {
+        analytics.event("Export", "Export", "Configuration errors", exportedFormsAmount);
```
dcbriccetti: A small point. Feel free to ignore. I would code the isEmpty case first, for two reasons. I’ll bet you know what they are.

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -91,17 +93,20 @@ public ExportPanel(TerminationFuture terminationFuture, ExportForms forms, Expor
 
     form.onExport(() -> new Thread(() -> {
       List<String> errors = export();
+      int exportedFormsAmount = forms.getSelectedForms().size();
```
dcbriccetti: I gave up making the distinction between “less” and “fewer” (“10 items or less” at the market no longer bothers me), but I do make the distinction between number and amount. Is it the *number* of exported forms?

src/org/opendatakit/briefcase/ui/reused/Analytics.java
ggalmazor: Nice!

src/org/opendatakit/briefcase/ui/reused/Analytics.java
ggalmazor: I think GA requires an `x` :(

You're right with the %d... I use %s as a reflex...

test/java/org/opendatakit/briefcase/ui/reused/NopAnalytics.java
ggalmazor: I've seen both... Honestly, I don't have a strong opinion. Let's be explicit and change it to `NoOp` everywhere...

src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java
ggalmazor: I can't call `Runtime.addShutdownHook(Thread hook)` otherwise... I agree that it would be much nicer to just pass a `Runnable`.

Any alternatives?

src/org/opendatakit/briefcase/ui/SettingsPanel.java
ggalmazor: Good catch! Thanks!

src/org/opendatakit/briefcase/ui/SettingsPanel.java
ggalmazor: Don't worry about this lines. They look much better (and without breaking encapsulation) on #353 (see [here](https://github.com/opendatakit/briefcase/pull/353/files#diff-71c30114fc8ce489cd3b85b0a27e9872R98))

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
ggalmazor: The limit is your imagination! :D

It is good that this long line hurts. It's yelling at us that we need to rethink our dependencies. The bad fix IMO would be to place some line breaks...

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
ggalmazor: I don't reject the idea of not negating that condition

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
ggalmazor: Hmmm, I didn't know about that distinction... I'll research, thanks!

src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java
dcbriccetti: Sorry, I missed that addShutdownHook takes a Thread.

BTW, I played around with using listeners instead of this hackier approach, but it seems you have to do a bunch of Mac-specific stuff to hook into Quit (as opposed to clicking the close button in the JFrame).

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
dcbriccetti: I’m amused—and maybe impressed—that you think it’s worse to break this line.

build.gradle
```diff
@@ -111,7 +111,7 @@ buildConfig {
     version = getVersionName()
     clsName = 'BuildConfig'
     packageName = 'org.opendatakit.briefcase.buildconfig'
-    buildConfigField 'String', 'GOOGLE_TRACKING_ID', "UA-91951913-3"
+    buildConfigField 'String', 'GOOGLE_TRACKING_ID', "UA-113741271-1"
```
yanokwa: Revert to the old ID so we can merge. You now have read access to this account.

src/org/opendatakit/briefcase/ui/reused/Analytics.java
```diff
@@ -0,0 +1,113 @@
+package org.opendatakit.briefcase.ui.reused;
```
yanokwa: Please add copyright notice!

test/java/org/opendatakit/briefcase/ui/reused/NoOpAnalytics.java
```diff
@@ -0,0 +1,45 @@
+package org.opendatakit.briefcase.ui.reused;
```
yanokwa: Please add copyright notice!

src/org/opendatakit/briefcase/ui/SettingsPanel.java
```diff
@@ -88,17 +89,19 @@ public void actionPerformed(ActionEvent arg0) {
...
     chkStorePasswordsConsent.addActionListener(e -> {
-      if (e.getSource() == chkStorePasswordsConsent) {
-        boolean isSelected = chkStorePasswordsConsent.isSelected();
-        BriefcasePreferences.setStorePasswordsConsentProperty(isSelected);
-        if (!isSelected) {
-          PullTransferPanel.PREFERENCES.remove(BriefcasePreferences.PASSWORD);
-          PushTransferPanel.PREFERENCES.remove(BriefcasePreferences.PASSWORD);
-        }
+      boolean isSelected = chkStorePasswordsConsent.isSelected();
+      analytics.event("Settings", "Save passwords", isSelected ? "Enable" : "Disable", null);
```
yanokwa: What's the rationale behind tracking this?

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -91,7 +93,11 @@ public ExportPanel(TerminationFuture terminationFuture, ExportForms forms, Expor
 
     form.onExport(() -> new Thread(() -> {
       List<String> errors = export();
-      if (!errors.isEmpty()) {
+      int numberOfExportedForms = forms.getSelectedForms().size();
+      if (errors.isEmpty())
+        analytics.event("Export", "Export", "Success", numberOfExportedForms);
+      else {
+        analytics.event("Export", "Export", "Configuration errors", numberOfExportedForms);
```
yanokwa: For consistency, I'd expect Failure instead of Configuration errors. Are these really configuration errors?

Also, why is this only done with export and not with pull/push?

src/org/opendatakit/briefcase/ui/SettingsPanel.java
ggalmazor: I thought having some metrics on how many people are disabling or enabling this would be useful...

yanokwa: I want to make sure we know what we'll do with the data before we collect it. In the case of tracking consent, that's useful because we want people to consent and so knowing when people are not consenting can help us change the messaging. 

So I don't mind that we collect this, but I do want you to defend it by explaining what action will we take based on the data we get. So yeah, if 100% save passwords what does that mean to you? And what about 50% or 0%?

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
ggalmazor: I wanted to differentiate when an export operation can't be done because there are configuration errors, from general export failures (which  have their our "Failure" label)

Pull and Push operations don't have (yet) the same concept of a configuration object and its validness. Once we do that, we will have the same "Configuration errors" label for them as well.

I thought that monitoring this could give us a measure of how difficult is to configure each operation. If we see lots of conf error events, maybe we need to focus on making the conf process easier...

src/org/opendatakit/briefcase/ui/SettingsPanel.java
ggalmazor: I understand the problem. I think that having metrics of a new feature we are going to release it's always useful to measure its impact. 

Also, being this password stuff somewhat controversial, the data could tell us how the users are feeling about it. My bias is that the pros of storing passwords overweight the cons, but I don't know how our users value their privacy/security. 

If all our users save passwords, we could consider securing Briefcase a little more. If none use this, we could ask on the forums for feedback (are they not using it because it's insecure? would they use it if it was secure? are they even aware that there's that possibility? etc.).

Anyway, at this point, I guess it's more about adding the ability to track into Briefcase more than the specific things we want to track from day zero... I will remove the tracking If you consider it's better that way ;)

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
ggalmazor: Oh, man! :( I was mixing in my head this PR and #353!!!!! :( :( :(

You're right. This one should be labeled "Failure". After we merge #353 we can add a new event for conf errors...

yanokwa: I thought I was going crazy for a second. Please push the change and we can have the testers very this.

ggalmazor: I need a fresher head to review this properly... I'll do it first thing tomorrow! Sorry for the confusion!

yanokwa:
@ggalmazor I don't see the code for where you enter and leave tab.

ggalmazor:
@yanokwa the enter/leave tracking code is done by calling to `addComponentListener(analytics.buildComponentListener("Pull"));` (there is one of those on each tab).

The way I've tested the GA stuff takes 2 days to do so. On the first day, you can check that your actions produce hits on the GA backend real-time report. On the second day, those hits get aggregated and you can check the meta-data.

I haven't been able to check the meta-data in the real-time report and in the aggregate reports I haven't found the view-port size information... (I haven't any GA experience yet)

About what's been manually tested:
- Entering and leaving all 4 tabs
- Pull success: pull some forms
- Pull failure: pulling with a user with no permissions at all and you will get a failure event
- Push success: push some forms
- Push failure: I have been able to do this by making a mess of an individual submission in my storage dir and then pushing it. The UI will say "Partially successful upload..." but a failure event is sent
- Export success: export some forms
- Export failure: export the form used to test the push failure
- Export configuration errors: try to export an encrypted form without setting the pem file or setting a wrong file
- Export success with errors: This will only happen if some of the submission can't be decrypted. I don't understand how this could happen and I haven't been able to reproduce it...
- Toggling tracking consent: Just toggle it and you should get enable/disable events.

yanokwa:
View port information is Browser size. You can select that as a secondary dimension under Technology > Browser and OS.

gradle.properties
```diff
@@ -0,0 +1 @@
+googleAnalyticsTrackingId=UA-91951913-3
```
yanokwa: This is very close, but it'll get us in trouble during a release. 

I would keep the googleAnalyticsTrackingId as a variable in build.gradle, then set the value in build.gradle so it's the default. Then on the release machine we can have a gradle.properties file that overrides that value. That gradle.properties file will never be in Git so there's no change for leakage or merge conflicts.

This way, we can just do a `./gradle clean jar` regardless of where we are and the right thing will happen.

ggalmazor: Ok, I need to test how that works...

ggalmazor:

ggalmazor:
I've manually rebased the commits from this PR. Automatic rebase was giving lots of conflicts due to changes from recent PR merges.

@kkrawczyk123 there are no changes in behavior at all but I've fixed some extra bugs I've detected:
- Now we enable tracking on launch only if the user has checked the tracking setting. If this setting is not checked, next time we launch Briefcase, tracking should be disabled.
- Now we ensure that a "disabled tracking" event reaches GA when the user unchecks the tracking setting. Before this fix, we would only receive the "enable tracking" events.

kkrawczyk123:
I am not Google analytics ninja and yesterday I saw it for the first time in my life, but I did some tests, mostly Real-Time tab. Checked if actions are visible on the same device or if GA see properly actions from other (tried different combination between my linux mac and windows devices) and every time my actions were tracked. (Poland has the biggest activity in statistics now ;D ) I have noticed that if I close and open briefcase, GA consider me as different users so it tracks every open action even if I open it on the same computer. All push, pull, export actions and Grant/revoke permission to track activity are tracked and visible in event diagram and event table. I also looked into the content of Events Tab in Behavior section, everything seemed to be just fine.
I’ve noticed that amount of export action is the same as exported forms but push and pull is just counted by action no matter how many forms you pull or push. 
I don’t know if I can test something more in this pr, if yes tell me what to verify more. I assume that I checked most of the pivotal cases.

> - Now we enable tracking on launch only if the user has checked the tracking setting. If this setting is not checked, next time we launch Briefcase, tracking should be disabled.
> - Now we ensure that a "disabled tracking" event reaches GA when the user unchecks the tracking setting. Before this fix, we would only receive the "enable tracking" events.

I verified it and if this option is not cheked, none actions are tracked.

kkrawczyk123:

yanokwa:
@kkrawczyk123 You said, "I have noticed that if I close and open briefcase, GA consider me as different users so it tracks every open action even if I open it on the same computer. " 

Our expectation was that Briefcase will generate a user ID and store it using Java user preferences. If these preferences get removed, a different ID will be created. Is this not happening?

ggalmazor:
I think we're mixing two different things here:
- Every time a user opens (and closes) Briefcase, an event is sent to GA. That's what's expected, I think.
(which doesn't have anything to do with)
- Every hit sent to GA goes with a UserID attached, which will help us identify unique users between different launches of Briefcase. Every time the preferences get wiped, a new UserID is generated.

I don't know how to check the UserIDs on GA, though.

yanokwa:
User Ids are under Audience, User Explorer and I can confirm that the User ID does not change between launches (at least on Windows).

PR #356, opened 2018-02-08
Issue 209 review log, system.err, and system.out CLI interactions

### LINKED ISSUE #209: Provide command-line mode feedback to System.out

### LINKED ISSUE #245: There is no error message displayed when you try to pull from aggregate form that doesn't exist by using command line.
#### Software versions 
Briefcase v1.8.0, Java v1.8.0_151, operating system - Ubuntu

#### Problem description
There is no error message displayed when you try to pull from aggregate form that doesn't exist by using command line. 
Command line answers in the same way as when pulling form file which actually exists.
 
1. Go to command line.
2. Select command that allows you to pull form data from aggregate server with id of non existing form.

#### Expected behavior
There should show up any error message to inform user that pulling didn't succeed.

#### Other information 
http://docs.opendatakit.org/briefcase-forms/#working-with-command-line
Here you can find commands by which you can use briefcase by terminal.

ggalmazor:
Closes #209
Closes #245

This PR is a complete review of the interactions between the CLI and `System.err`, `System.out` and `Log`.

The general guidelines are:
- `System.out` is to be used to give positive or neutral feedback to the user
- `System.err` is to be used to give negative (errors and warnings) feedback to the user. We always want to use human-friendly, non-techie language (no stack traces allowed here).
- The different log levels of `Log` should be used for:
  - info: information about key steps of processes. Should be a few per CLI operation
  - warn: non critical errors aka errors that wouldn't stop process execution
  - error: critical errors that would stop process execution
  - debug: debug information. Should be quite a bunch but this is not to be abused. Only relevant debugging information should be logged.
- No other part of Briefcase other than the CLI operations can interact with `System.out` or `System.err`. Any feedback must be addressed through the `EventBus`.
  - This includes normal output or errors due to the multi-threaded nature of Briefcase's business processes.
  - This is why both the CLI and the Swing UI rely on the `EventBus` to know when a Pull/Push/Export operation fails or has success.

### Error handling and Exceptions
Wherever an error happens, an unchecked exception should be thrown. This will be more relevant as we modernize the source code to benefit from more modern concurrency mechanisms introduced in Java8.

All known exceptions should be lifted to a `BriefcaseException` or any children class.

Exceptions can be caught for three main reasons:
1. To translate a low-level exception class to a (higher-level) business exception. Also to transform a checked exception to an unchecked exception.
      - For example, translate an `IOException` into an `ExportException`.
2. To ignore them. It is recommended to log them at least.
3. Ultimately, to terminate the process gracefully.

The launchers handle differently any `BriefcaseException` that reaches them. These represent known error situations that can be avoided by the user (usually by fixing some configuration). Any other `Throwable` will produce a generic error message with links to Slack channels and the support forum.

### Testing this PR
The key here is that the CLI operation's output only contains feedback for the user and the logs don't have feedback for the users.

- To test this, make sure you build Briefcase with a simple logger conf file that will get all log lines into a file. You can do so by copying `res/simplelogger.properties.example` into `res/simplelogger.properties`
- Then you can run all the operations with valid and invalid parameters
- The console output should print some messages telling you what's happened
- When the parameters are wrong or the CLI operation fails, the program's exit code should be 1 (error)
- When the CLI operation fails, you should get a stack trace on your log file

### Logging
An SLF4J SimpleLogger configuration file example has been added to the project at `res/simplelogger.properties.example`. The example comes with a basic configuration that will write into a `briefcase.log` file created wherever Briefcase is launched.

There is also a complete configuration file with all the possible configuration keys at `res/simplelogger.properties.example_complete`

Manual tests of all CLI operations, using both the new and the old launchers.

The old launcher forces you to specify a formId to pull from ODK, but this is ignored by the process, as it pull every form on the ODK directory.

N/A

It feels like the `EventBus` library we're using has some concurrency problems. Sometimes, an event can be lost even though apparently, both event subscription and publish happens correctly. No updates to the library have been made since 2010 and we're already using the "latest" version. I think we would need to replace this library...


ggalmazor:
I have changed error handling to make it more simple and less confusing for the user. Before the last commit, all errors produced the "contact maintainers" text and that was confusing and wrong.

Now, error handling is based on exceptions. Only positive or neutral feedback uses the EventBus.

I'm going to change the PR text to show this.

src/org/opendatakit/briefcase/operations/CliEventsCompanion.java
```diff
@@ -0,0 +1,67 @@
...
+import org.opendatakit.briefcase.model.FormStatusEvent;
+import org.opendatakit.briefcase.model.TransferSucceededEvent;
+
+/**
+ * This class will handle Events while we change from an annotation based,
+ * static API Event bus to something more flexible.
+ */
+class CliEventsCompanion {
+  static void attach(Log log) {
+    on(ExportProgressEvent.class, (ExportProgressEvent event) -> {
```
dcbriccetti: Can we let the type be inferred on these, for easier reading?

`on(ExportProgressEvent.class, event -> {`

src/org/opendatakit/briefcase/operations/CliEventsCompanion.java
```diff
@@ -0,0 +1,67 @@
...
+import org.opendatakit.briefcase.model.TransferSucceededEvent;
+
+/**
+ * This class will handle Events while we change from an annotation based,
+ * static API Event bus to something more flexible.
+ */
+class CliEventsCompanion {
+  static void attach(Log log) {
+    on(ExportProgressEvent.class, event -> {
+      if (event.getText().contains("Cause:")) {
```
dcbriccetti: Looking for a string in a string here (to determine severity?) concerns me a bit, but I see from the comment above this code will be replaced.

src/org/opendatakit/briefcase/operations/Export.java
```diff
@@ -22,157 +22,102 @@
...
 import org.opendatakit.common.cli.Operation;
 import org.opendatakit.common.cli.Param;
 
 public class Export {
-  private static final Log LOGGER = LogFactory.getLog(Export.class);
-  private static final DateFormat DATE_FORMAT = new SimpleDateFormat("yyyy/MM/dd");
+  private static final Log log = LogFactory.getLog(Export.class);
   private static final Param<Void> EXPORT = Param.flag("e", "export", "Export a form");
-  private static final Param<String> EXPORT_DIR = Param.arg("ed", "export_directory", "Export directory");
+  private static final Param<Path> EXPORT_DIR = Param.arg("ed", "export_directory", "Export directory", Paths::get);
```
dcbriccetti: This modernizing is good to see.

src/org/opendatakit/briefcase/operations/CliEventsCompanion.java
ggalmazor: Yeah, also, that's part of the old event-based export progress notification system. That code should throw an exception instead of publishing an `ExportProgressEvent`... One day...

dcbriccetti:
I’ll have more this afternoon or evening (PST -8).

dcbriccetti:
I’m continuing my review of this, finally.

src/org/opendatakit/briefcase/util/ExportToCsv.java
```diff
@@ -74,42 +74,37 @@
...
-
-  public ExportToCsv(File outputDir, BriefcaseFormDefinition lfd, TerminationFuture terminationFuture, String filename, boolean exportMedia, Boolean overwrite, Date start, Date end) {
+  private final Map<TreeElement, OutputStreamWriter> fileMap = new HashMap<TreeElement, OutputStreamWriter>();
+  private final Map<String, String> fileHashMap = new HashMap<String, String>();
+  private final TerminationFuture terminationFuture;
+  private final File outputDir;
+  private final BriefcaseFormDefinition briefcaseLfd;
+  private final String baseFilename;
+  private final boolean exportMedia; // true by default?
+  private final boolean overwrite; // false by default?
```
dcbriccetti: I’m confused by these “by default?” comments.

src/org/opendatakit/briefcase/util/ExportToCsv.java
```diff
@@ -74,42 +74,37 @@
...
-
-  public ExportToCsv(File outputDir, BriefcaseFormDefinition lfd, TerminationFuture terminationFuture, Date start, Date end) {
-    this(outputDir, lfd, terminationFuture, lfd.getFormName(), true, false, start, end);
-  }
-
-  public ExportToCsv(File outputDir, BriefcaseFormDefinition lfd, TerminationFuture terminationFuture, String filename, boolean exportMedia, Boolean overwrite, Date start, Date end) {
+  private final Map<TreeElement, OutputStreamWriter> fileMap = new HashMap<TreeElement, OutputStreamWriter>();
+  private final Map<String, String> fileHashMap = new HashMap<String, String>();
+  private final TerminationFuture terminationFuture;
+  private final File outputDir;
```
dcbriccetti: I find it soothing to see these fields become private and final. 😺

src/org/opendatakit/briefcase/util/ExportToCsv.java
```diff
@@ -74,42 +74,37 @@
...
+    this.overwrite = overwrite;
+    this.startDate = startDate;
+    this.endDate = endDate;
     this.outputMediaDir = new File(outputDir, MEDIA_DIR);
-    this.briefcaseLfd = lfd;
-    this.terminationFuture = terminationFuture;
+  }
+
 
+  private String stripCSVExtension(String filename) {
```
dcbriccetti: Remember Josh Bloch’s admonition (do you need the audio file?—:grinning:) stripCsvExtension

src/org/opendatakit/briefcase/util/ExportToCsv.java
```diff
@@ -1008,13 +1003,14 @@ public FilesSkipped totalFilesSkipped() {
...
     log.info("exporting to : " + exportDir);
     ExportToCsv action = new ExportToCsv(
-        exportDir.toFile(),
-        formDefinition,
         new TerminationFuture(),
-        startDate.map((LocalDate ld) -> Date.from(ld.atStartOfDay(ZoneId.systemDefault()).toInstant())).orElse(null),
+        exportDir.toFile(),
+        formDefinition, baseFilename,
+        exportMedia,
+        overwriteFiles, startDate.map((LocalDate ld) -> Date.from(ld.atStartOfDay(ZoneId.systemDefault()).toInstant())).orElse(null),
```
dcbriccetti: Any reason to not use type inference here?
```
... startDate.map(ld -> ...,
endDate.map(ld -> ...
```

src/org/opendatakit/briefcase/util/ExportToCsv.java
ggalmazor: I forgot to remove those. They had those values by default prior to these changes and I commented before transforming that into a final variable...

dcbriccetti: I sometimes flag such temporary changes with `// todo remove` or some such.

dcbriccetti:
Yaw suggested splitting this up. Is that coming, or shall I continue reviewing this as is?

yanokwa:
@dcbriccetti This used to be one commit and my suggestion was to break it up which @ggalmazor did.

src/org/opendatakit/briefcase/util/ExportToCsv.java
dcbriccetti: Might even warrant a function, for clarity and to remove duplication?
```
overwriteFiles, dateOrNull(startDate), dateOrNull(endDate)
⋮
private static Date dateOrNull(Optional<LocalDate> localDate) {
  return localDate.map(ld -> Date.from(ld.atStartOfDay(ZoneId.systemDefault()).toInstant())).orElse(null);
}
```

src/org/opendatakit/briefcase/operations/Export.java
```diff
@@ -22,157 +22,105 @@
...
+        .findFirst();
 
-    if (formDefinition == null) {
-      LOGGER.error("Form not found");
-      return;
-    }
+    if (!maybeFormDefinition.isPresent())
+      throw new FormNotFoundException(formid);
+
+    BriefcaseFormDefinition formDefinition = maybeFormDefinition.get();
```
dcbriccetti: I’ve learned the painful lesson (in Scala)—several times—never to call get on an Option. Even when I could prove it’s non-empty. I’m new to Java’s implementation of these functional constructs, but I think this guideline might apply here, too. What do you think of the following change?

never call get

dcbriccetti: Wait—something better coming

dcbriccetti: I just discovered Optional.orElseThrow. Perfect for this use, yes?

ggalmazor: Nice catch! Yep, some commits before that, I needed to make a side effect (sending an event) that I wouldn't be ok to do on an `orElseThrow` but then I removed the side effect and forgot about that block. Thanks!

dcbriccetti: I was very pleased to learn from the investigation! It’s great to be learning with you.

src/org/opendatakit/briefcase/operations/Export.java
```diff
@@ -75,43 +75,32 @@
...
+    BriefcaseFormDefinition formDefinition = maybeFormDefinition.orElseThrow(() -> new FormNotFoundException(formid));
 
     if (formDefinition.isFileEncryptedForm() || formDefinition.isFieldEncryptedForm()) {
-      if (!pemFile.isPresent())
-        throw new BriefcaseException("Missing pem file configuration");
-
-      if (!Files.exists(pemFile.get()))
-        throw new BriefcaseException("Missing pem file configuration");
+      Path pemFile = maybePemFile
+          .filter(file -> !Files.exists(file))
```
dcbriccetti: Ooh, nice! (The filter)

ggalmazor: In Spanish, I'd say it's a "monad-a" ("monada" means cute)

dcbriccetti: Hey, there’s `notExists`!

ggalmazor: Now that I say it out loud, the word for monad is "mónada" and for cute (adj) "monada", which only have different accents (first syllable tonic and unstressed respectively)

src/org/opendatakit/briefcase/operations/Export.java
```diff
@@ -22,157 +22,94 @@
...
-        LOGGER.error("Briefcase action failed: No specified PrivateKey file for encrypted form");
-        return;
-      }
-      pemFile = new File(pemKeyFile.get());
-      if (!pemFile.exists()) {
-        LOGGER.error("Briefcase action failed: No PrivateKey file for encrypted form");
-        return;
-      }
+      Path pemFile = maybePemFile
+          .filter(Files::notExists)
```
dcbriccetti: Whoops! I think we got this backwards. You want to keep it if the file exists.
`.filter(Files::exists)`

ggalmazor: Fixed!

ggalmazor:

ggalmazor:

ggalmazor:

kkrawczyk123:
"Testing this PR" in the description was very helpful, thank you @ggalmazor! 
**Tested with success on all three systems!**

kkrawczyk123:

ggalmazor:

lognaturel:
A few high-level things I noticed:
- Send users to the forum only and not the Slack for better async issue handling; also the Slack URL currently points to a specific message
- Should be specified that log file will be in the same folder as the jar
- No human-friendly message for `MissingArgumentException`
- #378 

They're all relatively minor and can be handled separately so I'll file issues so #368 can be reworked on top of these nice improvements!

ggalmazor:
@lognaturel are you going to create some issues for those things you noticed? Do you want me to do it?

lognaturel:
I think I got them! #378, #379, #380.

PR #364, opened 2018-02-10
Remove aggregate 1.0 text

### LINKED ISSUE #228: Replace user facing ODK Aggregate 1.0 text with ODK Aggregate
We use "Aggregate 1.0" in the user-facing sections of ODK Briefcase and it's a bit of a misnomer. 

"Aggregate 1.0" is a stand-in for any server that implements the OpenRosa 1.0 API. To users, this basically means ODK Aggregate v1.x. For consistency with ODK Collect, I'd propose replacing the GUI parts of ODK Briefcase that say "Aggregate 1.0" with "ODK Aggregate". I'd also be OK with "ODK Aggregate 1.x"

This came to mind because we've removed all 0.9.x from the code (https://github.com/opendatakit/briefcase/issues/200).

icemc:
Closes #228

All user facing Aggregate 1.0 text was replaced wit ODK Aggregate
This is the only possible solution
No


icemc:
@yanokwa
Oops RAS is a French short form of "Rien a signaler" this is frequently used to mean "Nothing else" sorry about that. Will edit the PR.

ggalmazor:
There's still some strings with the "Aggregate" text like "Aggregate server URL". These don't have the "1.0" suffix but I wonder if we should say "ODK Aggregate server URL instead" on these as well.

yanokwa:
@ggalmazor I think we should file that as a separate (maybe quick win) issue. And in that issue, we can discuss when we call it "ODK Aggregate" and when we call it "Aggregate".

PR #365, opened 2018-02-12
Issue 334 Reset HTTP context whenever we get an 401 Not Authorized response

### LINKED ISSUE #334: Incorrect credentials are cached?
This is an issue I see with v1.9.beta-0.

To reproduce, pull from a server using bad credentials. Now, use the correct credentials. It'll still fail. Now quit, use the correct credentials, and it'll work. 

I believe the reverse is true as well. If I pull using correct credentials, then change out the connections and try to pull using the wrong credentials, it works when it should fail.

ggalmazor:
Closes #334

This PR changes the HTTP requests engine to reset any HTTP context whenever we get a 401 Not Authorized response from the servers.

This fix prevents the effects of the current coupling problems that exist between the different business operations that require HTTP interaction in Briefcase but a deeper redesign is required (as I explain on [a comment on the issue](https://github.com/opendatakit/briefcase/issues/334#issuecomment-364908421)).

This PR also fixes a previous situation where the user cancels the server connection settings dialog after having successfully pulled some forms. If the user hasn't checked the "remember passwords" setting, the forms table must be cleared because pulling again will fail (we remove the password stored in memory when opening the dialog for the second time).

Manual tests on  Linux, Mac and Windows 10:
- On the connection dialog (both pull and push tabs):
  - Write the settings and use a wrong password. Clicking on connect should fail
  - Fix the password and click on connect. Now it should work.
  - On master, it should continue failing
- Pull tab:
  - With "remember" passwords off
    - Open the connection dialog, configure correctly a connection to a server using username and password, click on connect. Some forms should show in the forms table.
    - Open the connection dialog again and click on cancel
    - The forms table should be empty
    - On master, you should continue seeing the forms table and pulling a form should fail.
  - Same thing with "remember" passwords on
    - The forms table should stay

This is the fastest solution, not the best. We need to redesign the business operations that interact with HTTP to be self-contained.

None I can think of

### DIFF
--- a/src/org/opendatakit/briefcase/ui/PullTransferPanel.java
+++ b/src/org/opendatakit/briefcase/ui/PullTransferPanel.java
@@ -53,6 +53,7 @@ import org.opendatakit.briefcase.model.TransferFailedEvent;
 import org.opendatakit.briefcase.model.TransferSucceededEvent;
 import org.opendatakit.briefcase.util.FileSystemUtils;
 import org.opendatakit.briefcase.util.TransferAction;
+import org.opendatakit.briefcase.util.WebUtils;
 
 /**
  * Pull forms and data to external locations.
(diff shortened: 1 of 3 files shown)


kkrawczyk123:
Checked on Ubuntu, Mac and Windows and it works exactly like in description on all.

kkrawczyk123:

ggalmazor:
@opendatakit-bot label "tested"

PR #367, opened 2018-02-13
Update CONTRIBUTING.md

yanokwa:
Matching Collect


yanokwa:
@kkrawczyk123 @ggalmazor I wanted to bring your attention to this change. When testing is finished, you can mark the issue as "behavior verified".

kkrawczyk123:
I would like to suggest changing color of this label because it looks like "reviewed" and it's not very distinguishable ;)

PR #368, opened 2018-02-13
Sentry integration

ggalmazor:
This PR integrates Briefcase with Sentry through its logback appender option.

Manually run Briefcase, both UI and CLI and produced some errors by exporting an encrypted form without providing a valid PEM file. The errors should get into Sentry's dashboard.

Sentry has other ways to send errors but this was the most transparent one (one which requires close to zero code changes).

Other options require using older logging frameworks like log4j or directly using Sentry's SDK, which would produce duplication on error logging.

None.


lognaturel:
This will be so helpful to have! I've verified that both explicitly logged errors and fatal crashes are sent up. 

Would it be worth including a release build variant that uses the Sentry logback config? It seems like the kind of thing that will be really easy to forget otherwise and then we'd miss out on useful logging.

yanokwa:
@lognaturel I think that the existing setup is fine for release. That is, I have a separate folder hierarchy where I generate the release builds and so as long as logback.xml isn't committed to the repo, I can configure that folder hierarchy once and the building for release should be bulletproof.

Thanks for your review! I'll fix the conflicts and get this merged.

ggalmazor:
Yes, let's do that. I've had to do the same with #354...

ggalmazor:
I accidentally remove the branch on the fork and now I have to create a new PR, sorry! #381

PR #381, opened 2018-02-16
Sentry integration

ggalmazor:
This PR integrates Briefcase with Sentry through its logback appender option.

Manually run Briefcase, both UI and CLI and produced some errors by exporting an encrypted form without providing a valid PEM file. The errors should get into Sentry's dashboard.

Sentry has other ways to send errors but this was the most transparent one (one which requires close to zero code changes).

Other options require using older logging frameworks like log4j or directly using Sentry's SDK, which would produce duplication on error logging.

None.


ggalmazor:

ggalmazor:

ggalmazor:
@opendatakit-bot unlabel "behavior verified"

PR #391, opened 2018-02-21
Issue 384 Sentry metadata

### LINKED ISSUE #384: Improve Sentry error reports with environment & git commit/branch info
The Sentry Logback appender can be configured with Logback's MDC capabilities

In order to be able to process error reports correctly, we need at least to know the Git hash of the code that produced the error. Other information like environment and OS would help a lot while figuring out what happened.

When errors are submitted to Sentry, they should be sent along with some metadata including:
- Git commit hash and, when available, git branch/tag name
- Environment (development, production)
- OS (win, linux, mac)
- JVM (version and update number)

This information should be set up on `Launcher`, where the Sentry - Logback bridge is currently initialized.

More information at:
- https://docs.sentry.io/clients/java/modules/logback/
- https://logback.qos.ch/manual/mdc.html
- https://docs.sentry.io/clients/java/config/#mdc-tags

ggalmazor:
Closes #384

This PR configures Sentry appender to tag events with runtime env and release information

Sentry configuration options are:

| Key | Description |
| --- | --- |
| `sentry.enabled` | Enables Sentry configuration (defaults to false) |
| `sentry.dsn` | Sentry client DSN |
| ~`sentry.env`~ | ~Environment (defaults to `dev`)~ |

All these keys can be overwritten with a `gradle.properties` file on the project's root directory.

Manually run Briefcase, produce a syntactic error on a form and try to export it to send events to Sentry.
Confirmed that events arrive with all the specified metadata on Sentry dashboard.

Sentry also allows to set metadata through MDC properties, but I haven't been able to get it working.

None.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -126,7 +126,6 @@ buildConfig {
     buildConfigField 'String', 'GOOGLE_TRACKING_ID', getValue("googleAnalytics.trackingId", "UA-91951913-3")
     buildConfigField 'Boolean', 'SENTRY_ENABLED', getValue("sentry.enabled", "false")
     buildConfigField 'String', 'SENTRY_DSN', getValue("sentry.dsn", "https://b6f023a5f17e44d9a46447f8827e2a41:1b1f33d68b1b4404b85cd627563d37f9@sentry.io/287258")
-    buildConfigField 'String', 'SENTRY_ENV', getValue("sentry.env", "dev")
 }
 
 mainClassName = 'org.opendatakit.briefcase.Launcher'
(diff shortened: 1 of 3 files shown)


yanokwa:
I agree with the decision to default to `sentry.enabled` to false. I propose we drop the `sentry.env` tag because I don't think we'll use it. That is, just having the version information is enough for us to know where the problems are.

ggalmazor:
> I propose we drop the sentry.env tag 

OK, we can always add it afterwards if we see we need it.

> tradeoffs of not using MDC

I only can think about the downside of using it: MDCs are global and Sentry's keys could produce conflicts with other keys we could need, for example, to write an export log file per form.

PR #392, opened 2018-02-21
WIP - Improve export efficiency for big datasets

ggalmazor:
Just the results of a technical spike to improve threading and IO when exporting lots of submissions.

Tested with 50k submissions (1 repeat each). Took less than 3 seconds to export everything on my machine.


yanokwa:
Couple of quick comments. Could you attach the data set you used to test this? Also, how much faster is master vs this branch on your machine?

ggalmazor:
You can use [this form here](https://drive.google.com/open?id=1gbCO_AyuSSxrxmWyAzhyvz9lO2OWPKSF]. Just unzip that into your Briefcase Storage dir and launch Briefcase. In order to Briefcase notice the new form, you'll need probably to do the trick of changing the storage dir somewhere and then back to where you unzipped the form.

I didn't benchmark it but it felt like 30s with the UI blocked vs 3s without blocking the UI.

yanokwa:
This PR is faster but it also produces different output. I'm using a real dataset and the sample logback.xml.

**this pr (105-g4f3c877)**
```
time java -jar build/libs/ODK\ Briefcase\ v1.9.0-105-g4f3c877.jar --form_id independent-monitoring-revised --storage_directory /Users/yanokwa/Downloads/ --export_directory export --export_filename g4f3c877.csv
```

real	0m24.664s
user	1m27.447s
sys	0m12.015s

g4f3c877-hh.csv (9.5MB, 47 cols)
g4f3c877.csv (3.9MB, 43 cols)


**master (108-g98971cf)**

```
time java -jar build/libs/ODK\ Briefcase\ v1.9.0-108-g98971cf.jar --form_id independent-monitoring-revised --storage_directory /Users/yanokwa/Downloads/ --export_directory export --export_filename g98971cf.csv
```

real	1m12.820s
user	1m7.136s
sys	0m8.385s

g98971cf_hh.csv (11.4MB, 46 cols)
g98971cf.csv (4.2MB, 35 cols)


The differences when I run it is that master has lots of `o.o.briefcase.util.ExportToCsv - repeating tag at 17 skipping hh` being printed, so my guess is that the PR isn't executing all the code. The difference in time might also be because of the difference in what the PR prints to the console.

ggalmazor:
> This PR is faster but it also produces different output

For the changes I've pushed so far, I'm happy when I see that the number of rows in both the main CSV and the repeats CSV are what's expected and that all rows have different unique UUIDs. Going further, old and new outputs have to be compared and verified before we can approve any change. 

I wanted to have some the feedback of running this PR on another machine and see if my results can be reproduced.

> The difference in time might also be because of the difference in what the PR prints to the console

During my work on this spike, I identified things that were contributing in varying degrees to the slowness/blocking issues. These are the main issues that this PR addresses:
- Contingency issues in a map that was accessed by all the threads to get OutputStreams and write new lines on them
- IO delays due to the fact that each line was written directly on the files in random order
- Contingency issues in another map that was used to check file hashes by all the threads
- Massive use of the EventBus to notify progress, leading to a block of the main thread by saturation of events

Intuitively, other things like parsing each form twice, writing too many lines on logs, or saving all the submissions in the same folder don't help, but there are the ones listed above contribute much more to Briefcase's performance. I've even tried disabling all logging and the difference wasn't appreciable.

> master has lots of `o.o.briefcase.util.ExportToCsv - repeating tag at 17 skipping hh`

I just removed those log lines because at that time they felt cryptic and of little use.

yanokwa:
The rows are the same, yes, but the columns are not. And the CSV that is being produced doesn't look well-formed. As long as that is the case, it's hard to compare.

**master**
```
"Jan 25, 2018, 12:55:00 AM","Jan 25, 2018, 12:43:26 AM","Jan 25, 2018, 12:46:50 AM",12345,"",collector,"",practice,intra,"Jan 25, 2018",BAN,ADA,SOM-XYZ,"",1,"",2,uuid:1b908599-3fcd-4323-a778-feb01176e63f/
```

**pr**
```
2018-01-25T08:55:00.852,Jan 25, 2018, 12:43:26 AM,Jan 25, 2018, 12:46:50 AM,12345,,collector,,practice,intra,Jan 25, 2018,BAN,ADA,SOM-XYZ,,1,,2,uuid:1b908599-3fcd-4323-a778-feb01176e63f/hh,1,,No,No,Bob,1,2,2,1,1,0,1,0,,uuid:1b908599-3fcd-4323-a778-feb01176e63f,intra-SOM-XYZ-2018-01-25,uuid:1b908599-3fcd-4323-a778-feb01176e63f
```

ggalmazor:
OK, you're right. Maybe I rushed a little bit when pushing the PR. Next iteration will be focused on contents

ggalmazor:
I think I just made a breakthrough and got to a (very inefficient) working implementation. I tested it with a simple form, a complex form and a nested repeat group form (see the new `ExportToCsvTest`, `ExportToCsv2Test` and `ExportToCsv3Test` classes).

I don't know if I'm going to be able to work more on this until Monday. It would be great if anyone could give it a spin (@yanokwa, @lognaturel). It would be great to test run the implementation with different forms.

ggalmazor:
Closing in favor of #457

PR #393, opened 2018-02-21
Fix #378 - Allow yyyy/MM/dd and yyyy-MM-dd for start and end dates.

### LINKED ISSUE #378: Match former expected date format in CLI and include it in help
#### Software versions 
Briefcase v1.9.0

#### Problem description
Dates used to be provided in yyyy/MM/dd format. Now they are expected in yyyy-MM-dd. Additionally, the format is not specified in the help text and the user sees an unexpected error when the format is wrong.

Run an export at the CLI with start and end date specified.

#### Expected behavior
The date format is backwards-compatible and the user is given a nice error message when the date is invalid.

mayank8318:
Closes #378 

Manually ran the following two commands - 
1) java -jar ODK\ Briefcase\ v1.9.0-102-g4d7e2ff-dirty.jar -e -id all-widgets --storage_directory /home/mayank/ --export_directory /home/mayank/ODKExport/ --export_filename abcd.csv -start 2018/01/20

2) java -jar ODK\ Briefcase\ v1.9.0-102-g4d7e2ff-dirty.jar -e -id all-widgets --storage_directory /home/mayank/ --export_directory /home/mayank/ODKExport/ --export_filename abcd.csv -start 2018-01-20

Positive results for both the tests.

This way we are supporting both the date formats "/" and "-".

None that I am aware of.

### DIFF
--- a/src/org/opendatakit/briefcase/operations/Export.java
+++ b/src/org/opendatakit/briefcase/operations/Export.java
@@ -52,8 +52,8 @@ public class Export {
   private static final Param<Void> EXPORT = Param.flag("e", "export", "Export a form");
   private static final Param<Path> EXPORT_DIR = Param.arg("ed", "export_directory", "Export directory", Paths::get);
   private static final Param<String> FILE = Param.arg("f", "export_filename", "Filename for export operation");
-  private static final Param<LocalDate> START = Param.arg("start", "export_start_date", "Export start date", LocalDate::parse);
-  private static final Param<LocalDate> END = Param.arg("end", "export_end_date", "Export end date", LocalDate::parse);
+  private static final Param<LocalDate> START = Param.localDate("start", "export_start_date", "Export start date");
+  private static final Param<LocalDate> END = Param.localDate("end", "export_end_date", "Export end date");
   private static final Param<Void> EXCLUDE_MEDIA = Param.flag("em", "exclude_media_export", "Exclude media in export");
   private static final Param<Void> OVERWRITE = Param.flag("oc", "overwrite_csv_export", "Overwrite files during export");
   private static final Param<Path> PEM_FILE = Param.arg("pf", "pem_file", "PEM file for form decryption", Paths::get);
--- a/test/java/org/opendatakit/common/cli/ParamTest.java
+++ b/test/java/org/opendatakit/common/cli/ParamTest.java
@@ -1,0 +1,29 @@
+package org.opendatakit.common.cli;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+import java.time.LocalDate;
+import org.junit.Test;
+import org.opendatakit.briefcase.reused.BriefcaseException;
+
+public class ParamTest {
+
+  @Test(expected = BriefcaseException.class)
+  public void throwBriefcaseExceptionForInvalidDateFormat() {
+      Param<LocalDate> localDate = Param.localDate("start", "export_start_date", "Export start date");
+      localDate.map("2018.01.20");
+  }
+
+  @Test
+  public void acceptsIso8601Dates () {
+      Param<LocalDate> localDate = Param.localDate("start", "export_start_date", "Export start date");
+      assertThat(localDate.map("2018-01-20"), is(LocalDate.of(2018, 1,20)));
+  }
+
+  @Test
+  public void normalizesInputSlashesToHyphens() {
+      Param<LocalDate> localDate = Param.localDate("start", "export_start_date", "Export start date");
+      assertThat(localDate.map("2018/01/20"), is(LocalDate.of(2018, 1, 20)));
+  }
+}
--- a/src/org/opendatakit/briefcase/ui/BriefcaseCLI.java
+++ b/src/org/opendatakit/briefcase/ui/BriefcaseCLI.java
@@ -76,8 +76,8 @@ public class BriefcaseCLI {
             fileName,
             exportMedia,
             overwrite,
-            Optional.ofNullable(startDateString).map(LocalDate::parse),
-            Optional.ofNullable(endDateString).map(LocalDate::parse),
+            Optional.ofNullable(startDateString).map(s -> LocalDate.parse(s.replaceAll("/", "-"))),
+            Optional.ofNullable(endDateString).map(s -> LocalDate.parse(s.replaceAll("/", "-"))),
             Optional.ofNullable(pemKeyFile).map(Paths::get)
         );
     } catch (BriefcaseException e) {
--- a/src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java
+++ b/src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java
@@ -67,7 +67,8 @@ public class MainBriefcaseWindow extends WindowAdapter implements UiStateChangeL
   final StorageLocation storageLocation;
 
   public static final String AGGREGATE_URL = "aggregate_url";
-  public static final String DATE_FORMAT = "yyyy/MM/dd";
+  public static final String DATE_FORMAT1 = "yyyy-MM-dd";
+  public static final String DATE_FORMAT2 = "yyyy/MM/dd";
   public static final String EXCLUDE_MEDIA_EXPORT = "exclude_media_export";
   public static final String EXPORT_DIRECTORY = "export_directory";
   public static final String EXPORT_END_DATE = "export_end_date";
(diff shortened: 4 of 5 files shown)


mayank8318:
@ggalmazor Now both the formats work in both the modes. I have added the help text in Legacy mode.
But wasn't sure where to put the check in the new mode. Should I put it in here :

operations.forEach(operation -> {
        if (cli.hasOption(operation.param.shortCode)) {
          checkForMissingParams(cli, operation.requiredParams);
          operation.argsConsumer.accept(Args.from(cli, operation.getAllParams()));
          executedOperations.add(operation);
        }
      });

mayank8318:
@ggalmazor Noticed one more thing - Since we are replacing all occurrences of "/" with "-" we are also allowing dates with mixed formats like 2018-01/20.

ggalmazor:
> Noticed one more thing - Since we are replacing all occurrences of "/" with "-" we are also allowing dates with mixed formats like 2018-01/20. 

I guess so. The general idea is that we will try to normalize input to an ISO8601... 

Users do weird stuff :)

src/org/opendatakit/briefcase/operations/Export.java
```diff
@@ -52,8 +52,8 @@
   private static final Param<Void> EXPORT = Param.flag("e", "export", "Export a form");
   private static final Param<Path> EXPORT_DIR = Param.arg("ed", "export_directory", "Export directory", Paths::get);
   private static final Param<String> FILE = Param.arg("f", "export_filename", "Filename for export operation");
-  private static final Param<LocalDate> START = Param.arg("start", "export_start_date", "Export start date", LocalDate::parse);
-  private static final Param<LocalDate> END = Param.arg("end", "export_end_date", "Export end date", LocalDate::parse);
+  private static final Param<LocalDate> START = Param.arg("start", "export_start_date", "Export start date", s -> LocalDate.parse(s.replaceAll("/", "-")));
```
ggalmazor: You can add the valid format in the description of both Params

src/org/opendatakit/common/cli/Cli.java
```diff
@@ -95,6 +97,12 @@ public void run(String[] args) {
       operations.forEach(operation -> {
         if (cli.hasOption(operation.param.shortCode)) {
           checkForMissingParams(cli, operation.requiredParams);
+          if (operation.param.shortCode.equals("e")) {
```
ggalmazor: `Cli` must not know anything about the operation it runs. These checks can't be done here. I think it's fine to let `LocalDate.parse()` throw a `DateTimeParseException`. We could catch that on the Param's mapper and transform it to a `BriefcaseException` that will get gracefully caught by `Cli` afterwards.

The only problematic thing is that the mapper is getting quite big. Maybe it's time we create a factory for `Param<LocalDate>` like this:

```java
class Param<T> {
public static Param<LocalDate> localDate(String shortCode, String longCode, String description) {
    return Param.arg(
        shortCode,
        longCode,
        description + " (yyyy-MM-dd or yyyy/MM/dd)",
        dateAsText -> {
          try {
            return LocalDate.parse(dateAsText.replaceAll("/", "-"));
          } catch (DateTimeParseException e) {
            throw new BriefcaseException("Invalid date format");
          }
        }
    );
  }
}
```

mayank8318: I will submit the changes by tonight :)

mayank8318:
@ggalmazor Where should I add the tests?

ggalmazor:
Same package as `Param` :)

ggalmazor:
> Not in the briefcase_test package? 

I think you're referring maybe to the test source set that IntelliJ separates for you. If that's it, yes, place the test on the test source set (I think it's on `test/java`). The test class itself should be on the same package as `Param` :)

mayank8318:
@ggalmazor Added 2 tests :)

test/java/org/opendatakit/common/cli/ParamTest.java
```diff
@@ -0,0 +1,28 @@
...
+  public void throw_briefcase_exception_for_invalid_date_format() {
+      Param<LocalDate> localDate = Param.localDate("start", "export_start_date", "Export start date");
+      localDate.map("2018.01.20");
+  }
+
+  @Test
+  public void parse_correctly () {
+      Param<LocalDate> localDate = Param.localDate("start", "export_start_date", "Export start date");
+      // FORMAT yyyy-MM-dd
+      assertEquals(localDate.map("2018-01-20").isEqual(LocalDate.of(2018, 01,20)), true);
```
ggalmazor: Make this an individual test method and make it follow the pattern `assertThat(value, is(expectedValue))`.

Name the method according to what's being asserted. Something like `acceptsIso8601Dates()`

test/java/org/opendatakit/common/cli/ParamTest.java
```diff
@@ -0,0 +1,28 @@
...
+      // FORMAT yyyy/MM/dd
+      assertEquals(localDate.map("2018/01/20").isEqual(LocalDate.of(2018, 01,20)), true);
```
ggalmazor: Make this an individual test method and make it follow the pattern `assertThat(value, is(expectedValue))`

Name the method according to the behavior that's being asserted. Something like `normalizesInputSlashesToHyphens()`

test/java/org/opendatakit/common/cli/ParamTest.java
```diff
@@ -0,0 +1,28 @@
...
+      // FORMAT MIXED '-' and '/'
+      assertEquals(localDate.map("2018-01/20").isEqual(LocalDate.of(2018, 01,20)), true);
```
ggalmazor: Drop this assertion. This behavior is accidental, meaning that we don't really want to support that particular format. Tests serve as documentation and we don't want to suggest that mixing separators is something we want our users to do.

test/java/org/opendatakit/common/cli/ParamTest.java
```diff
@@ -0,0 +1,28 @@
...
+  @Test(expected = BriefcaseException.class)
+  public void throw_briefcase_exception_for_invalid_date_format() {
+      Param<LocalDate> localDate = Param.localDate("start", "export_start_date", "Export start date");
+      localDate.map("2018.01.20");
+  }
+
+  @Test
+  public void parse_correctly () {
+      Param<LocalDate> localDate = Param.localDate("start", "export_start_date", "Export start date");
+      // FORMAT yyyy-MM-dd
```
ggalmazor: Remove all redundant comments that can be replaced by naming methods and variables. (see following comments)

test/java/org/opendatakit/common/cli/ParamTest.java
```diff
@@ -0,0 +1,30 @@
...
+  @Test(expected = BriefcaseException.class)
+  public void throwBriefcaseExceptionForInvalidDateFormat() {
+      Param<LocalDate> localDate = Param.localDate("start", "export_start_date", "Export start date");
+      localDate.map("2018.01.20");
+  }
+
+  @Test
+  public void acceptsIso8601Dates () {
+      Param<LocalDate> localDate = Param.localDate("start", "export_start_date", "Export start date");
+      assertThat(localDate.map("2018-01-20").isEqual(LocalDate.of(2018, 01,20)), is(true));
```
ggalmazor: These assertions are not leveraging Hamcrest matchers correctly. To see that, change the expected local date to produce a test failure and see the error message:

```xunit
java.lang.AssertionError: 
Expected: is <true>
     but: was <false>
Expected :is <true>
     
Actual   :<false>
```

Now compare it when you change the assertion to this:

```java
assertThat(localDate.map("2018-01-20"), is(LocalDate.of(2018, 2,20)));
```

```
java.lang.AssertionError: 
Expected: is <2018-02-20>
     but: was <2018-01-20>
Expected :is <2018-02-20>
     
Actual   :<2018-01-20>
```

That failure message is much more useful.

test/java/org/opendatakit/common/cli/ParamTest.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  @Test
+  public void normalizesInputSlashesToHyphens() {
+      Param<LocalDate> localDate = Param.localDate("start", "export_start_date", "Export start date");
+      assertThat(localDate.map("2018/01/20").isEqual(LocalDate.of(2018, 01,20)), is(true));
```
ggalmazor: Change this assertion too.

ggalmazor:

kkrawczyk123:
I am not able to test it, I guess that something is still wrong with checkstyle because I get error message when try to build. Screenshot attached:
screenshot_checkstyle

kkrawczyk123:

ggalmazor:
Sorry @kkrawczyk123, my bad for labelling it. I didn't notice that the build was failing.

ggalmazor:
@opendatakit-bot unlabel "reviewed"

mayank8318:
@ggalmazor Didn't change that part

ggalmazor:
> Didn't change that part

Sorry, we can't merge your PR if the build is failing!

I think that checkstyle is complaining because the new `Param.localDate` method separates the two `Param.arg` overloads. Placing `Param.localDate` below the last `Param.arg` method solves that.

There are checkstyle issues on `ParamTest` also... Review those and solve them as well, please.

mayank8318:
@ggalmazor Sorry didn't realise it meant that. Solved it now

ggalmazor:

kkrawczyk123:
### Tested with success!
**Verified on: Ubuntu 16, MacOS 10.11.6 and Windows 10**

Issue #352, opened 2018-02-06
Bulk exporting more than one encrypted form fails


ggalmazor:
#### Software versions 
Briefcase v1.9.9, Java v1.8.0_161, Linux (Debian Buster)

#### Problem description
When the user selects more than one encrypted forms for export, there's a filesystem concurrency problem and only one form is successfully exported.

#### Steps to reproduce the problem
Select more than one encrypted forms and try to export. **The selected forms have to be at least one encrypted submission.**

#### Expected behavior
All forms should be exported successfully (given correct confs)

#### Other information 
Stacktrace:
```
[ForkJoinPool-1-worker-4] ERROR org.opendatakit.briefcase.util.FileSystemUtils - Error decrypting:submission.xml.enc
java.io.FileNotFoundException: /tmp/.temp/submission.xml (No existe el archivo o el directorio)
	at java.io.FileOutputStream.open0(Native Method)
	at java.io.FileOutputStream.open(FileOutputStream.java:270)
	at java.io.FileOutputStream.<init>(FileOutputStream.java:213)
	at java.io.FileOutputStream.<init>(FileOutputStream.java:162)
	at org.opendatakit.briefcase.util.FileSystemUtils.decryptFile(FileSystemUtils.java:466)
	at org.opendatakit.briefcase.util.FileSystemUtils.decryptSubmissionFiles(FileSystemUtils.java:566)
	at org.opendatakit.briefcase.util.FileSystemUtils.decryptAndValidateSubmission(FileSystemUtils.java:754)
	at org.opendatakit.briefcase.util.ExportToCsv.processInstance(ExportToCsv.java:877)
	at org.opendatakit.briefcase.util.ExportToCsv.doAction(ExportToCsv.java:176)
	at org.opendatakit.briefcase.export.NewExportAction.export(NewExportAction.java:83)
	at org.opendatakit.briefcase.ui.export.ExportPanel.lambda$export$11(ExportPanel.java:161)
	at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1382)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.ForEachOps$ForEachTask.compute(ForEachOps.java:291)
	at java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:731)
[...]

PR #395, opened 2018-02-22
[WIP] Check validity of pem file

### LINKED ISSUE #351: Toggle the export button depending on form encryption and validness of PEM file location
If the user selects an encrypted form, don't enable the export button until the user sets a valid default PEM file location or overrides each selected form's export configuration with a valid PEM file location.

For a PEM file location to be valid it must pass the following tests:
- Be a file
- Be readable
- Be parseable by `PEMParser`
- Contain a private key

These tests should be checked eagerly after the user selects a PEM file location.

Issue tracking:
- [ ] toggle export button depending on PEM file location configuration
- [ ] toggle export button depending on validness of configured PEM file
- [ ] show an error dialog when the selected PEM file location doesn't contain a private key
- [ ] refactor all export operations to depend upon a `PrivateKey` instance instead of a `File` pointing to the PEM file location

icemc:
Closes #351 
Doesn't close #351 but fixes part of the issue. This PR is part of a general solution I'm working on to solve #351 
This has been tested manually. But this work is still in progress, so more tests will be done when its finished
There is nothing different I could think of and this solution is rather simple and straight forward.
Yes since it's not yet complete


icemc:
Heading to school will check the failing tests when I get back

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -266,6 +275,8 @@ public void ifPullBeforeOverridePresent(Consumer<PullBeforeOverrideOption> consu
       errors.add("Missing date range start definition");
     if (!isDateRangeValid())
       errors.add(INVALID_DATE_RANGE_MESSAGE);
+    if (!isPemFilePresent())
```
ggalmazor: This wouldn't be right as is: An `ExportConfiguration` is valid without a PEM file. Pem file is only required for encrypted forms. Maybe we can tell the `ExportConfiguration` if it's a configuration for an encrypted form or not and use that to compute its validness

icemc: Ok got you.

icemc:
Closed this for testing purpose.

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -49,26 +48,29 @@
   private static final String START_DATE = "startDate";
   private static final String END_DATE = "endDate";
   private static final String PULL_BEFORE = "pullBefore";
+  private static final String ENCRYPTED_FORM_CONFIG = "encryptedFormConfig";
```
ggalmazor: Name suggestion: `formNeedsPrivateKey`

icemc: Ok will change it that's a better name

icemc:
@ggalmazor  this PR will focus on checking the validity of the PEM file. Even though it adds extra checks
On the next PR I will remove the PEM file and introduce the PrivateKey Class
Next PR will enable and disable the export button depending on the the validity of the export configuration of the selected forms. What do you think?

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -154,7 +163,15 @@ public ExportConfiguration setPemFile(Path path) {
   }
 
   public boolean isPemFilePresent() {
-    return pemFile.isPresent();
+    if (pemFile.isPresent())  {
```
ggalmazor: I think this is going in the right direction, but all these checks should be happening on `setPemFile(Path path)`. The idea is that we don't ever want an `ExportConfiguration` instance to hold an invalid value. We will only set a value on `ExportConfiguration.pemFile` if it's a path to a file with a valid `PrivateKey`

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -154,7 +163,15 @@ public ExportConfiguration setPemFile(Path path) {
...
+      Path path = pemFile.get();
+      //Check if path is a file and is readable
+      if (Files.isRegularFile(path) && Files.isReadable(path)) {
```
ggalmazor: When you find yourself writing `if` blocks and using `Optional<T>` instances, odds are that you can do it better. In this case, you could use `Optional<T>.filter(Predicate<T>)` to encode those `if` blocks and then do a `map(path->true).orElse(false)`.

src/org/opendatakit/briefcase/ui/export/components/FormsTableViewModel.java
```diff
@@ -101,10 +102,18 @@ private JButton buildOverrideConfButton(FormStatus form) {
       );
       dialog.onRemove(() -> removeConfiguration(form));
       dialog.onOK(configuration -> {
-        if (configuration.isEmpty())
+        if (configuration.isEmpty() &&
+                (((BriefcaseFormDefinition) form.getFormDefinition()).isFileEncryptedForm() ||
```
ggalmazor: The configuration should already know if it belongs to a form that needs a `PrivateKey` or not. This shouldn't be needed here.

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -131,7 +139,8 @@ public ExportConfiguration copy() {
         startDate,
         endDate,
         pullBefore,
-        pullBeforeOverride
+        pullBeforeOverride,
+            formNeedsPrivateKey
```
ggalmazor: Watch the format ;)

src/org/opendatakit/briefcase/export/ExportConfiguration.java
ggalmazor: This is coming from Value Object design pattern. You can read about it here: https://martinfowler.com/bliki/ValueObject.html

src/org/opendatakit/briefcase/export/ExportForms.java
```diff
@@ -60,8 +61,14 @@ public static ExportForms load(ExportConfiguration defaultConfiguration, List<Fo
     forms.forEach(form -> {
       String formId = getFormId(form);
       ExportConfiguration load = ExportConfiguration.load(exportPreferences, buildCustomConfPrefix(formId));
-      if (!load.isEmpty())
+      if (((BriefcaseFormDefinition) form.getFormDefinition()).isFileEncryptedForm() ||
+              ((BriefcaseFormDefinition) form.getFormDefinition()).isFieldEncryptedForm()) {
+        //Check if this is an encrypted form. If true then set config to know it needs a PEM file
+        load.setFormNeedsPrivateKey(Boolean.TRUE);
```
ggalmazor: Think that the only reason to change this value would be if a conf changes its form and this doesn't make sense. This value needs to be set on creation and is never to be changed (make it `final`, by the way)

icemc: Yeah agree that this should be made final. Had a lot of trouble finding all the possible scenarios where users could change config

src/org/opendatakit/briefcase/export/ExportForms.java
```diff
@@ -60,8 +61,14 @@ public static ExportForms load(ExportConfiguration defaultConfiguration, List<Fo
     forms.forEach(form -> {
       String formId = getFormId(form);
       ExportConfiguration load = ExportConfiguration.load(exportPreferences, buildCustomConfPrefix(formId));
-      if (!load.isEmpty())
+      if (((BriefcaseFormDefinition) form.getFormDefinition()).isFileEncryptedForm() ||
```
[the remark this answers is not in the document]
icemc: Ok

src/org/opendatakit/briefcase/export/ExportConfiguration.java
icemc: WTF this didn't show up in my editor

src/org/opendatakit/briefcase/export/ExportForms.java
icemc: Thanks for your review will update the PR

icemc:
Hello @ggalmazor adding this new config variable was kinda tedious for me but i think I've gone through it. Now we can always know if any form needs a pem file  from its configuration and check if the pem file is valid before enabling the export button. I'm tempted to solve all of #351 in PR
So from here I may move all the pem file operations in PrivateKey class.
and next enable/disable the export button depending on the validity of the selected forms export configuration.

icemc:
@ggalmazor  Having some failing tests due to the changes I introduced. Would love to have a review from you and get your opinion on the changes.

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -77,18 +83,28 @@ public static ExportConfiguration load(BriefcasePreferences prefs) {
...
         prefs.nullSafeGet(END_DATE).map(LocalDate::parse),
         prefs.nullSafeGet(PULL_BEFORE).map(Boolean::valueOf),
-        prefs.nullSafeGet(PULL_BEFORE_OVERRIDE).map(PullBeforeOverrideOption::from)
+        prefs.nullSafeGet(PULL_BEFORE_OVERRIDE).map(PullBeforeOverrideOption::from),
+        prefs.nullSafeGet(FORM_NEEDS_PRIVATE_KEY).map(Boolean::valueOf)
     );
   }
 
-  public static ExportConfiguration load(BriefcasePreferences prefs, String keyPrefix) {
+  public static ExportConfiguration load(Optional<FormStatus> form, BriefcasePreferences prefs, String keyPrefix) {
```
ggalmazor: This is only used by `ExportForms.load` and we will always have a form to pass to `ExportConfiguration.load`. I think you can safely change the type of the `form` argument to `FormStatus` instead of `Optional<FormStatus>`.

Also, let's change the name to explicitly state that this factory is for loading override configuration for forms.

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -77,18 +83,28 @@ public static ExportConfiguration load(BriefcasePreferences prefs) {
         prefs.nullSafeGet(START_DATE).map(LocalDate::parse),
         prefs.nullSafeGet(END_DATE).map(LocalDate::parse),
         prefs.nullSafeGet(PULL_BEFORE).map(Boolean::valueOf),
-        prefs.nullSafeGet(PULL_BEFORE_OVERRIDE).map(PullBeforeOverrideOption::from)
+        prefs.nullSafeGet(PULL_BEFORE_OVERRIDE).map(PullBeforeOverrideOption::from),
+        prefs.nullSafeGet(FORM_NEEDS_PRIVATE_KEY).map(Boolean::valueOf)
```
ggalmazor: This is only used to load the default configuration. We need to default this to Optional.empty().

Also, let's change the name to explicitly state that this factory is meant only for loading the default configuration.

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -77,18 +83,28 @@ public static ExportConfiguration load(BriefcasePreferences prefs) {
...
         prefs.nullSafeGet(keyPrefix + EXPORT_DIR).map(Paths::get),
         prefs.nullSafeGet(keyPrefix + PEM_FILE).map(Paths::get),
         prefs.nullSafeGet(keyPrefix + START_DATE).map(LocalDate::parse),
         prefs.nullSafeGet(keyPrefix + END_DATE).map(LocalDate::parse),
         prefs.nullSafeGet(keyPrefix + PULL_BEFORE).map(Boolean::valueOf),
-        prefs.nullSafeGet(keyPrefix + PULL_BEFORE_OVERRIDE).map(PullBeforeOverrideOption::from)
+        prefs.nullSafeGet(keyPrefix + PULL_BEFORE_OVERRIDE).map(PullBeforeOverrideOption::from),
+        //Check if this forms FORM_NEEDS_PRIVATE_KEY pref is set. If not this is its first config
+        //Determine if it needs private key or not and set the pref depending on the form definition
+        oldpref.map(value -> Optional.of(value)).orElse(Optional.of(formNeedsPrivateKey))
```
ggalmazor: If you do the previous suggestions, this becomes `prefs.nullSafeGet(keyPrefix + FORM_NEEDS_PRIVATE_KEY).map(Boolean::valueOf)` (mirroring what's done in the other fields)

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -144,17 +163,41 @@ public ExportConfiguration setExportDir(Path path) {
...
     return pemFile;
   }
 
   public ExportConfiguration setPemFile(Path path) {
-    this.pemFile = Optional.ofNullable(path);
+    //Check if path is a file and is readable
+    if (Files.isRegularFile(path) && Files.isReadable(path)) {
+      //todo check if file is parsable by PEMParser and contains private key
+      this.pemFile = Optional.ofNullable(path);
+    }
```
ggalmazor: What if we don't want to set it? How does the UI know that this has happened?

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -290,11 +340,13 @@ public void ifPullBeforeOverridePresent(Consumer<PullBeforeOverrideOption> consu
...
     if (!isDateRangeValid())
       errors.add(INVALID_DATE_RANGE_MESSAGE);
+    if (formNeedsPrivateKey.orElse(false) && !isPemFilePresent())
+      errors.add(INVALID_PEM_FILE);
 
     return errors;
   }
 
-  public boolean isEmpty() {
+  public boolean isUserConfigsEmpty() {
```
ggalmazor: Why do we need this and the new `isEmpty()`. Is there any option to just have one definition of an empty conf? What would you need to change?

src/org/opendatakit/briefcase/export/ExportForms.java
```diff
@@ -140,7 +142,12 @@ public ExportConfiguration getConfiguration(String formId) {
   }
 
   public void removeConfiguration(FormStatus form) {
-    customConfigurations.remove(getFormId(form));
```
ggalmazor: Wow, this change makes this method a liar! We can't do this... You can change the implementation, but not the meaning or the intended behavior. Here we want to **remove** a configuration, not empty its contents.

src/org/opendatakit/briefcase/ui/export/components/FormsTableViewModel.java
```diff
@@ -128,7 +129,8 @@ private void updateDetailButton(FormStatus form, JButton button) {
   }
 
   private void updateConfButton(FormStatus form, JButton button) {
-    button.setForeground(forms.hasConfiguration(form) ? NO_CONF_OVERRIDE_COLOR : DARK_GRAY);
```
ggalmazor: This is getting outside of the ExportConfiguration stuff that we shouldn't have to care about here. I'd like to maintain the restriction of having to ask for `forms.hasConfiguration(form)`, not to see if `forms.getCustomConfiguration(form).map(f -> !f.isUserConfigsEmpty()).orElse(false)`.

This comment also affects the rest of changes that you did to support this:
- The different definitions of emptiness of an ExportConfiguration `isUserConfigsEmpty()` vs. `isEmpty()`
- The change on the behavior of the `removeConfiguration(FormStatus form)` method

test/java/org/opendatakit/briefcase/export/ExportConfigurationTest.java
```diff
@@ -87,7 +88,7 @@ public void the_factory_can_load_prefixed_keys_on_saved_preferences() {
     BriefcasePreferences prefs = new BriefcasePreferences(InMemoryPreferences.empty());
     prefs.putAll(validConfig.asMap("some_prefix"));
 
-    ExportConfiguration load = load(prefs, "some_prefix");
+    ExportConfiguration load = load(Optional.empty(), prefs, "some_prefix");
```
ggalmazor: This factory should receive a `FormStatus`, not an `Optional<FormStatus>`.

icemc:
Hello @ggalmazor seen you comments above. I think this PR is becoming more tedious. So I will suggest we write down a detailed process how we are gonna solve #351. 
Your Comments in the issue are clear but lack some specific details. I will confess I went into this issue without knowing much about the export code base(since I've done just little patches in this section in the past) but now I can say I understand whole export process and its code base. We can write down the details here or in the issue. Then from there I will provide new PRs to tackle #351. I suggest the following.
- Move all pem file functionalities in a new class Privatekey.
- Add check to verify if user selected pem file is valid before accepting it
- Determine if all selected forms have valid config 
- Disable/enable export depending on the validity of the form configs.

Each step above will be provided as a PR. Will love to to know your what you think. When we an agreed road map I will start providing the PRs. Sorry for the inconvenience. Cheers

ggalmazor:
Hi @icemc! Before starting to work on the roadmap, which I'm totally OK with (we can do whatever you want), I'd want to know what's particularly making you want to do this. I'm surprised because you were making progress and you're just following the same process I tend to follow while developing... 

Is it because I asked you to reconsider some changes? For me, it's very normal to explore one solution and change course on certain decisions... Some things work and some others don't. Let's hear your thoughts :)

Also, I was preparing a changeset built upon your PR to explain with concrete details what I was asking you to change.

Regarding the discussion about the roadmap, my only complaint is that each step should be focused on a behavior change: something we can test by running Briefcase and noticing that a button launches some dialog that before that didn't, etc. 

I assume that the list of four PRs you propose should be expanded and discussed but, a priori, from the four steps you propose, the only one with such visible changes in behavior would be the fourth one. 

I think we would assume a big risk by merging three PRs with no behavior change and waiting until the end to check if the expected behavior matches our expectations. You have to consider that it is possible to discover that the resulting behavior is not what we really need (even if we discuss it a lot beforehand, that can happen). Each PR merged should bring something valid by itself.

I've reviewed my original task list from #351 and I must admit that it's not perfect in that sense either. The first item is just some technical detail that it doesn't change visible behavior.

If we finally choose to review the roadmap, we have to consider that to create meaningful PRs for each task we define.

src/org/opendatakit/briefcase/export/ExportConfiguration.java
icemc: Ok got it

src/org/opendatakit/briefcase/export/ExportForms.java
icemc: @ggalmazor with this changes a form config is never empty and contains at least one field which is the formNeedsPrivateKey which is always either true or false and this cant be changed by the user. So technicaly this field can not be edited by the user while the others can be. This brings in the need to verify if the fields that can be edited by the user are empty (isUserConfigsEmpty()). Using this logic when a user clears a config removing the config totally will go against the logic. So I went ahead and instead cleared all the fields except the formNeedsPrivateKey field which will be required to display the export button as disabled/enabled. But the previous functionality of the clearall button before this change is not changed just the logic. I agree the removeConfig method should do what it refers to, but in this case I might change it to clearConfig since we never remove a config technically. We might also decide to remove configs for a form when the user clicks on the press all button. But this means the form config will depend on the default configuration upon export

src/org/opendatakit/briefcase/export/ExportConfiguration.java
icemc: Still working on this. This logic will be ported in the PrivateKey class which will contain a static method to check the validity of a pem file(path). This method will be used at the UI level to verify if the path provided is a valid pem if not an error message will be shown and the path will be rejected.

icemc:
Hello @ggalmazor thanks for the quick reply its really great having you around as a briefcase contributor. That said my main reason to try considering closing this issue is that I'm seeing my self working on other parts which don reflect the original intent of this PR. So my worry is I intended to resolve issue #351 with 3 PRs but as you mentioned above it won't be appropriate to merge PRs which don't improve briefcase behavior at the interface level. With this in mind I think I will go ahead and work on a fix of #351 directly in this PR. This will provide me with a clear goal which will enable me to focus and create a clean solution for the issue which we can the discuss and improve as time goes by before merging. Thanks very much I'm really learning as days go by. Cheers

icemc:
Since you reviewed you original task list of #351 I will be glad if you could share the new one here so I know what is required exactly.

src/org/opendatakit/briefcase/export/ExportConfiguration.java
ggalmazor: Good point @icemc! I'd suggest another name for the class because `PrivateKey` already exists and we're already using it on the export processes.

I think it's a great idea to use the new validation you're going to define as soon as possible. You could define a new filter for the FileChooser instance if you want (maybe you already thought of this).

If you do that, I'd change the member `ExportConfiguration.pemFile` for a `Optional<PrivateKey> privateKey` member. Otherwise, if you maintain the `pemFile` member, you'll have to validate it again in the setter. Every instance of `ExportConfiguration` must be always valid. 

If the UI already validates the pemFile location, it can extract the `PrivateKey` instance (maybe on the new class you're going to extract) and pass it to the `ExportConfiguration` guaranteeing that's always valid.

Also, remember to write unit tests for the new class that will handle cryptographic stuff.

src/org/opendatakit/briefcase/export/ExportForms.java
ggalmazor: In order to make the smallest possible change, I'd propose this:
- if the user removes the configuration, we just remove it. That's one thing.
- The other thing is how we decide that a conf is empty. To do that, we add a new condition following this criteria:
  - if it needs pem file and the pemfile is empty, then the conf is empty
  - if it needs pem file and the pemfile is not empty, then the conf is not empty
  - If it doesn't need pem file, then the conf is not empty

icemc:
@ggalmazor I think this new attempt is rather simple and does the job well. Please review cheers.

src/org/opendatakit/briefcase/model/BriefcasePreferences.java
```diff
@@ -51,7 +51,7 @@
   private static final String BRIEFCASE_UNIQUE_USER_ID_PROPERTY = "uniqueUserID";
 
   static {
-    // load the security provider
+    // loadFormConfig the security provider
     Security.addProvider(new BouncyCastleProvider());
```
icemc: This one was done by Intellij

icemc:
If the new changes are good to go then I think we are left with just three things to do
- Show error dialog when user selects an invalid pem file
- If some selected forms don't have valid config display a message to the user describing this problem.
- Correct the failing tests due to invalid pem files. 
@ggalmazor @lognaturel  waiting for your reviews

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -149,12 +161,25 @@ public ExportConfiguration setExportDir(Path path) {
   }
 
   public ExportConfiguration setPemFile(Path path) {
-    this.pemFile = Optional.ofNullable(path);
+    //Check if path is a valid pem file with privateKey
+    if (PrivateKeyUtils.isValidPrivateKey(path)) {
+      this.pemFile = Optional.ofNullable(path);
+    }
     return this;
```
ggalmazor: Did you consider to throw a `BriefcaseException`? How do third parties know that their call to `ExportConfiguration.setPemFile(Path path)` didn't have any effect on it?

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -327,14 +354,17 @@ public boolean isValidAsCustomConf() {
     return endDate.map(mapper);
   }
 
-  public ExportConfiguration fallingBackTo(ExportConfiguration defaultConfiguration) {
+  public ExportConfiguration fallingBackTo(ExportConfiguration defaultConfiguration, FormStatus form) {
```
ggalmazor: Remember that an `ExportConfiguration` object always knows if it needs a private key or not. Why not trust the value of the `defaultConfiguration` arg directly?

src/org/opendatakit/briefcase/export/ExportForms.java
```diff
@@ -133,10 +135,10 @@ public void updateDefaultConfiguration(ExportConfiguration configuration) {
     defaultConfiguration = configuration;
   }
 
-  public ExportConfiguration getConfiguration(String formId) {
-    return Optional.ofNullable(customConfigurations.get(formId))
+  public ExportConfiguration getConfiguration(FormStatus form) {
+    return Optional.ofNullable(customConfigurations.get(form.getFormDefinition().getFormId()))
```
ggalmazor: Duplicated code. It should be `getFormId(form)`. Also, if/when you change `ExportConfiguration.fallingBackTo()`, the signature change on `ExportForms.getConfiguration()` should be rolled back...

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -90,10 +91,12 @@ public ExportPanel(TerminationFuture terminationFuture, ExportForms forms, Expor
           preferences.putAll(configuration.asMap(buildCustomConfPrefix(formId)))
       );
 
-      if (forms.someSelected() && (form.getConfPanel().isValid() || forms.allSelectedFormsHaveConfiguration()))
+      if (forms.someSelected() && (form.getConfPanel().isValid() && forms.allSelectedFormsHaveConfiguration()))
```
ggalmazor: Good catch! Do you think it would be safe to remove the `form.getConfPanel().isValid()`? I haven't checked but now it feels like it's a redundant check because it's implicitly made in `forms.allSelectedFormsHaveConfiguration()`

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanel.java
```diff
@@ -40,7 +40,11 @@
       triggerOnChange();
     });
     form.onSelectPemFile(path -> {
-      configuration.setPemFile(path);
+      if (path == null) {
```
ggalmazor: The other callbacks don't have any conditionals... This null checks should be done inside setPemFile(path)

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -149,12 +161,25 @@ public ExportConfiguration setExportDir(Path path) {
...
+    //Check if path is a valid pem file with privateKey
+    if (PrivateKeyUtils.isValidPrivateKey(path)) {
+      this.pemFile = Optional.ofNullable(path);
+    }
     return this;
   }
 
-  public boolean isPemFilePresent() {
-    return pemFile.isPresent();
+  public ExportConfiguration removePemFile() {
```
ggalmazor: This is the only remove method. Other members rely on `Optional.ofNullable(value)` to sort that out. Please, mirror what we do on other members and remove this method unless it's really necessary.

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -149,12 +161,25 @@ public ExportConfiguration setExportDir(Path path) {
   }
 
   public ExportConfiguration setPemFile(Path path) {
-    this.pemFile = Optional.ofNullable(path);
+    //Check if path is a valid pem file with privateKey
+    if (PrivateKeyUtils.isValidPrivateKey(path)) {
```
ggalmazor: This is a wrong use of Optionals. You could write this instead: 

```java
public ExportConfiguration setPemFile(Path path) {
  this.pemFile = Optional.ofNullable(path).filter(PrivateKeyUtils::isValidPrivateKey)
}
```

If you do that, you don't need `removePemFile()` any more.

src/org/opendatakit/briefcase/ui/export/components/FormsTableViewModel.java
```diff
@@ -101,10 +101,11 @@ private JButton buildOverrideConfButton(FormStatus form) {
       );
       dialog.onRemove(() -> removeConfiguration(form));
       dialog.onOK(configuration -> {
-        if (configuration.isEmpty())
```
ggalmazor: Be careful with these unnecessary changes to the code base, please.

src/org/opendatakit/briefcase/ui/export/components/FormsTableViewModel.java
```diff
@@ -128,7 +129,7 @@ private void updateDetailButton(FormStatus form, JButton button) {
   }
 
   private void updateConfButton(FormStatus form, JButton button) {
-    button.setForeground(forms.hasConfiguration(form) ? NO_CONF_OVERRIDE_COLOR : DARK_GRAY);
+    button.setForeground(forms.hasConfiguration(form) ? NO_CONF_OVERRIDE_COLOR : DARK_GRAY );
```
ggalmazor: White space before last  `)`

ggalmazor:
@icemc 
> Since you reviewed you original task list of #351 I will be glad if you could share the new one here so I know what is required exactly.

I've updated #351 with this two lists:

New behavior:
- [ ] show an error dialog when the selected PEM file location is not valid or it doesn't contain a valid private key
- [ ] toggle export button only if every selected form which requires a private key has one configured (and the rest of current conditions are also met)
- [ ] give feedback to the user about the lack of required private keys.

Technical concerns:
- [ ] When an `ExportConfiguration` is bound to a form it should know on creation if it requires a private key or not and this should never change. 
- [ ] The default export configuration won't ever need a private key because it's not bound to any form.
- [ ] The `ExportConfiguration` should hold an `Optional<PrivateKey>` instead of an `Optional<Path>` to the PEM file location.
- [ ] Any validation of the private key should be done just after the user selects a PEM file location, not when the user clicks on the export button.

I think any PR should be focused on solving items on the new behavior list. The technical concern list is just to track technical decisions we have discussed.

In retrospect, I don't want to impose on you on how to work them. If you want to open one PR for everything, it's OK. I only ask you commit often to ease the review process. We can simplify the commit history afterwards if needed  (I think @lognaturel will agree with me on this).

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
icemc: Yeah that's true.

src/org/opendatakit/briefcase/export/ExportConfiguration.java
icemc: will remove this. Your next comment really shows its not  needed

src/org/opendatakit/briefcase/export/ExportConfiguration.java
icemc: wow will change it.

src/org/opendatakit/briefcase/export/ExportConfiguration.java
icemc: Yeah Exportconfiguration always know if a form needs a private key or not. But when we reach this case it means the form in question doesn't have any custom config and hence we fall back to the default export configuration to prepare a config for the form. This happens when the user clears all fields of the form custom config dialog, removing this config from memory and thus making the form not to have any custom config.
  But as you may have noticed we are just copying fields from the default configuration to create a config for this form and as you know the field formNeedsPrivatekey for default config is set to Optional.empty(). We still need to be able to know if this form needs a private key that's why a form status is passed to the fallingBackTo method. In case the form needs a private key and the default config doesn't have a pem file set, the config for this form will be invalid and the export button will be disabled.
Remember the clear button in the Custom configuration dialog causes the custom config to be removed from memory. Actually there is a hash map that keeps track of all custom configs and we check this hash map to get a custom config for a particular form if not present we fall back to the default configuration.

ggalmazor: OK, I understand your point. 

The thing is that before this, we were happy just by letting it fail: you could select a form that needs decryption and not set any PEM file location. Then Briefcase fails, showing the user the error.

The `fallingBackTo(ExportConfiguration conf)` method should not know anything about forms. It's just a method to merge two configurations. That's why I'm resisting to pass the form. A form should have to know how to fallback to another form without any other information.

If the problem is that, by removing confs, we're losing the information about the need for a private key, I'd suggest to change the `ExportForms.getConfiguration(FormStatus form)` method like this:

```java
public ExportConfiguration getConfiguration(FormStatus form) {
  return Optional.ofNullable(customConfigurations.get(form.getFormDefinition().getFormId()))
      .orElse(ExportConfiguration.empty(form))
      .fallingBackTo(defaultConfiguration);
}
```

I think that adding an `ExportConfiguration.empty(FormStatus form)` factory could be better because it lets us create an empty conf for a form, which we couldn't do before. Doing so, requires some other changes on other parts, like where the conf buttons are created and the Dialog factories, etc. I've prepared a branch if you want to check it out: https://github.com/ggalmazor/briefcase/commit/ad110ea3a47333767485ec9b9eb18a4197c6680c

Note that I've change the name of the arg in `ExportConfiguration.fallingBackTo()` to avoid confusion: instead of `defaultConfiguration`, now it's called `other`, implying no assumptions on the object.

icemc:
@ggalmazor sorry for the late reply. Replied to you reviews above. Please check them. Working on the changes

ggalmazor:
>  propose we add a simple label between the select all button and the export button that displays a short error message about some encrypted forms not having private key. We could still use a dialog but i don't really like dialogs and It's not really necessary here. Or better still look for an appropriate position for the label. What are your thoughts?

I was going to say that the label was a really good and simple idea, but I see you've gone ahead with the dialog. That works too! :)

ggalmazor:
By the way, I've reviewed your last commit and it's looking good. No suggestions so far. Thanks @icemc.

icemc:
Hello all. I'm preparing for my first semester exams. So I won't be active
for some time. Due to the current crisis here we are forced to write exams
now. Will update this PR as soon as I get some time. Cheers

icemc:
Will be done with exams on Monday. Will update the PR ASAP

ggalmazor:
> Will be done with exams on Monday. Will update the PR ASAP
Very good luck with your exams, @icemc!

icemc:
Thanks. Gsoc and exams don't schedule well this year for me.

icemc:
@ggalmazor I'm unable to extend the ExportPanelForm.form using Intelliji GUI (I'm getting a form file is invalid error message) designer class to include a label for the error message in case some forms dont have valid config. Could you please help me with this.

icemc:
Seems I have to resolve some conflicts. Working on it

icemc:
Hello @ggalmazor I've included a label to show some message why the export button is diabled. Also I;m trying to fix some failing tests, since Briefcase now checks if the path provided is a valid pem file some tests that were written now fail since they just passed but a simple file to the set pem file field. My attempt was to provide a sample content for the pem file but I get this error when running the test 

> ERROR org.opendatakit.briefcase.util.PrivateKeyUtils - Error while reading PEM file
org.bouncycastle.openssl.PEMException: problem parsing PRIVATE KEY: java.security.NoSuchProviderException: no such provider: BC

Here is the sample code
`try {
          List<String> lines = Arrays.asList(privateKey);
          VALID_PEM_FILE = Paths.get(createTempFile("private-key_", ".pem").toUri());
          Files.write(VALID_PEM_FILE, lines, Charset.forName("UTF-8"));
          Files.createDirectories(VALID_EXPORT_DIR);
    } catch (IOException e) {
      throw new RuntimeException(e);
    }`

ggalmazor:
Hi @icemc! It's not working because the file you're providing on the test doesn't contain a private key. When you use that file at `validConfig.setPemFile(VALID_PEM_FILE)`, the setter checks for the presence of a private key and if it doesn't find it, it assigns an `Optional.empty()`.

I'd suggest you create a real pem file and save it at `test/resources` (use a proper package for that). Then, in your test, you copy that file to a temp file and use that in the test.

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -92,10 +92,14 @@ public ExportPanel(TerminationFuture terminationFuture, ExportForms forms, Expor
           preferences.putAll(configuration.asMap(buildCustomConfPrefix(formId)))
       );
 
-      if (forms.someSelected() && (form.getConfPanel().isValid() && forms.allSelectedFormsHaveConfiguration()))
+      if (forms.someSelected() && (form.getConfPanel().isValid() && forms.allSelectedFormsHaveConfiguration())) {
+        form.setErrorMsg("No errors! Ready to export form(s)");
```
ggalmazor: Since this feedback won't change what the user is doing, I would remove it. Let's give feedback when we need the user to do something or if some process is really long.

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -92,10 +92,14 @@ public ExportPanel(TerminationFuture terminationFuture, ExportForms forms, Expor
...
         form.enableExport();
-      else {
-        //TODO show message to user why export button is disabled
+      } else {
+        if (!forms.someSelected())
+          form.setErrorMsg("No form(s) selected for export");
+        else
+          form.setErrorMsg("Some form(s) need encryption key. You can set this on the general settings above");
```
ggalmazor: Since the user can decide to set while overriding the general configuration, let's get rid of the last part and leave it "Some form(s) need a decryption key"

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.form
```diff
@@ -41,6 +41,15 @@
                   <text value="Clear All"/>
                 </properties>
               </component>
+              <component id="3e003" class="javax.swing.JLabel" binding="errorMsg">
+                <constraints/>
+                <properties>
```
ggalmazor: This new label should be not visible by default. If no errors are present, we should hide it.

icemc:
Hello @ggalmazor I think this one is good to go. Though some I'm left with some minor code cleanup in ExportAction. Please review when u have some time

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -174,8 +173,7 @@ public ExportConfiguration setExportDir(Path path) {
   }
 
   public ExportConfiguration setPemFile(Path path) {
-    //Check if path is a valid pem file with privateKey
-    this.pemFile = Optional.ofNullable(path).filter(PrivateKeyUtils::isValidPrivateKey);
+    this.pemFile = Optional.ofNullable(path);
```
ggalmazor: Are you sure about this? We talked about how we want this class to behave like a Value Object and how important is to validate any value we want to set... I think you shouldn't remove that validation from here. If you did this because you had trouble in other parts, you need to fix the other parts, not this one...

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.form
```diff
@@ -45,9 +45,11 @@
                 <constraints/>
                 <properties>
                   <autoscrolls value="true"/>
+                  <enabled value="false"/>
```
ggalmazor: This feels weird. Could you try to not disable it?

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -305,8 +303,6 @@ public void ifOverwriteExistingFilesPresent(Consumer<Boolean> consumer) {
       errors.add("Missing date range start definition");
     if (!isDateRangeValid())
       errors.add(INVALID_DATE_RANGE_MESSAGE);
-    if (formNeedsPrivateKey.orElse(false) && !isPemFilePresent())
```
ggalmazor: I think that removing the if is not right... This error list is used in two scenarios:
- On the default configuration: since it has an Optional.empty() on formNeedsPrivateKey, the if doesn't affect it.
- On the resolved configuration (with fallback to the default conf) of a form. Here we need to make sure that if the form needs decryption, we have set a pem file.

Maybe explain why you need to remove the check and we can discuss alternatives.

src/org/opendatakit/briefcase/export/ExportForms.java
```diff
@@ -164,7 +164,7 @@ public void putTransferSettings(FormStatus form, ServerConnectionInfo transferSe
   public boolean allSelectedFormsHaveConfiguration() {
     return getSelectedForms().stream()
         .map(this::getConfiguration)
-        .allMatch(ExportConfiguration::isValid);
+        .allMatch(ExportConfiguration::isValidAsCustomConf);
```
ggalmazor: As I've explained in my previous comment, the correct validation method would be isValid, because here we are checking resolved (with fallback to the default conf) configurations.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationDialog.java
```diff
@@ -35,7 +35,7 @@
       else
         form.disableClearAll();
 
-      if (this.confPanel.getConfiguration().isValidAsCustomConf())
+      if (true)
```
ggalmazor: This must be wrong... :P

test/java/org/opendatakit/briefcase/export/ExportFormsTest.java
```diff
@@ -124,8 +124,6 @@ public void knows_if_all_selected_forms_have_a_valid_configuration() {
     FormStatus form = forms.get(0);
     form.setSelected(true);
 
-    assertThat(forms.allSelectedFormsHaveConfiguration(), is(false));
```
ggalmazor: I think that we should get back this assertion as a consequence of previous comments

test/java/org/opendatakit/briefcase/model/PemFileTest.java
```diff
@@ -0,0 +1,27 @@
...
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+
+
+/**
+ * Created by Abanda on 3/30/2018.
+ */
+public class PemFileTest {
+
+    public static Path getSomePemFile() {
```
ggalmazor: We should take care of removing those files after the tests end...

ggalmazor:
@icemc sorry for the delay on my review! Last week I was on leave for some days :)

icemc:
Hello @ggalmazor welcome back. Made some changes as requested. Please have a look at the comment i placed in the setPemFile method. There is something weird going on when I check the if the path contains a PrivateKey. That check always fails (but passes on the UI code when the user selects the pem file). I've been battling with this for a while now and haven;t still understood why its failing on here. Also note that the path received by setPemFile is correct. Some help here will be appreciated

ggalmazor:
OK, so now we have a test harness to work on `PrivateKeyUtils.isValidPrivateKey(Path path)`. Now we're sure the method works when provided a Path to a PEM file. https://github.com/ggalmazor/briefcase/commit/ea16886bf1b8b26825f27b5fa8b97c342a8a1791

ggalmazor:
@icemc, I'm unable to reproduce an error in the UI with PEM files. I've recorded a GIF to show you what I did. If you write some instructions to reproduce it I can try again.

briefcase_pem_files

icemc:
Hello @ggalmazor I guess what I'm seeing is a build of my branch. On my
system exporting encrypted forms fail even though I have provided the
correct Pem file. That's because set pem file fails to set the pem file to
the configuration. Please could you provide me with the form you used to
test.

ggalmazor:
> Hello @ggalmazor do all the tests pass on your system. When u build my
branch?

I'm getting a failure `ExportConfigurationTest`. I think the PEM file is corrupted. Maybe you can change that to any of the private keys attached to #352

```
2018-04-09 10:24:41,771 [main] ERROR o.o.briefcase.util.PrivateKeyUtils - Error while reading PEM file
org.bouncycastle.openssl.PEMException: problem parsing PRIVATE KEY: java.security.NoSuchProviderException: no such provider: BC
	at org.bouncycastle.openssl.PEMReader$PrivateKeyParser.parseObject(Unknown Source)
	at org.bouncycastle.openssl.PEMReader.readObject(Unknown Source)
	at org.opendatakit.briefcase.util.PrivateKeyUtils.isValidPrivateKey(PrivateKeyUtils.java:25)
	at java.util.Optional.filter(Optional.java:178)
	at org.opendatakit.briefcase.export.ExportConfiguration.setPemFile(ExportConfiguration.java:184)
	at org.opendatakit.briefcase.export.ExportConfigurationTest.setUp(ExportConfigurationTest.java:69)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:24)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
Caused by: java.security.NoSuchProviderException: no such provider: BC
	at sun.security.jca.GetInstance.getService(GetInstance.java:83)
	at sun.security.jca.GetInstance.getInstance(GetInstance.java:206)
	at java.security.KeyFactory.getInstance(KeyFactory.java:211)
	... 36 common frames omitted
```

icemc:
I'm using java version "1.8.0_141". Do you think this can be the problem?

ggalmazor:
> I'm using java version "1.8.0_141". Do you think this can be the problem?

Yup! You need to update to 1.8.0_161 or newer

icemc:
I've updated my java and also changed the pem file to the ones you pointed
above and it doesn't still work. I've been digging into this for the past
5days noe and still don't understand the problem. Tagging @yanokwa here
maybe he could be of any help.

ggalmazor:
Hi @icemc! 

Just to understand in what point you are:
- Did you try the test I prepared which verifies `PrivateKeyUtils.isValidPrivateKey(Path path)`? Does it pass on your machine? 
- Did you also replace the PEM file at `test/resources/private-key.pem`?

icemc:
Yes I replaced the pem file with others. Haven't yet built the test you
prepared. Let me do so and give you feedback

icemc:
Hello sorry for not updating this issue for some days now. Was out of town
and didn't go with my PC.

icemc:
The build seems to pass on CircleCI

ggalmazor:
Don't worry, @icemc. I see that the CircleCI build works. It also works on my machine.

When you have some time, try to run the test I created on your machine and let's keep going until we discover why on your machine it doesn't work. Check your Java version too ;)

ggalmazor:
Hi, @icemc! I'm doing a little housekeeping today. Is it cool if I close this PR while we figure out how to proceed with this feature in the future?

The code has changed quite a bit and we would have to adapt your work before going on with this one.

icemc:
Ok no problem I'm fine with that
Le 30 août 2018 12:48 PM, "Guillermo Gutérrez" <notifications@github.com> a
écrit :

PR #403, opened 2018-02-27
[WIP] Provide option to overwrite on export in GUI

### LINKED ISSUE #202: Provide option to overwrite on export in GUI
As described in https://forum.opendatakit.org/t/duplicates-on-exporting-via-briefcase/9879

The option to overwrite an existing export is provided by the `-oc,–overwrite_csv_export` flag in CLI mode and should be surfaced in the UI.

@yanokwa @daveycrockett you will probably have ideas on where this should go. We should get agreement on that before it gets built!

emajj:
Closes #202


emajj:
Sorry will close this and create another. On the wrong branch.

PR #404, opened 2018-02-27
Add option to overwrite existing files

### LINKED ISSUE #202: Provide option to overwrite on export in GUI

emajj:
Closes #202 

This has been tested manually from the UI. 

This is the solution discussed and agreed on at #202 

Feature tested and working as expected.


src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -38,14 +38,8 @@
...
 import java.util.function.Consumer;
-import javax.swing.JButton;
-import javax.swing.JCheckBox;
-import javax.swing.JComboBox;
-import javax.swing.JComponent;
-import javax.swing.JLabel;
-import javax.swing.JPanel;
-import javax.swing.JTextField;
-import javax.swing.JTextPane;
+import javax.swing.*;
```
ggalmazor: I think this won't pass checkstyle rules...

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -71,12 +65,14 @@
   JComboBox<PullBeforeOverrideOption> pullBeforeOverrideField;
   JTextPane pullBeforeHintPanel;
   JLabel pullBeforeOverrideLabel;
+  private JCheckBox overwriteExistingFilesCheckBox;
```
ggalmazor: Make it final. In general, we want everything to be final (immutable) where possible.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -100,11 +96,11 @@ protected ConfigurationPanelForm(ConfigurationPanelMode mode) {
     layout.setConstraints(pullBeforeHintPanel, constraints);
 
     exportDirChooseButton.addActionListener(__ ->
-        buildExportDirDialog().choose().ifPresent(file -> setExportDir(Paths.get(file.toURI())))
+            buildExportDirDialog().choose().ifPresent(file -> setExportDir(Paths.get(file.toURI())))
```
ggalmazor: Make sure you have IntelliJ setup for indentation with 2 spaces.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -100,11 +96,11 @@ protected ConfigurationPanelForm(ConfigurationPanelMode mode) {
...
     );
     exportDirCleanButton.addActionListener(__ -> clearExportDir());
     pemFileChooseButton.addActionListener(__ ->
-        buildPemFileDialog().choose().ifPresent(file -> setPemFile(Paths.get(file.toURI())))
+            buildPemFileDialog().choose().ifPresent(file -> setPemFile(Paths.get(file.toURI())))
```
ggalmazor: Same here

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -248,17 +249,17 @@ private FileChooser buildPemFileDialog() {
 
   private FileChooser buildExportDirDialog() {
     return directory(
-        container,
-        fileFrom(exportDirField),
-        f -> f.exists() && f.isDirectory() && !isUnderBriefcaseFolder(f) && !isUnderODKFolder(f),
-        "Exclude Briefcase & ODK directories"
+            container,
```
ggalmazor: indentation this and following lines

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -270,6 +271,24 @@ private void triggerChangePullBeforeOverride() {
   }
 
 
+  private void triggerOverwriteExistingFiles() {
+    if (overwriteExistingFilesCheckBox.isSelected()) {
+      requestConfirmation();
+    }
+    onChangeOverwriteExistingFilesCallbacks.forEach(callback -> callback.accept(overwriteExistingFilesCheckBox.isSelected()));
```
ggalmazor: We should call the callbacks only if the user confirms that she wants to check the overwrite conf.

emajj: Ok

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
emajj: When I do that I get two compile errors. One on the declaration saying the field may not have been initialized and the second in the $$$setupUI$$$ method saying cannot assign to final field.

ggalmazor: You're right, sorry!

ggalmazor:
So, checkstyle is failing because of the imports ordering. These code style settings will solve that for you: 
[Briefcase.zip](https://github.com/opendatakit/briefcase/files/1780316/Briefcase.zip)

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -193,6 +201,15 @@ public ExportConfiguration setPullBeforeOverride(PullBeforeOverrideOption value)
     return this;
   }
 
+  public Optional<Boolean> getOverwriteExistingFiles() {
```
ggalmazor: This method is not used. Can we remove it?

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -119,6 +124,7 @@ protected ConfigurationPanelForm(ConfigurationPanelMode mode) {
     });
     pullBeforeField.addActionListener(__ -> triggerChangePullBefore());
     pullBeforeOverrideField.addActionListener(__ -> triggerChangePullBeforeOverride());
+    overwriteExistingFilesCheckBox.addActionListener(__ -> triggerOverwriteExistingFiles());
```
ggalmazor: The expected behavior of this checkbox is:
1. Ask for confirmation of this action by the user and 
2. Trigger the change if we get the confirmation

Reading this line, it's like we always trigger the change (no traces of the confirmation request here). If the confirmation request happens inside `triggerOverwriteExistingFiles()` method, then its name should be `requestConfirmationAndOverwriteExistingFilesIfTheUserConfirms()`, which is more accurate but painfully long and weirdly specific. This is a code-smell telling us that the `requestConfirmationAndOverwriteExistingFilesIfTheUserConfirms()` method does too much stuff and it should be simplified. The best function is the one that does just one thing, and this one does three things.

Instead of renaming the method I'd suggest to be more explicit and write this listener:
```java 
overwriteExistingFilesCheckBox.addActionListener(__ -> {
  if (requestOverwriteExistingFilesConfirmation())
    triggerOverwriteExistingFiles();
});
```

This way, the confirmation is explicitly expressed inside the action listener and it's easy to understand that we only trigger the change if the user confirms the action.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -270,6 +284,22 @@ private void triggerChangePullBeforeOverride() {
   }
 
 
+  private void triggerOverwriteExistingFiles() {
```
ggalmazor: Following my comment on the checkbox's action listener... This method's name doesn't accurately express what's happening inside. If we can't find a proper name (because it's too long or too complex) it's often a sign telling us that we need to split the method and separate concerns until we can do so.

This method, and others like this one, handle triggering changes. If you take a look to other `triggerXYZ()` methods, they don't have any conditions inside them. They only iterate their corresponding callbacks and call them. This is another sign telling us that `triggerOverwriteExistingFiles()` should not do anything more than calling callbacks.

emajj: Makes sense. Will fix it. Removing the condition to the constructor before triggering this method as you requested above makes more sense.

emajj:
@ggalmazor I noticed something strange with the UI. Now when I run the UI, the checkbox comes checked by default even though in the form designer I did not check Selected and when I run the code through a debugger and break in the constructor and evaluate `overwriteExistingFilesCheckBox.isSelected()` it returns false. Not sure which part of my code sets it to true right before it displays.

ggalmazor:
Hi @emajj. That's because in your previous run of Briefcase you selected the checkbox. Briefcase remembers the settings from one launch to the next.

I've just downloaded your PR and the first time the checkbox is not selected.

The part of the code that is responsible for loading the saved preferences is at `ConfigurationPanel#37`:

```java
// inside the constructor
configuration.ifOverwriteExistingFilesPresent(form::setOverwriteExistingFiles);
```

This happens **after** the `ConfigurationPanelForm` is built.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanel.java
```diff
@@ -34,6 +34,7 @@
     configuration.ifEndDatePresent(form::setEndDate);
     configuration.ifPullBeforePresent(form::setPullBefore);
     configuration.ifPullBeforeOverridePresent(form::setPullBeforeOverride);
+    configuration.ifOverwriteExistingFilesPresent(form::setOverwriteExistingFiles);
```
emajj: @ggalmazor I am getting some strange behavior when I include this line of code. The overwrite existing file check box comes checked by default. I'm not sure why but seems the initial configuration has overwrite existing files specified although I'm not sure how. I spent the whole of last night debugging and I tracked it down to this [emajj:MainBriefcaseWindow.java#L249](https://github.com/emajj/briefcase/blob/master/src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java#L249)
```java
exportPanel = ExportPanel.from(exportTerminationFuture, BriefcasePreferences.forClass(ExportPanel.class), appPreferences, BACKGROUND_EXECUTOR, analytics); 
```
The object returned by 
```java
BriefcasePreferences.forClass(ExportPanel.class)
```
Has overwriteExistingFiles. I placed a break point inside [emajj:BriefcasePreferences.java#L65-L68](https://github.com/emajj/briefcase/blob/master/src/org/opendatakit/briefcase/model/BriefcasePreferences.java#L65-L68)
```java
private BriefcasePreferences(Class<?> node, PreferenceScope scope) {
    this.node = node;
    this.preferences = scope.preferenceFactory(node);
  }
```
And when I evaluated preferences.get("overwriteExistingFiles", "false") after it has been initialized as seen in the constructor, it returns true instead of false.

As a result when ExportConfiguration.load(BriefcasePreferences) is called, overwriteExisting files is set to true by default.

[emajj:ExportConfiguration.java#L73-L82](https://github.com/emajj/briefcase/blob/master/src/org/opendatakit/briefcase/export/ExportConfiguration.java#L73-L82)
```java
public static ExportConfiguration load(BriefcasePreferences prefs) {
    return new ExportConfiguration(
        prefs.nullSafeGet(EXPORT_DIR).map(Paths::get),
        prefs.nullSafeGet(PEM_FILE).map(Paths::get),
        prefs.nullSafeGet(START_DATE).map(LocalDate::parse),
        prefs.nullSafeGet(END_DATE).map(LocalDate::parse),
        prefs.nullSafeGet(PULL_BEFORE).map(Boolean::valueOf),
        prefs.nullSafeGet(PULL_BEFORE_OVERRIDE).map(PullBeforeOverrideOption::from),
        prefs.nullSafeGet(OVERWRITE_EXISTING_FILES).map(Boolean::valueOf)
    );
  }
```
Not sure what is causing that. Using the debugger I have been able to place breakpoints in the code to get to the root cause but when I get here [emajj:BriefcasePreferences.java#L76](https://github.com/emajj/briefcase/blob/master/src/org/opendatakit/briefcase/model/BriefcasePreferences.java#L76)

I can't seem to figure out what is happening. Is there a configuration file somewhere that it's reading?

emajj:
@ggalmazor I suspected that but when I ran and unchecked the box and close the application while the box is unchecked, and run again. It comes checked by default.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanel.java
ggalmazor: I've been able to reproduce this problem... I'll see what I can find :P

ggalmazor: OK. I think the code is correct and it does what we need.

I've found that, like me, maybe you were changing the value of the checkbox in a configuration that was not valid. For the default conf to be valid, it has to have the export dir filled in. Invalid confs are not saved as user preferences and then, even though you are changing the value, it doesn't effectively change what was saved from the previous launch.

I've also reviewed the code on the listener and `requestOverwriteExistingFilesConfirmation()`. I think it could be more clear to have something like this:

```java
overwriteFilesField.addActionListener(__ -> {
  if (overwriteFilesField.isSelected() || (overwriteFilesField.isSelected() && confirmOverwriteFiles()))
    triggerOverwriteFiles();
});

// and then...

private boolean confirmOverwriteFiles() {
  if (showConfirmDialog(this, "Overwrite existing files?", "", YES_NO_OPTION) == YES_OPTION)
    return true;
  overwriteFilesField.setSelected(false);
  return false;
}
```

The changes here are:
- Simplified shorter names (not so important, but makes reading the code easier. Also, changed `xyzCheckbox to xyzField to mirror other field members)
- The decision to trigger is entirely made on the listener
- The `confirmOverwriteFiles()` method will always launch a confirm dialog. Before this, it was launching it conditionally, which can be unexpected.

emajj: Thanks for looking into this. One question

Isn't  `overwriteFilesField.isSelected() || ((overwriteFilesField.isSelected() && confirmOverwriteFiles()) ` effectively the same as `overwriteFilesField.isSelected()`

I mean because of short circuiting, if overwriteFilesField.isSelected() is true, the other expressions will not be executed making the entire expression true. And if it's false the and in the second expression will make the second expression false making the entire expression false. Please correct me if I'm wrong.

ggalmazor: Sorry! I was missing a bang before the first member on the condition. I meant this:

```java
overwriteFilesField.addActionListener(__ -> {
  if (!overwriteFilesField.isSelected() || (overwriteFilesField.isSelected() && confirmOverwriteFiles()))
    triggerOverwriteFiles();
});
```

emajj: Ah I c. Thanks.

emajj: >The confirmOverwriteFiles() method will always launch a confirm dialog. Before this, it was launching it conditionally, which can be unexpected.

Looking at the code above. Not sure that will happen. If the action is unchecking the box, the the confirm dialog will not be shown. Which is what I was doing before.

What I was doing before is, if action is box checking, request confirmation. If confirmation is approval, call triggerOverwriteFiles(). But if action is unchecking, don't show dialog. Do you mean we should show the dialog even when the user is unchecking the box? But the code above will not achieve that because if user is unchecking the box,` overwriteFilesField.isSelected()` will return *false* and the *!* will make it *true* and due to to short circuit the other expression that contains the call to `confirmOverwriteFiles()` will not be called hence no confirmation box will be shown.

But the above your code looks shorter and cleaner. I just want to be sure I'm getting the expected behavior right.

ggalmazor: My idea was:
- If the user unchecks the checkbox, the confirm dialog gets shortcircuited (we don't show it), and we trigger the change
- If the user checks the checkbox, we show the confirm dialog
  - If the user confirms the action, trigger the change
  - If the user doesn't confirm the action, uncheck the checkbox and don't trigger the change

Now that I've reviewed it to write this text I realize that you're right. The condition could be:

```java
overwriteFilesField.addActionListener(__ -> {
  if (!overwriteFilesField.isSelected() || confirmOverwriteFiles())
    triggerOverwriteFiles();
});
```

A secondary goal for me was to make sure each time we execute `confirmOverwriteFiles()` a confirm dialog is shown (remove its conditional behavior).

emajj: > secondary goal for me was to make sure each time we execute confirmOverwriteFiles() a confirm dialog is shown (remove its conditional behavior).

Now I get it. Yeah looking at my code again, it's a little misleading as it ask for confirmation based 
on a condition. Thanks for the help :)

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.form
```diff
@@ -199,6 +199,15 @@
           </component>
         </children>
       </grid>
+      <component id="bd9c9" class="javax.swing.JCheckBox" binding="overwriteExistingFilesCheckBox" default-binding="true">
```
ggalmazor: The `binding` should be `overwriteFilesField`

src/org/opendatakit/briefcase/export/ExportAction.java
```diff
@@ -72,13 +72,14 @@ public static void export(BriefcaseFormDefinition formDefinition, ExportConfigur
...
     }
+    boolean overwriteExistingFiles = configuration.getOverwriteExistingFiles().isPresent() ? configuration.getOverwriteExistingFiles().get() : false;
     ExportToCsv action = new ExportToCsv(
         terminationFuture,
         configuration.getExportDir().orElseThrow(() -> new RuntimeException("Export dir not present")).toFile(),
         formDefinition,
         formDefinition.getFormName(),
         true,
-        false,
+        overwriteExistingFiles,
```
ggalmazor: This line should be `configuration.getOverwriteExistingFiles().orElse(false),`

Remember that if you are handling an `Optional<T>` and using `if`, it's very probable that you're not using it right.

ggalmazor:
@emajj after fixing the two things from my last review, I think you should squash all the commits into two:
- One commit for the changes in `ExportAction` and `ExportConfiguration`, telling that you have added a new value to the configuration
- One commit for the rest of changes, telling that you have wired the new configuration value in the UI

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.form
emajj: Thanks. Will fix that. What is the use of that binding? Because it seems to work even though the binding is wrong.

ggalmazor: The .form files are just for the UI Designer. Since the UI Designer generates part of the code of the  corresponding .java file, we need to keep them in sync.

A trick to do that is to change name of members in the UI Designer and then, it takes care of everything :)

ggalmazor:
OK, I think this is ready! :)

Can you rebase your commits into the two I mentioned in my last comment?

emajj:
> @emajj after fixing the two things from my last review, I think you should squash all the commits into two:
One commit for the changes in ExportAction and ExportConfiguration, telling that you have added a new value to the configuration
One commit for the rest of changes, telling that you have wired the new configuration value in the UI

@ggalmazor I will need a little help with this. I've only squashed multiple commits into one before using interactive rebase. But this seems I'd have to do some sort of commit selecting and group.

emajj:
@ggalmazor Ok what I want to try now is, squash all the commits into one then uncommit this commit so that my changes will show as though I've not committed then commit the files now putting appropriate changes in their own commit. Does this sound like a good approach?

emajj:
@ggalmazor done :smile:

ggalmazor:
Awesome! 

Your solution sound much more elegant :) I'll try that next time!

In this situation, I would normally rename the branch, create a new one and then cherry pick the changes for each commit.

ggalmazor:

yanokwa:
I don't think these are blockers to testing the functionality (which worked great for me), but we'll have to make some adjustments to the spacing. @ggalmazor might be best positioned since he can run on all platforms.

For these two, I'd add more space so they look like the are on separate lines
screen shot 2018-03-07 at 14 22 41
screen shot 2018-03-07 at 14 22 56

For the dialog, I'd remove the icon
screen shot 2018-03-07 at 14 22 47

ggalmazor:
@emajj, @yanokwa,

I can review your work on the three platforms and produce a branch with the required fixes for the UI. You can merge that to complete this PR after that.

Regarding the icon on the confirm dialog, I'm assuming that @emajj hasn't modified its previous implementation on `JOptionPane` but we can check if this is configurable. @emajj, could you check if there's a configuration to disable that icon, please?

If there's no pre-existing way to do this, I'd file this change as a separate issue, as `JOptionPane` is used everywhere on the app for other things as well.

ggalmazor:
I've added a vertical spacer to get what we need. I've tested it on Linux and Mac. I'll test it on Win this evening (I had a problem with the VM). Here: https://github.com/opendatakit/briefcase/commit/2b69d95bc0c99dfaa5de368beace1e704d859f63

kkrawczyk123:
Tested on all three systems. The functionality works well with single and multiple exports both. I tested cases with export of submissions from different date scope, removing something from .csv files manually or cleaning file - I got overwrite every time. In a case with no changes, I did not see any duplicates which I think is great. 
I agree with @yanokwa and I don't like icon on confirmation window too which by the way looks differently on all three systems. Screenshots attached. 
overwrite_linux
overwrite_mac
overwrite_windows
Maybe after removing icon it would be good to center the text in the dialog window.

I have also noticed that this dialog doesn't have any title - all of other dialog windows like choosing storage locations, connection to the server, detail status windows, override export configuration and even error messages do have titles so maybe it should be unified and also added. What do you think @lognaturel @yanokwa @ggalmazor? 

Issue #366, opened 2018-02-12
Unexpected behavior related to the connection settings dialog


ggalmazor:
#### Software versions 
Briefcase v1.9.0

#### Problem description
Given that a user doesn't select the "remember passwords" setting, when opening the connection settings dialog, then the current password from the `ServerConnectionInfo` object that's saved in memory on each `PullTransferPanel` and `PushTransferPanel` is removed.

If the user just cancels the dialog after that, the next pull would fail (we have lost the password). To prevent this problem, in #365 we changed it so that **the forms table gets cleared, which is very confusing**.

#### Steps to reproduce the problem
- Ensure that "remember passwords" setting is not selected
- Pull some forms from an Aggregate server
- Open the connection dialog and click on cancel
- The forms table gets cleared

#### Proposed solution
The problem is twofold:
- The UI component for form source/target configuration is kind of confusing because it doesn't have a UX flow that clearly separates a "configured" state from a "not configured" state.
- The UI code and the business logic are quite coupled between them and other parts of Briefcase.

I think that solving just one thing is not going to cut it. That's why I'd like to propose to work on a new UI component that will replace current form source/target selection:
- It has to visually separate "configured" and "not configured" states
- The UI code has to be self-contained. It has to be easily reusable and testable in isolation.
- The relied upon business logic has to be side-effect free and has to make it easy to know the validness of a given form source/target configuration.

I propose to work on this without touching the existing code, to lower the risk of accidentally affecting other coupled parts of Briefcase. After completing this, we should remove any unused part of the "old" code.

After conversations with @yanokwa and @lognaturel, they also propose we get rid of the mounted SD card option. It's something that Briefcase doesn't really need, now that modern Android devices let you mount their filesystem locally. The custom directory option is enough for that.

The following UX flow represents these ideas (ignore any SD Card reference):
![briefcase_mockup_1](https://user-images.githubusercontent.com/205913/36172883-8c24dd10-1107-11e8-9c20-e752bf305a99.jpg)
![briefcase_mockup_2](https://user-images.githubusercontent.com/205913/36172885-8c51d3d8-1107-11e8-8d00-7f8cf918176f.jpg)
![briefcase_mockup_4](https://user-images.githubusercontent.com/205913/36172887-8c8ab536-1107-11e8-96eb-510028f5fdef.jpg)

### Task checklist

These are "broad stroke" tasks. It's more like an implementation script to try to follow an incremental approach to the solution. We build progressively from small to large.

[...]

PR #405, opened 2018-02-28
WIP #366 - Create a UI form for the Aggregate server connection param…

mayank8318:
…s dialog.

WIP #366 
Created a UI form for aggregate server connection parameters.

-
-
-


mayank8318:
@ggalmazor Am I in the right direction here?

ggalmazor:
Great progress! I know that using the UI Designer is a challenge in itself :)

I see some layout issues. If you compare it with the dialog form you can notice some differences:

dialogs comparison

- Separators around the content
- The text fields should have weight x to 1
- There's some gap to the left of the cancel button

There are also some issues with consistency/naming:
- The UI view you're designing should be called `AggregateServerConnectionDialogForm`. We need to reserve `AggregateServerConnectionDialog` for the UI component (like with `ConfigurationDialog` and `ConfigurationDialogForm`)
- You should follow the same convention as in other UI views. For example, instead of `txtUsername` it should be `usernameField`, etc.

mayank8318:
@ggalmazor One doubt. Is that some special view or just the dark theme?

ggalmazor:
> One doubt. Is that some special view or just the dark theme?

It's the standard Darcula :)

mayank8318:
Woah! It is perfect for the GUI editor

mayank8318:
Sorry, I mixed the two components together. I will split it up into `form` and `dialog`

mayank8318:
@ggalmazor Made the changes but I don't see any UI code generated by IntelliJ?

ggalmazor:
Yo need tomanually build to see changes on your code. Also, configure the UI designer to generate source code (search for it on the settings page). @emajj had the same problem and maybe can help. I’m going to be away from a computer for several hours...

mayank8318:
@ggalmazor I have made the changes and added a static main method to see the dialog. Please review when you are free :)

ggalmazor:
Sorry, @mayank8318, I should have explained myself a little better...

When I said to look to `ConfigurationDialog` I didn't want you to create separate views for `AggregateServerConnectionPanelForm` and `AggregateServerConnectionDialogForm`. We don't need a separate view `AggregateServerConnectionPanelForm` because it's not going to be reused anywhere else.

In the case of the `ConfigurationDialogForm`, we have a separated `ConfigurationDialogForm` because it's used directly on the `ExportPanel` as well (it's reused).

What I wanted to explain is that the classes with a `Form` prefix on their names represent a component's  view object and they need a class to be their controllers. In the case of the `AggregateServerConnectionPanelForm`, I'm still missing a controller `AggregateServerConnectionPanel` class. The equivalent classes you can check out to mirror them are these:

| Function | Class | Equivalent |
| -- | -- | -- |
| Controller | `AggregateServerConnectionPanel` | `ConfigurationPanel` |
| View | `AggregateServerConnectionPanelForm` | `ConfigurationPanelForm` |

You can study these classes and see how they collaborate to mirror it with your Aggregate Server Connection classes.

**Also**

You can rebase master in your branch to make the review easier and avoid merging conflicts :)

mayank8318:
@ggalmazor Please see if this structure is valid

ggalmazor:
Yep, this is it :)

mayank8318:
@ggalmazor Now the button clicks should be handled in the view class right?

ggalmazor:
Yes. Any UI event should be handled by the view

mayank8318:
@ggalmazor I know `asTarget` and other variables belong to the Controller but how should I do it?

mayank8318:
@ggalmazor Is this the right way of doing it?

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialog.java
```diff
@@ -0,0 +1,68 @@
...
+  private boolean asTarget;
+  private boolean isSuccessful;
+  private ServerConnectionInfo serverInfo = null;
+  private TerminationFuture terminationFuture = new TerminationFuture();
+
+  AggregateServerConnectionDialog(AggregateServerConnectionDialogForm aggregateServerConnectionDialogForm, boolean asTarget) {
+    form = aggregateServerConnectionDialogForm;
+    this.asTarget = asTarget;
+
+    form.addConnectButtonCallback(e -> verifyParams());
```
ggalmazor: The convention we're following is `onConnect( <some callback here> )`

ggalmazor: Also, the callback should receive the user's input from the form in some structured object. Therefor, the callback should be of type `Consumer<ServerConnectionParams>` or something like that.

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialog.java
```diff
@@ -0,0 +1,68 @@
+package org.opendatakit.briefcase.ui.reused;
+
+import java.awt.Cursor;
```
ggalmazor: Ideally, this class wouldn't have any imports from Swing. Remember: the view knows about the UI. The controller knows about business

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialog.java
```diff
@@ -0,0 +1,68 @@
...
+    final ServerConnectionInfo info = new ServerConnectionInfo((form.getURLFieldText()).trim(),
+        form.getUsernameFieldText(), form.getPasswordFieldPassword());
+
+    form.disableConnectButton();
+    form.disableCancelButton();
+    String errorString;
+
+    Cursor saved = form.getCursor();
+    try {
+      form.setCursor(Cursor.getPredefinedCursor(Cursor.WAIT_CURSOR));
```
ggalmazor: This is the view's concern. Publish some method on the view to let the controller call it to change the cursor type

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialog.java
```diff
@@ -0,0 +1,68 @@
...
+      form.paint(form.getGraphics());
+      terminationFuture.reset();
+      ServerConnectionTest backgroundAction = new ServerConnectionTest(info, terminationFuture, asTarget);
```
ggalmazor: We need to develop our own ServerConnectionTest service. You can start by duplicating it. The reason behind it is that we want to get rid of the side-effects of the original implementation. While you do that, get rid of the `TerminationFuture` and all the code that uses it.

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialog.java
```diff
@@ -0,0 +1,68 @@
...
+      backgroundAction.run();
+      isSuccessful = backgroundAction.isSuccessful();
+      errorString = backgroundAction.getErrorReason();
+    } finally {
+      form.setCursor(saved);
+    }
+
+    if (isSuccessful) {
+      serverInfo = info;
+      form.setVisible(false);
```
ggalmazor: This should be hidden inside the form. You can create a `form.hide()` method to do so.

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialog.java
```diff
@@ -0,0 +1,68 @@
...
+    } else {
+      String errorMessage = errorString.trim();
+
+      if (!(errorMessage.endsWith("!") || errorMessage.endsWith("."))) {
+        errorMessage += ". ";
+      }
+
+      ODKOptionPane.showErrorDialog(form,
+          errorMessage + MessageStrings.PROXY_SET_ADVICE, "Invalid Server URL");
+      form.enableConnectButton();
```
ggalmazor: When did we disable them?

mayank8318: Start of the method body

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialog.java
mayank8318: `TerminationFuture` is used in a lot of places and most of the util ops are based on it. Should I just delete it?

src/org/opendatakit/briefcase/ui/export/AggregateServerConnectionConfiguration.java
```diff
@@ -0,0 +1,58 @@
+package org.opendatakit.briefcase.ui.export;
```
ggalmazor: Wrong package... Should be `org.opendatakit.briefcase.ui.reused`

src/org/opendatakit/briefcase/ui/export/AggregateServerConnectionConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+import java.net.URL;
+import java.util.Arrays;
+import java.util.Objects;
+
+public class AggregateServerConnectionConfiguration {
+  private URL url;
+  private String username;
+  private char[] password;
+
+  public AggregateServerConnectionConfiguration(String url, String username, char[] password) throws MalformedURLException {
```
ggalmazor: Let's use a `String` for the password. It's simpler than the `char[]`

src/org/opendatakit/briefcase/ui/export/AggregateServerConnectionConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+    this.url = new URL(url);
+    this.username = username;
+    this.password = password;
+  }
+
+  public URL getUrl() {
+    return url;
+  }
+
+  public void setUrl(String url) throws MalformedURLException {
```
ggalmazor: Don't throw a checked exception. Wrap the `new URL()` inside a try-catch and wrap the `MalformedURLException` inside a `BriefcaseException`

You'll need to change `BriefcaseException` to this: 
```java
public class BriefcaseException extends RuntimeException {
  public BriefcaseException(String message) {
    super(message);
  }

  public BriefcaseException(String message, Throwable cause) {
    super(message, cause);
  }
}
```

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialog.java
```diff
@@ -14,25 +15,25 @@
   private boolean isSuccessful;
   private ServerConnectionInfo serverInfo = null;
   private TerminationFuture terminationFuture = new TerminationFuture();
+  private AggregateServerConnectionConfiguration aggregateServerConnectionConfiguration;
 
-  AggregateServerConnectionDialog(AggregateServerConnectionDialogForm aggregateServerConnectionDialogForm, boolean asTarget) {
+  AggregateServerConnectionDialog(AggregateServerConnectionDialogForm aggregateServerConnectionDialogForm, boolean asTarget, AggregateServerConnectionConfiguration aggregateServerConnectionConfiguration) {
```
ggalmazor: Let's name those long args simply `form` and `configuration` for brevity's sake

src/org/opendatakit/briefcase/ui/export/AggregateServerConnectionConfiguration.java
mayank8318: :+1:

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialog.java
mayank8318: Ok

mayank8318:
@ggalmazor I will rectify the checkstyle later, is the right approach?

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Objects;
+
+public class AggregateServerConnectionConfiguration {
+  private URL url;
+  private String username;
+  private String password;
+
+  public AggregateServerConnectionConfiguration(String url, String username, String password) {
```
ggalmazor: The constructor should receive a `URL` instance. If you need to make transformations to create an object, we encode that as a static factory method

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+
+import org.opendatakit.briefcase.reused.BriefcaseException;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Objects;
+
+public class AggregateServerConnectionConfiguration {
+  private URL url;
+  private String username;
```
ggalmazor: The username and password are not required. You should implement a `Credentials` ValueObject class to store them and this class should have an `Optional<Credentials>` member instead of username and password

ggalmazor:
Hi @mayank8318! I'm sorry for having taken so long to check your work. I've had problems with my computer today :(

Anyway, I think we need to work a little bit more the relation between the view, the controller and the conf object. It's easier to see it with code and I've prepared a branch for you to check: https://github.com/ggalmazor/briefcase/commit/9084ca5b9e3192d31e5cfbd4e43f42ea084740e1

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialog.java
```diff
@@ -0,0 +1,80 @@
...
+
+public class AggregateServerConnectionDialog {
+  final AggregateServerConnectionDialogForm form;
+  private boolean isSuccessful;
+  private ServerConnectionInfo serverInfo = null;
+  private TerminationFuture terminationFuture = new TerminationFuture();
+  private AggregateServerConnectionConfiguration configuration;
+  Predicate<AggregateServerConnectionConfiguration> confValidator;
+
+  AggregateServerConnectionDialog(AggregateServerConnectionDialogForm form, Predicate<AggregateServerConnectionConfiguration> confValidator, AggregateServerConnectionConfiguration configuration) {
```
ggalmazor: I've prepared an example of what I meant by injecting the predicate here. It's on https://github.com/ggalmazor/briefcase/commit/9084ca5b9e3192d31e5cfbd4e43f42ea084740e1

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialogForm.java
```diff
@@ -0,0 +1,275 @@
...
+
+  public void onConnect(Runnable callback) {
+    connectButton.addActionListener(__ -> callback.run());
+  }
+
+  public void hideForm() {
+    this.setEnabled(false);
+  }
+
+  public Cursor getSavedCursor() {
```
ggalmazor: No object outside this class should use thiese methods. These are details that only the view needs to know.

mayank8318: Why do we have `onConnect()` and `triggerConnect()` in both the view and the controller?

ggalmazor: In this case, it feels redundant but it isn't. The view offers an `onConnect()` to let the controller bind to the UI event of clicking the button and the controller offers an `onConnect()` to let third parties bind to the business event of having a new valid Aggregate connection configuration object. The third parties, in this case, are the Pull and Push tabs.

src/org/opendatakit/briefcase/ui/reused/Credentials.java
```diff
@@ -0,0 +1,36 @@
+package org.opendatakit.briefcase.ui.reused;
+
+import java.util.Objects;
+import org.apache.http.util.TextUtils;
+
+public class Credentials {
+  private String username;
```
ggalmazor: Make it final

src/org/opendatakit/briefcase/ui/reused/Credentials.java
```diff
@@ -0,0 +1,36 @@
...
+  private String password;
```
ggalmazor: Make it final

src/org/opendatakit/briefcase/ui/reused/Credentials.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public Credentials(String username, String password) {
```
ggalmazor: Make it private

src/org/opendatakit/briefcase/ui/reused/Credentials.java
```diff
@@ -0,0 +1,36 @@
...
+  @Override
+  public int hashCode() {
+    return Objects.hash(username, password);
+  }
+
+  public static Credentials from(String username, String password) {
+    if (!(TextUtils.isEmpty(username) || TextUtils.isEmpty(password))) {
+      return new Credentials(username, password);
+    } else {
+      return null;
```
ggalmazor: We need to throw a `BriefcaseException` here. Returning null doesn't make sense in a `ValueObject`

src/org/opendatakit/briefcase/ui/reused/Credentials.java
```diff
@@ -0,0 +1,36 @@
...
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    Credentials that = (Credentials) o;
+    return Objects.equals(username, that.username) &&
+        Objects.equals(password, that.password);
+  }
+}
```
ggalmazor: You can add a `toString()` override as well

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+import org.opendatakit.briefcase.reused.BriefcaseException;
+
+public class AggregateServerConnectionConfiguration {
+  private URL url;
+  private Optional<Credentials> credentials;
+
+  public AggregateServerConnectionConfiguration(URL url, Optional<Credentials> credentials) {
```
ggalmazor: Can be private

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+    this.url = url;
+    this.credentials = credentials;
+  }
+
+  public static AggregateServerConnectionConfiguration from(String urlText, String username, char[] password) {
+    URL url;
+    try {
+      url = new URL(urlText);
+    } catch (MalformedURLException e) {
+      throw new BriefcaseException("Invalid URL " + urlText);
```
ggalmazor: Never lose the causing exception :)

```java
throw new BriefcaseException("Invalid URL " + urlText, e);
```

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+
+    return new AggregateServerConnectionConfiguration(url,
+        Optional.ofNullable(Credentials.from(username, Arrays.toString(password))));
+  }
+
+  public URL getUrl() {
+    return url;
+  }
+
+  public void setUrl(String url) {
```
ggalmazor: This method is not used. Remove it

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.Optional;
+
+import org.opendatakit.briefcase.reused.BriefcaseException;
+
+public class AggregateServerConnectionConfiguration {
+  private URL url;
```
ggalmazor: Make it final

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+  private Optional<Credentials> credentials;
```
ggalmazor: Make it final

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+    return Objects.equals(url, that.url) &&
+        Objects.equals(credentials, that.credentials);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(url, credentials);
+  }
+
+}
```
ggalmazor: You can add a `toString()` method too :)

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialogForm.java
```diff
@@ -0,0 +1,259 @@
...
+@SuppressWarnings("checkstyle:MethodName")
+public class AggregateServerConnectionDialogForm extends JDialog {
+  private JPanel dialog;
+  private JPanel actions;
+  private JButton cancelButton;
+  private JButton connectButton;
+  private JTextField urlField;
+  private JTextField usernameField;
+  private JPasswordField passwordField;
+  private Cursor savedCursor = null;
```
ggalmazor: Unused. Remove

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialogForm.java
```diff
@@ -0,0 +1,259 @@
...
+    addWindowListener(new WindowAdapterBuilder().onClosing(e -> dispose()).build());
+
+    dialog.registerKeyboardAction(e -> dispose(), getKeyStroke(VK_ESCAPE, 0), WHEN_ANCESTOR_OF_FOCUSED_COMPONENT);
+
+    cancelButton.addActionListener(e -> dispose());
+
+    connectButton.addActionListener(__ -> {
+      connectButton.setEnabled(false);
+      cancelButton.setEnabled(false);
+      Cursor.getPredefinedCursor(Cursor.WAIT_CURSOR);
```
ggalmazor: I think this should be:
```java
setCursor(Cursor.getPredefinedCursor(WAIT_CURSOR));
```

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialogForm.java
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  public void hideDialog() {
+    setVisible(false);
+  }
+
+  public void enableUI() {
+    connectButton.setEnabled(true);
+    cancelButton.setEnabled(true);
+    Cursor.getPredefinedCursor(DEFAULT_CURSOR);
```
ggalmazor: I think this should be:
```java
setCursor(Cursor.getPredefinedCursor(DEFAULT_CURSOR));
```

src/org/opendatakit/briefcase/ui/reused/Credentials.java
mayank8318: But, the username and password can be blank right?

ggalmazor: OK, let's back up a little:
- I think we agree that having a `Credentials` without a password or without username doesn't make sense. Therefore, we won't produce any instance of Credentials if we can't provide a username **and** a password.
- We also have decided that, in Briefcase, we use the `Optional<T>` type to model the absence of some value.
- We have defined the factory `Credentials.from(...)` as something that returns an instance of `Credentials`.

Taking all this into consideration, what would that `return null` mean?

If it's "there's nothing to build, so we don't return anything", returning an `Optional.empty()` would be more appropriate, but the signature prevents us from doing so and it would be a little weird that a factory for a Value object would return `Optional<T>` anyway.

If it's "we can't compute that, so we return null to signify an invalid execution of this factory", then we are inventing a new error management mechanism, where `null` means error. We should avoid doing this kind of things because Java already has an error management mechanism: exceptions.

Another insight would be that returning null is too open for interpretation (is it absence of value, is it a special value that means error, is it something else?), modeling return values with types and throwing exceptions is the way to go if we want to be explicit about our business logic.

mayank8318: So how do we handle blank credentials? If we call the factory method with the blank values, an exception would be thrown according to the current implementation. Should we change the way it is called?

mayank8318: I thought returning null here made sense since the Option.ofNullable() will handle that

ggalmazor: > I thought returning null here made sense since the Option.ofNullable() will handle that

Yeah, that would make sense if we didn't have `Optional<T>`. This idea is something we need to spread into the code more, but ideally, we should never use `null` for anything.

ggalmazor: Some examples:

```java

public static Credentials from(String username, String password) {
  if ((TextUtils.isEmpty(username) || TextUtils.isEmpty(password)))
    throw new IllegalArgumentException("You need to provide non empty username and password");
  return new Credentials(username, password);
}

// and then...

Optional<String> maybeUsername = Optional.ofNullable(username);
Optional<String> maybePassword = Optional.of(password).map(Arrays::toString).filter(String::isEmpty);
// A char[] can't ever be null. That's why we filter for non-empty strings at the end
Optional<Credentials> from = maybeUsername.isPresent() && maybePassword.isPresent()
    ? Optional.of(Credentials.from(maybeUsername.get(), maybePassword.get()))
    : Optional.empty();

return new AggregateServerConnectionConfiguration(url, from);
```

mayank8318:
@ggalmazor Why is export panel test failing? It is passing for me locally though.

ggalmazor:
> Why is export panel test failing? It is passing for me locally though.

UI tests are not so stable. Sometimes they fail. Don't worry

ggalmazor:
OK, I think this is ready... Now you could focus on using the dialog on the pull tab, for example

ggalmazor:
Hi @mayank8318!

Briefcase **really needs** to migrate the Pull and Push tabs to use the GUI Deisgner's `.form` files, etc., but, please, just change the old view. 

I didn't expect you to change all that by yourself because it's a lot of work and we will need to test a lot of things that aren't really related to the feature we're trying to add into Briefcase, making the review and release of this change much harder for everyone.

I know that changing the current layout is tricky, but can we try before deciding to make this PR a much bigger change, please?

If we get it right, we can work together on the migration after this PR is merged...

ggalmazor:
@mayank8318, this would get you going with the old layout:

```java

sourceConfPanel = new SourceConfigurationPanel(new SourceConfigurationPanelForm());

GroupLayout groupLayout = new GroupLayout(this);
groupLayout.setHorizontalGroup(groupLayout
    .createSequentialGroup()
    .addContainerGap()
    .addGroup(
        groupLayout
            .createParallelGroup(Alignment.LEADING)
            .addComponent(sourceConfPanel.getForm(), GroupLayout.PREFERRED_SIZE, GroupLayout.PREFERRED_SIZE,
                Short.MAX_VALUE)
            // scroll pane
            .addComponent(scrollPane, GroupLayout.PREFERRED_SIZE, GroupLayout.PREFERRED_SIZE,
                Short.MAX_VALUE)
            .addGroup(
                groupLayout.createSequentialGroup().addComponent(btnSelectOrClearAllForms)
                    .addPreferredGap(ComponentPlacement.RELATED).addComponent(lblDownloading))
            .addGroup(
                Alignment.TRAILING,
                groupLayout.createSequentialGroup().addPreferredGap(ComponentPlacement.RELATED)
                    .addComponent(btnTransfer).addPreferredGap(ComponentPlacement.RELATED)
                    .addComponent(btnCancel))).addContainerGap());
groupLayout.setVerticalGroup(groupLayout
    .createSequentialGroup()
    .addComponent(sourceConfPanel.getForm(), 36, 36, Short.MAX_VALUE)
    .addComponent(scrollPane, 200, GroupLayout.PREFERRED_SIZE, Short.MAX_VALUE)
    .addPreferredGap(ComponentPlacement.RELATED)
    .addGroup(
        groupLayout.createParallelGroup(Alignment.BASELINE)
            .addComponent(btnSelectOrClearAllForms).addComponent(lblDownloading)
            .addComponent(btnTransfer).addComponent(btnCancel)).addContainerGap());
setLayout(groupLayout);
```

mayank8318:
Ok, I will try to add the UI components to the existing code.

mayank8318:
@ggalmazor The form is not being rendered :
screenshot from 2018-03-21 18-15-32

ggalmazor:
@mayank8318 you need to use the view's `container` `JPanel`

mayank8318:
Getting this :
```java.lang.IllegalStateException: org.opendatakit.briefcase.ui.reused.SourceConfigurationPanelForm[,0,0,0x0,invalid,alignmentX=0.0,alignmentY=0.0,border=,flags=0,maximumSize=,minimumSize=,preferredSize=] is not attached to a horizontal group```

screenshot from 2018-03-21 19-27-23


screenshot from 2018-03-21 19-27-42

**Comment by mayank8318** [comment]
But if I use `return form.container` in getForm() and change the return type to `JPanel` it is working

**Comment by mayank8318** [comment]
@ggalmazor This works but it is not the right way to it, right? I am using swing components in the Controller. But the above-mentioned method is throwing an error(attached in the previous comment). What am I missing here?

File: src/org/opendatakit/briefcase/ui/PullTransferPanel.java [hunk 29]
```diff
@@ -293,7 +295,7 @@ public void focusLost(FocusEvent e) {
         .addGroup(
             groupLayout
                 .createParallelGroup(Alignment.LEADING)
-                .addComponent(sourceConfPanel.getForm(), GroupLayout.PREFERRED_SIZE, GroupLayout.PREFERRED_SIZE,
+                .addComponent(sourceConfPanel.getFormSourceConfig().getContainer(), GroupLayout.PREFERRED_SIZE, GroupLayout.PREFERRED_SIZE,
```
ggalmazor: This class must not know anything about how the internals of `SourceConfigurationPanel` work. I think it was OK like it was before.

All the things we need to do with the views must happen inside `SourceConfigurationPanel`.

So, let's revert this class and see wherever else we can work this out.

src/org/opendatakit/briefcase/ui/reused/AggregateServerConnectionDialog.java
```diff
@@ -13,13 +14,14 @@
   private final AggregateServerConnectionDialogForm form;
   private final List<Consumer<AggregateServerConnectionConfiguration>> onConnectCallbacks = new ArrayList<>();
 
-  AggregateServerConnectionDialog(AggregateServerConnectionDialogForm form, Predicate<AggregateServerConnectionConfiguration> confValidator, AggregateServerConnectionConfiguration initialConfiguration) {
+  AggregateServerConnectionDialog(AggregateServerConnectionDialogForm form, Predicate<AggregateServerConnectionConfiguration> confValidator, AggregateServerConnectionConfiguration initialConfiguration, BiConsumer<String, String> onSuccess) {
```
ggalmazor: Why do we need another callback? We already have an `onConnectCallbacks`... Also, why would this inner class execute some code from an ancestor so far apart?

src/org/opendatakit/briefcase/ui/PullTransferPanel.java
ggalmazor: Think about encapsulation:
- `PullTransferPanel` is a UI component that is composed of `SourceConfigurationPanel` and other children components
- `SourceConfigurationPanel` encapsulates the UI and behavior of the top panel of this UI, which lets the user select a source for Pulling/Pushing forms
- Does it make sense to let the `PullTransferPanel` the logic of changing `SourceConfigurationPanel`'s internal state and views? This would break encapsulation. Then, **we can't inject it with an `onSuccessCallback`**

src/org/opendatakit/briefcase/ui/reused/SourceConfigurationPanel.java
```diff
@@ -1,16 +1,22 @@
 package org.opendatakit.briefcase.ui.reused;
 
-import javax.swing.JPanel;
-
 public class SourceConfigurationPanel {
   private final SourceConfigurationPanelForm form;
+  private SourceDisplayPanelForm formSourceDisplay;
```
ggalmazor: To pull this off, I think we need a third "virtual" UI component. An empty `JPanel` would work, I think.

This JPanel is the container of this component and it will let you load **all the components** at once. Then you can make one of them visible, hide the rest, and toggle their visibility depending on the user's actions. 

We could create a `VirtualContainerForm` with just that `JPanel` and an API that would let us:
- `add(String name, JPanel view)`
- `navigateTo(String name)`

We could reuse this `VirtualContainerForm in other similar situations where we need to change what we're showing in a container

src/org/opendatakit/briefcase/ui/reused/SourceConfigurationPanelForm.java
```diff
@@ -13,10 +14,13 @@
 public class SourceConfigurationPanelForm extends JComponent {
   private JComboBox sourceComboBox;
   private JButton configureButton;
-  public JPanel container;
+  private JPanel container;
 
-  public SourceConfigurationPanelForm() {
+  private BiConsumer<String, String> onSuccess;
```
ggalmazor: Remove like on the other places

src/org/opendatakit/briefcase/ui/reused/SourceDisplayPanelForm.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.awt.GridBagConstraints;
+import java.awt.GridBagLayout;
+import javax.swing.JButton;
+import javax.swing.JComponent;
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+
+@SuppressWarnings("checkstyle:MethodName")
+public class SourceDisplayPanelForm extends JComponent {
```
ggalmazor: The layout seems odd, but we can work on this later. Let's focus on getting the code that changes views working.

src/org/opendatakit/briefcase/ui/PullTransferPanel.java
mayank8318: So let us maintain a `current form` variable in SourceConfigPanel. We will pass this to PullTransferPanel whenever it requests a top panel

src/org/opendatakit/briefcase/ui/reused/SourceConfigurationPanel.java
mayank8318: This will be in the Controller or the View class?

ggalmazor:
@mayank8318 check out https://github.com/opendatakit/briefcase/commit/4ded303a181e5b2d9f83e39d1e38a65ca5562bb5 where I have implemented the scaffolding for the navigation on top of your latest commit.

This is a self-similar design:
- A view exposes callbacks to tell its controller about changes on state
- A view exposes an API (methods) to let its controller change its initial state
- A component exposes callbacks to tell parent components about changes on state
- This structure of communication repeats on each layer.

On the Pull tab we have this hierarchy:
- The Main window and its view, which is a `JPanel` with a `TabbedPanel`
  - The `PullTransferPanel` class, which is at the same time controller and view all mixed up.
    - The `SourceConfigurationPanel`, which uses a `VirtualContainerForm`
      - The `SourceConfigurationPanelForm`
        - The `AggregateServerConnectionDialog`
      - The `SourceDisplayPanelForm`
    - Other stuff

Also, we should rename the `SourceConfigurationPanelForm`  to `SourceSelectionPanelForm`

mayank8318:
@ggalmazor The switching works perfectly now :) Both the UIs need a lot of polishing I know :(
So for different source options, will we wrap the configuration object in another class with an enum to specify which source it is?
Something like this:

```
public class Source {
private Object o;
enums....;
}
```

ggalmazor:
Hi @mayank8318! I think we will want to encode some things in Enums eventually but, for now, the simplest way to go in my opinion is to let the `SourceConfigurationPanel` component expose two different methods that the Pull/Push tabs would use to react when an aggregate connection configuration or a custom directory has been set.

I'd rename `SourceConfigurationPanel.onConfiguredSource(Consumer<AggregateServerConnectionConfiguration> callback)` to `SourceConfigurationPanel.onAggregateConnection(Consumer<AggregateServerConnectionConfiguration> callback)` and create a new `SourceConfigurationPanel.onCustomDir(Consumer<CustomDirConfiguration> callback)`.

That way, the Pull/Push tabs can attach to those events and react accordingly and all the code for aggregate connections and custom dirs gets nicely separated on their corresponding callbacks.

mayank8318:
@ggalmazor Should I make a value object for Custom Dirs as well?

ggalmazor:
In this case, I’d try to use just a `Path` as long as we don’t need to encapsulate more info

src/org/opendatakit/briefcase/ui/reused/SourceDisplayPanelForm.java
```diff
@@ -33,11 +34,16 @@ public void onReset(Runnable callback) {
     onResetCallbacks.add(callback);
   }
 
-  public void readConfiguration(AggregateServerConnectionConfiguration conf) {
+  public void readConfigAggregate(AggregateServerConnectionConfiguration conf) {
```
ggalmazor: You could give this method and the next the same name `readConfiguration` and let the overload work out which one must run...

src/org/opendatakit/briefcase/ui/reused/SourceSelectPanelForm.java
```diff
@@ -30,17 +39,37 @@ public SourceSelectPanelForm() {
   }
 
   private void triggerConfigure() {
-    AggregateServerConnectionDialog dialog = AggregateServerConnectionDialog.empty(__ -> true);
-    dialog.onConnect(this::trigureConfiguredSource);
-    dialog.getForm().setVisible(true);
+    String source = (String) sourceComboBox.getSelectedItem();
```
ggalmazor: This can be null. Wrap it inside an Optional

src/org/opendatakit/briefcase/ui/reused/SourceSelectPanelForm.java
```diff
@@ -30,17 +39,37 @@ public SourceSelectPanelForm() {
...
+    if (source.equals(SOURCE_AGGREGATE)) {
+      AggregateServerConnectionDialog dialog = AggregateServerConnectionDialog.empty(__ -> true);
+      dialog.onConnect(this::triggerAggregateConnection);
+      dialog.getForm().setVisible(true);
+    } else if (source.equals(SOURCE_CUSTOM_DIR)) {
+      // Invoke the file chooser here
+      WrappedFileChooser fc = new WrappedFileChooser(SourceSelectPanelForm.this,
```
ggalmazor: Don't use `WrappedFileChooser`. Use `FileChooser` instead

src/org/opendatakit/briefcase/ui/reused/SourceSelectPanelForm.java
ggalmazor: I think it's time for an Enum. This could work:

```java
interface Source<T> {
  Source<AggregateServerConnectionConfiguration> AGGREGATE = callback -> {
    AggregateServerConnectionDialog dialog = AggregateServerConnectionDialog.empty(__ -> true);
    dialog.onConnect(callback);
    dialog.getForm().setVisible(true);
  };
  Source<Path> CUSTOM_DIR = callback -> {
    WrappedFileChooser fc = new WrappedFileChooser(this,
        new ODKCollectFileChooser(this));

    int retVal = fc.showDialog();

    if (retVal == JFileChooser.APPROVE_OPTION) {
      Path path = Paths.get(fc.getSelectedFile().getAbsolutePath());
      callback.accept(path);
    }
  };

  void onSelect(Consumer<T> callback);
}

private void triggerConfigure() {
  Optional
      .ofNullable((Source<?>) sourceComboBox.getSelectedItem())
      .ifPresent(source -> {
        if (source.equals(AGGREGATE))
          AGGREGATE.onSelect(this::triggerAggregateConnection);
        if (source.equals(CUSTOM_DIR))
          CUSTOM_DIR.onSelect(this::triggerCustomDirSource);
      });
}
```

src/org/opendatakit/briefcase/ui/reused/SourceDisplayPanelForm.java
mayank8318: Ok :)

src/org/opendatakit/briefcase/ui/reused/SourceSelectPanelForm.java
mayank8318: Should I place this in the same class or create a new interface?

ggalmazor: New file :)

mayank8318: If I declare it in a new file, I cannot access the container required for making the file chooser. Is there a workaround for this?

ggalmazor: If you change `onSelect` to receive a `Container` and the `Consumer<T>`, you could pass it along with the callback to a `Source`

mayank8318:
@ggalmazor PR updated, please review :)

src/org/opendatakit/briefcase/ui/reused/Source.java
```diff
@@ -0,0 +1,46 @@
...
+import static org.opendatakit.briefcase.util.FileSystemUtils.isODKInstancesParentFolder;
+import static org.opendatakit.briefcase.util.FileSystemUtils.isUnderODKFolder;
+
+import java.awt.Container;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+import java.util.function.Consumer;
+
+public interface Source<T> {
```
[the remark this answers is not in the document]
mayank8318: @ggalmazor I have updated the PR. Did I rightly understand the concept?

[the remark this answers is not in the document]
ggalmazor: Perfect!

mayank8318:
@ggalmazor What should be the next target?

mayank8318:
We haven't tested the server connection part yet, currently passing it as `true`.

ggalmazor:
Don’t worry about that yet ;) Let’s make the push tab work with the new component :)

mayank8318:
@ggalmazor Tried to make it flexible so that it can work with Push and Pull tabs together

src/org/opendatakit/briefcase/ui/reused/SourceConfigurationPanel.java
```diff
@@ -13,10 +13,21 @@
   private final List<Consumer<AggregateServerConnectionConfiguration>> onAggregateSourceCallbacks = new ArrayList<>();
   private final List<Consumer<Path>> onCustomDirectorySourceCallbacks = new ArrayList<>();
   private final List<Runnable> onResetCallbacks = new ArrayList<>();
+  public static final String PULL = "pull";
+  public static final String PUSH = "push";
+
+  public SourceConfigurationPanel(String action) {
```
ggalmazor: This is good!

Putting aside that passing a string feels arbitrary (maybe better an enum for that), constructors should be much simpler. 

This kind of conditional construction logic would go better into a pair of factories. You could have one factory for push and another for pull. Then, the factories could pass to the constructor both views.

src/org/opendatakit/briefcase/ui/PushTransferPanel.java
```diff
@@ -211,33 +214,17 @@ public PushTransferPanel(TerminationFuture terminationFuture, BriefcasePreferenc
     formTransferTable.setSourceSelected(false);
     JScrollPane scrollPane = new JScrollPane(formTransferTable);
 
+    SourceConfigurationPanel sourceConfPanel = new SourceConfigurationPanel(PUSH);
```
ggalmazor: Better: `SourceConfigurationPanel.push()`

src/org/opendatakit/briefcase/ui/PullTransferPanel.java
```diff
@@ -284,7 +284,7 @@ public void focusLost(FocusEvent e) {
     formTransferTable.setSourceSelected(true);
     JScrollPane scrollPane = new JScrollPane(formTransferTable);
 
-    sourceConfPanel = new SourceConfigurationPanel();
+    SourceConfigurationPanel sourceConfPanel = new SourceConfigurationPanel(PULL);
```
ggalmazor: Better: `SourceConfigurationPanel.pull()`

mayank8318: Ok onto it :+1:

mayank8318:
@ggalmazor Please review the new changes :)

mayank8318:
Ok, I will try using the default URL since we know that it is correct and requires no validation

mayank8318:
The methods to update the tables should be passed as callbacks right?

ggalmazor:
> Ok, I will try using the default URL since we know that it is correct and requires no validation

You don't have to use any specific URL. Just make sure you enter the params correctly since we're not validating anything ;)

> The methods to update the tables should be passed as callbacks right?

Yes. The pull and push panels can subscribe to the `onAggregateConnection`, `onCustomDir`, and `onReset` events to react and change the UI state.

mayank8318:
@ggalmazor I have a couple of lab exams, I will update the PR by Friday without fail. Hope that's okay :)

ggalmazor:
Of course, don't worry about it! Good luck with your exams!

mayank8318:
@ggalmazor Sorry for not updating the PR as promised. I have my final exams this week. I will try to update it if I get time :)

ggalmazor:
Don't worry, @mayank8318! Your exams are more important! Best of luck!

ggalmazor:
@mayank8318, we will continue the development of this PR in #463. Could you close this one, please?

Thanks!

mayank8318:
@ggalmazor Yes, I will close it now. Learned a LOT in this PR. 
Thank you :)

PR #414, opened 2018-03-14
Issue 20 push from cli

### LINKED ISSUE #20: Extend command line interface to match graphical user interface
Currently, Briefcase's command line interface only allows for pulling data from ODK Aggregate. It would be great to have support for also pushing data to servers.

ggalmazor:
Closes #20

Manually executed the new CLI operation forcing errors (wrong form_id, wrong credentials, etc.) and actually getting some forms to my GAE instance

Manually executed `-pa` flag and veryfied that it outputs the expected messages and that exits with a non-zero status

This solution mirrors what we are currently doing for pushing forms.

This PR deprecates `-pa` flag in favor of `-plla` to be more symetric with `-psha`.

Running Briefcase with `-pa` will show a message, show help and exit with non-zero status:
```
The param -pa has been deprecated. Run Briefcase again with -plla instead
```

### DIFF
--- a/src/org/opendatakit/briefcase/util/PushFromServerException.java
+++ b/src/org/opendatakit/briefcase/util/PushFromServerException.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (C) 2018 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.opendatakit.briefcase.util;
+
+import static java.util.stream.Collectors.joining;
+
+import java.util.List;
+import org.opendatakit.briefcase.model.FormStatus;
+import org.opendatakit.briefcase.reused.BriefcaseException;
+
+public class PushFromServerException extends BriefcaseException {
+  public PushFromServerException(List<FormStatus> forms) {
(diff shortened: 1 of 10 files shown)


src/org/opendatakit/briefcase/util/PushFromServerException.java
```diff
@@ -0,0 +1,28 @@
+/*
```
ggalmazor: This file should have gone with the first commit... Sorry!

lognaturel: If you want, it should be pretty easy to fix. I would do an interactive rebase, leave this commit open for editing, `git reset PushFromServerException.java` to unstage this one, `git stash`, `git rebase --continue` then interactive rebase again leaving the first commit open, `git stash apply`, `git add -A`, git commit --amend`. Ok, maybe it's not super easy...

ggalmazor: I think you're right! I'll do that :)

ggalmazor: OK, I did it but I had to do something different:
- Interactive rebase of https://github.com/opendatakit/briefcase/pull/414/commits/77dba135eb2f8f2742682b35b0fd1cb38aca64bb to edit it
- Since git first applies the commit and then it lets you edit, I had to `git rm` the file
- Interactive rebase of https://github.com/opendatakit/briefcase/pull/414/commits/42b7a5c7e1399abb58e4dd6e3c258051814484b1 to edit it
- Create the file from the raw version in https://github.com/opendatakit/briefcase/pull/414/commits/77dba135eb2f8f2742682b35b0fd1cb38aca64bb that still exists on GitHub
- Profit :1st_place_medal:

lognaturel: Oh, I'm surprised the `reset` didn't do the trick! But hey, you got to the most important step. 🥇 🤣

kkrawczyk123:
### Tested with success!
Verified on Ubuntu, MacOS and Windows and gained the same results.
I can only say that those processes take a while.

src/org/opendatakit/briefcase/util/PushFromServerException.java
ggalmazor: It turns out that I could have done a `git reset HEAD~` or maybe even a `git reset PushFromServerException.java HEAD~`

PR #417, opened 2018-03-18
show exporting feedback label. closes #402

### LINKED ISSUE #402: Provide feedback when export is ongoing
Prior to multiple exports being added, it was obvious that an export was ongoing because the export button would be disabled and the status button was clickable. Now there is no feedback that an export is ongoing.

A simple approach would be to add "Exporting..." to the right of the `Select All` button with dots added each time an event is fired. If all exports complete, remove the text. There's some chance that completion won't be detected because the event bus is not reliable but because it's just text and not enabled/disabled buttons it's not a very big deal.

@ggalmazor interested in knowing what you think is possible here.

Thodoris1999:
Closes #402 

The "Exporting..." label is not positioned correctly as described in the issue. I have not worked with windows in java so I am still figuring it out

test on Windows 10

 A new Timer and TimerTask object is created each time the label gets updated, I'm not sure if they get garbage collected. I didn't notice any lag though.


ggalmazor:
Thanks for your contribution @Thodoris1999!

Regarding the UI placement of the new label, you need to put the new label as the first child of the grid with `binding="rightActions"`, like this (right side is the correct one):

image

Regarding your implementation, some pointers:
- Could you migrate your code to use `ScheduledExecutorService` instead of the `Timer` API? `ScheduledExecutorService` is a better, more up-to-date API to solve scheduling.
- Your current solution will create a new timer for every progress event we receive (that's OK) but it won't cancel any previous timer that was already going on. It should cancel them to ensure that, at any given time, we only have 1 countdown timer that will hide the label
  - Using `ScheduledExecutorService`, you could save the resulting `ScheduledFuture` in the view object to cancel it in case it exists. Please, use `Optional<T>` to avoid having to do null checks.
  - Since many threads will be sending events, you have to think about concurrent access to that object (maybe `synchronize` `updateExportingLabel()`?)

Thodoris1999:
Thank you for the detailed review! However, I don't understand how I save a `ScheduledFuture` and what the view object is. Is there a way to avoid having the `ScheduledFuture` as a class variable?

ggalmazor:
> how I save a ScheduledFuture and what the view object is. 
We call views the classes that have a companion `.form` file. They represent the UI and normally will handle UI events, etc. They expose a low-level UI API to the component controllers react to UI events or set the UI state.

In this case, `ExportPanel` would be the component controller and `ExportPanelForm` the view.

>Is there a way to avoid having the ScheduledFuture as a class variable?

I don't think so. We want to cancel any previously scheduled execution of `hideExporting()` before creating a new one.

To do so, you can save into a field the `ScheduledFuture` you get when creating the scheduled task, and then cancel it if it's set (from a previous execution) like so:

```java
public class ExportPanelForm {
  private static final ScheduledExecutorService SCHEDULED_EXECUTOR = new ScheduledThreadPoolExecutor(1);
  private Optional<ScheduledFuture<?>> scheduledHideExportingLabel = Optional.empty();
  
  //...

  synchronized public void updateExportingLabel() {
    // update the label text
    scheduledHideExportingLabel.ifPresent(scheduledFuture -> scheduledFuture.cancel(false));
    scheduledHideExportingLabel = Optional.of(SCHEDULED_EXECUTOR.schedule(this::hideExporting, 5, SECONDS));
  }

  // ...
}
```
We wrap the `ScheduledFuture` inside an `Optional` to avoid having to do null checks (and because we model the absence of value with the `Optional` type elsewhere in Briefcase).

ggalmazor:
I almost forgot! I pushed a branch to my fork should you want to see it in action or cherry pick the changes https://github.com/ggalmazor/briefcase/commit/60d6a8d03abf1ea9545b7d0a8eef819a1234cd27

Note that I removed the `form.hideExporting()` on the `ExportPanel` because the export process happens asynchronously and that line hides the message too soon.

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -178,7 +181,9 @@ private void export() {
               terminationFuture
           );
         });
+    //notify UI
     form.enableUI();
+    form.hideExporting();
```
ggalmazor: We have to remove this call from here. Exporting happens asynchronously and this will get executed before the export process ends.

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.java
```diff
@@ -41,8 +50,12 @@
   private JPanel rightActions;
   private JButton selectAllButton;
   private JButton clearAllButton;
+  private JLabel lblExporting;
   JButton exportButton;
   private boolean exporting;
+  private static final ScheduledExecutorService SCHEDULED_EXECUTOR =
```
ggalmazor: Move this member up, with the other `static final` member

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.java
```diff
@@ -41,8 +50,12 @@
   private JPanel rightActions;
   private JButton selectAllButton;
   private JButton clearAllButton;
+  private JLabel lblExporting;
```
ggalmazor: I forgot this one before, sorry. Rename the field to `exportingLabel` for conformity with other fields.

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.java
```diff
@@ -120,6 +133,22 @@ private void setExporting(boolean active) {
...
+  synchronized public void updateExportingLabel() {
+    String text = lblExporting.getText();
+    if (text.equals(EXPORTING_DOT_ETC)) {
+      text = "Exporting.";
+    } else {
+      text += ".";
+    }
+    lblExporting.setText(text);
+
+    //set a timeout of 5 secs and hide label if there are no new updates
```
ggalmazor: I think the code is quite self-explaining. Remove comment

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.java
```diff
@@ -120,6 +133,22 @@ private void setExporting(boolean active) {
...
+    scheduledHideExportingLabel.ifPresent(scheduledFuture ->
```
ggalmazor: For conformity with the rest of the code (at least the new parts we're working on), don't break long lines.

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.java
```diff
@@ -120,6 +133,22 @@ private void setExporting(boolean active) {
...
+            scheduledFuture.cancel(false));
+    scheduledHideExportingLabel = Optional.of(SCHEDULED_EXECUTOR.schedule(this::hideExporting,
```
ggalmazor: For conformity with the rest of the code (at least the new parts we're working on), don't break long lines.

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.java
```diff
@@ -120,6 +133,22 @@ private void setExporting(boolean active) {
...
+            5, TimeUnit.SECONDS));
```
ggalmazor: Also, `import static` the `TimeUnit.SECONDS` member to be more concise. This line should read: 

```java
scheduledHideExportingLabel = Optional.of(SCHEDULED_EXECUTOR.schedule(this::hideExporting, 5, SECONDS);
```

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -104,6 +105,8 @@ public ExportPanel(TerminationFuture terminationFuture, ExportForms forms, Expor
 
 
     form.onExport(() -> backgroundExecutor.execute(() -> {
+      //notify user that export started
```
ggalmazor: Ugh, sorry! This comment slipped too. I think the code is self-explanatory. Remove comment :)

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.java
```diff
@@ -31,6 +39,10 @@
 
 @SuppressWarnings("checkstyle:MethodName")
 public class ExportPanelForm {
+  private static final String EXPORTING_DOT_ETC = "Exporting..........";
+  private static final ScheduledExecutorService SCHEDULED_EXECUTOR =
```
ggalmazor: Sorry, this one slipped too! Don't break lines

ggalmazor:

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.java
```diff
@@ -120,6 +133,19 @@ private void setExporting(boolean active) {
     exporting = active;
   }
 
+  synchronized public void updateExportingLabel() {
+    String text = exportingLabel.getText();
```
dcbriccetti: These six lines could be one by using ? :  .

ggalmazor: I think you're right, but I think the result could be harder to understand than in its current state...
There's something not quite right in checking against `EXPORTING_DOT_ETC` and then building everything else with magic strings... For example, if I change the value of `EXPORTING_DOT_ETC` and I forget to change that block, I break it.

I don't have a better solution right now, but I'm sure we will be coming back to this block...

ggalmazor: In other words: using a ternary doesn't help us come to a better solution, in my opinion

dcbriccetti: Glad you’ll be coming back.

dcbriccetti: As you have probably noticed, I love the *Conditional Operator* (the only ternary¹ operator).
https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.25

¹composed of three parts

ggalmazor: Conditional expressions are the best conditionals :)

Thodoris1999:
did I miss anything? As for what you mentioned, how about declaring a `EXPORTING = "Exporting"` and then have `EXPORTING_DOT_ETC` be equal to `EXPORTING + "........."`?

ggalmazor:
> did I miss anything? As for what you mentioned, how about declaring a EXPORTING = "Exporting" and then have EXPORTING_DOT_ETC be equal to EXPORTING + "........."?

Don't worry, @Thodoris1999! I think you don't have to change anything.

I just think that there's a hidden behavior that might benefit from its own abstraction, but **I really think it's not justified yet**. 

Maybe, if we find the same need for cycling a label's text on other tabs, we can come back here and extract some method that handles it. 

I've written a [post about this](https://medium.com/@ggalmazor/indirection-without-abstraction-an-example-31ef14aa2f09) this morning.

Thodoris1999:
@ggalmazor I like the Cycler class! There are also two more occurrences of the cycling code for pushing and pulling (which is why I wrote it that way, to make it uniform). Can I implement it and refactor those other two as well?

ggalmazor:
> There are also two more occurrences of the cycling code for pushing and pulling (which is why I wrote it that way, to make it uniform)

Oh, I didn't know about that... I'd suggest separating that change into another PR. 

@yanokwa, @lognaturel, what do you think about this? Could we create a follow-up issue to refactor this triplicated code and improve the design? @Thodoris1999 could work on that right away after merging this PR.

Thodoris1999:
Can I also place the label on the left actions? The other two are placed that way and that will remove the "shaking" of the label.

ggalmazor:
> Can I also place the label on the left actions? The other two are placed that way and that will remove the "shaking" of the label while cycling.

Feedback for the export button should be next to the button... Can you try giving the label a fixed minimum width to see if the text stops shaking?

Thodoris1999:
I assume I'll b doing the same for the other labels.

> Feedback for the export button should be next to the button... Can you try giving the label a fixed minimum width to see if the text stops shaking?

How about to the right of the export button?

Also, is crating a `CyclingLabel extends JLabel` a better approach? It will allow us to move the time out logic there too. I can't think of any other possible use for `Cycler` outside labels.

ggalmazor:
> How about to the right of the export button?

I think it's OK to the left of the button. 

From a UX perspective, we're trying to set positive or neutral actions on the right bottom corner and negative actions on the left corner. Therefore, any label should be interior.

> Also, is crating a CyclingLabel extends JLabel a better approach?

Since the behavior we're defining doesn't need to be dynamic depending on the subtype of `JLabel` we need to support, we can't use inheritance. Inheritance (extension) is used to implement polymorphic types, which is not the case here. We need to use composition.

In this case, we're perfectly OK by composing a `JLabel` and our `Cycler` class (when we have it) inside the `ExportPanelForm`. I wouldn't make any change regarding this.

> It will allow us to move the time out logic there too.

I think it doesn't make sense coupling a scheduling behavior with the text cycling behavior. Doing so would only limit our options to compose those behaviors separately and it would make harder to test them in isolation.

Again, since we are composing, it's OK to wrap the cycling stuff in one class and the scheduling stuff in another class, and then compose them with the `JLabel` in the `ExportPanelForm`.

> I can't think of any other possible use for Cycler outside labels.

That doesn't mean we should make the decision of limiting it just for `JLabel` instances. By defining small pieces that do just one thing, we gain the option of reusing them and composing our software the way we need.

Also, bear in mind that we will need to provide some unit tests of the `Cycler` class once we work on it and it will be much easier if that class isn't bound to UI components.

Thodoris1999:
Thank you for such detailed answers :D

kkrawczyk123:
Verified on: Ubuntu, MacOs and Windows, I have some suggestions:

- We should probably move it to the left side as showing status it is shaking, you can notice that especially when bulk export because the amount of dots is changing. (Downloading status and Uploading on pull and push are much more esthetic)

- I have noticed some weird graphical effect on ubuntu, screenshot attached. It is not connected with this PR it occurs on master too - but when exporting status added it’s more visible.
screenshot_weird graphical effect

- When something interrupts exporting process (I didn’t add PEM key file for encrypted form - try it to reproduce that) and dialog window display - exporting is not ongoing, the export button is available but “Exporting……” status is displayed, I reproduced that on all three systems. I guess that it should be fixed. Screenshot attached:
screenshot_broken export

- We should probably consider disabling Export button in the middle of ongoing exports (as pull and push buttons are)

Thodoris1999:
I get `export_button_should_be_enabled_when_there_are_no_configuration_errors` even though I did not mess with the commit. 
Also, since i  remove the `hideExporting()` call at the end of the `void export()` the label doesn't go away (except from timeout). I searched in `org.opendatakit.briefcase.model` for a callback when all exports are finished but the ones I found are called when a single form is exported.

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.java
```diff
@@ -64,6 +64,9 @@ private ExportPanelForm(ConfigurationPanel confPanel, FormsTable formsTable) {
     this.formsTableForm = formsTable.getView();
     $$$setupUI$$$();
 
+    exportingLabel.setMinimumSize(exportingLabel.getPreferredSize());
```
ggalmazor: Please, set those params using IntelliJ's UI Designer.

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.java
```diff
@@ -64,6 +64,9 @@ private ExportPanelForm(ConfigurationPanel confPanel, FormsTable formsTable) {
...
+    exportingLabel.setPreferredSize(exportingLabel.getPreferredSize());
```
ggalmazor: Remove this one: you're setting the value to its current value :)

ggalmazor:
@Thodoris1999 
> I get export_button_should_be_enabled_when_there_are_no_configuration_errors even though I did not mess with the commit.

Don't worry. That test is unstable and will fail intermittently. We need to improve that, but I'll create an issue for that.

> Also, since i remove the hideExporting() call at the end of the void export() the label doesn't go away (except from timeout). I searched in org.opendatakit.briefcase.model for a callback when all exports are finished but the ones I found are called when a single form is exported.

I think that's the intended behavior. With the current concurrency model, we can't know for sure when all exports end to force hiding the label. I'd suggest decreasing the delay to 2 seconds.

Thodoris1999:
@ggalmazor I know it looks silly, butt doesn't works without setting both... Try it yourself.

ggalmazor:
> I know it looks silly, but it doesn't works without setting both... Try it yourself.

Yeah... Swing... ¯\\\_(ツ)_/¯

I'm happy just by setting those on the UI Designer, really :)

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -178,6 +180,7 @@ private void export() {
               terminationFuture
           );
         });
+    //notify UI
```
ggalmazor: Oops, single comment slipped again. Remove :)

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.form
```diff
@@ -57,6 +51,14 @@
             <properties/>
             <border type="none"/>
             <children>
+              <component id="b7dc5" class="javax.swing.JLabel" binding="exportingLabel">
+                <constraints/>
+                <properties>
+                  <minimumSize width="86" height="21"/>
+                  <preferredSize width="86" height="21"/>
+                  <text value="Exporting.........."/>
```
ggalmazor: So, I've tested this and you should give it at least 110 width... Otherwise it doesn't fit.

I'd like to suggest using "Exporting..." (three dots) and a width of 80 to make it compact.

This also is missing a `<visible value="false"/>` to make it not visible when you first visit the tab.

src/org/opendatakit/briefcase/ui/export/ExportPanelForm.java
```diff
@@ -31,6 +40,9 @@
 
 @SuppressWarnings("checkstyle:MethodName")
 public class ExportPanelForm {
+  private static final String EXPORTING_DOT_ETC = "Exporting..........";
```
ggalmazor: If you change it to "Export...", remember to update this one too.

Thodoris1999:
@ggalmazor thank you for replying so quickly and for all the help and insight :D I set the timout to 3 seconds because it did lag for 3 seconds once for me, causing the label to disappear before the exports finished

ggalmazor:
Thank you, @Thodoris1999, for you contribution :)

lognaturel:
Thank you @Thodoris1999!

> Could we create a follow-up issue to refactor this triplicated code and improve the design?

This is perhaps replaced by #421?

ggalmazor:
> > Could we create a follow-up issue to refactor this triplicated code and improve the design?
>
>This is perhaps replaced by #421?

I think so!

PR #427, opened 2018-04-03
Issue 423 alphanumeric form versions

### LINKED ISSUE #423: Allow non-integer strings in the form definition version field
#### Software versions 
Briefcase v1.9.0

#### Problem description
Briefcase fails to validate the form definition, even though the form is valid (created through pyxform, validated through ODK Validate). Once we know what the issue is we can fix it on the [KoBo](https://github.com/kobotoolbox/kpi) side, but maybe this is also a bug on the Briefcase side given the error message.

```
Fetching form definition
resolving against briefcase form definitions
Error parsing form definition: org.opendatakit.aggregate.exception.ODKIncompleteSubmissionData: The data model does not have an id or xmlns attribute.  Add an id="your.domain.org:formId" attribute to the top-level instance data element of your form.
```

1. Create any form with kobo.humanitarianresponse.info, deploy
1. Set Briefcase with the correct credentials and hit Pull

#### Expected behavior
Form definition is parsed successfully OR error message displays relevant information 

#### Other information 
This applies to any form generated through KoBo's `kpi` tool; whereas forms created through KoBo's legacy `kobocat` tool succeed with Briefcase. 

Here are the XML files that [don't work](https://0bin.net/paste/bnxbtqchxjkUOU4O#veNWV211FExy2nvX7xwxQ0hTJfErDf7cxxKXIoAHIwC) (kpi-generated) and [do work](https://0bin.net/paste/fuoXYCT2PkH-gfSF#1tDxaOdQkAWSiZArcWRy1fTfaA+ZWzBzPXWdow0RAv+) (kobocat generated). The only minor difference is the name of the XML data model/formID. 

[Related code section in the parser](https://github.com/opendatakit/briefcase/blob/57eea1d7782da7ad77794530d503f182e02f6ad8/src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java#L560-L563)

ggalmazor:
Closes #423 

Created new unit tests that verify the change in behavior:
- A new test that parses an XForm with different version values (null, numeric and alphanumeric)
- (collateral) A new test that shows and verifies `XFormParameters.compareTo()` method

Also installed manually a form with alphanumeric version into Briefcase's storage directory and checked that Briefcase launches and is able to export the form. This is the form I've used: 
[alphanumericversion.xml.txt](https://github.com/opendatakit/briefcase/files/1871367/alphanumericversion.xml.txt)

Tested pulling and pushing forms with alphanumeric versions from/to a Kobo server.

This is the smallest change I could think of: just avoid making any type change of the incoming version string.

None I can't think of. This change relaxes prior constraints.

### DIFF
--- a/test/java/org/opendatakit/briefcase/model/BriefcaseFormDefinitionTest.java
+++ b/test/java/org/opendatakit/briefcase/model/BriefcaseFormDefinitionTest.java
@@ -1,6 +1,5 @@
 package org.opendatakit.briefcase.model;
 
-
 import static java.nio.file.Files.createTempDirectory;
 import static java.nio.file.Files.delete;
 import static java.nio.file.Files.readAllBytes;
--- a/test/java/org/opendatakit/briefcase/model/BriefcaseFormDefinitionTest.java
+++ b/test/java/org/opendatakit/briefcase/model/BriefcaseFormDefinitionTest.java
@@ -1,0 +1,68 @@
+package org.opendatakit.briefcase.model;
+
+
+import static java.nio.file.Files.createTempDirectory;
+import static java.nio.file.Files.delete;
+import static java.nio.file.Files.readAllBytes;
+import static java.nio.file.Files.write;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+import static org.opendatakit.briefcase.ui.SwingTestRig.classPath;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collection;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.opendatakit.briefcase.util.BadFormDefinition;
+
+@RunWith(value = Parameterized.class)
+public class BriefcaseFormDefinitionTest {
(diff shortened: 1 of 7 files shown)


test/java/org/opendatakit/briefcase/model/BriefcaseFormDefinitionTest.java
```diff
@@ -0,0 +1,67 @@
...
+        {"Numeric version", "1"},
+        {"Alphanumeric version", "vHG9gooNbDqA2DxdU9cP5Q"},
+    });
+  }
+
+  @Before
+  public void setUp() throws IOException {
+    formsDir = createTempDirectory("briefcase_test");
+    form = formsDir.resolve("form.xml");
+    String tpl = new String(readAllBytes(classPath("org/opendatakit/aggregate/parser/test-version.tpl.xml")), Charset.defaultCharset());
```
lognaturel: Templating forms is cool! Would not have thought to do that.

lognaturel:
@kkrawczyk123 I am merging this without usual QA because there should be no effect for most users -- it will be verified as part of the release testing. @ggalmazor has verified it with a Kobo server already.

@ggalmazor maybe at some point we can do away with the `versionString` field?

ggalmazor:
> maybe at some point we can do away with the versionString field?

Indeed :)

yanokwa:
So we don't forget, can you file an issue?

PR #428, opened 2018-04-03
Issue 401 sentry sensitive info

### LINKED ISSUE #401: Don't log usage issues to Sentry, especially if they include user info
Currently, some error text logged to Sentry includes user info such as full paths on their machines or form elements.

For example:
- lines 428 and 431 of `FileSystemUtils`
- lines 136 and 138 of `AggregateUtils`

These logged items don't correspond to problems with Briefcase but rather with current state (missing file, server down, etc). Ideally this information makes it back to the user through the primary UI (GUI or CLI). The personal info should never be sent to Sentry.

Those specific examples could be logged at info level, I think.

ggalmazor:
Closes #401

The project compiles and launches GUI.
Run all tests

This is the smallest change that solves this problem towards the next release.

The best possible solution would be to explicitly express our intent of sending error reports to Sentry. We need that to send to Sentry only actionable errors, avoiding to send errors related to the user's environment or issues with specific forms and submission files.
 
In the process of removing sensitive info from the errors, we have lost information that _maybe_ is valuable while debugging. I think this is a risk we can assume while we do a revision or work on a way to separate the errors that should go to Sentry from other errors.

### DIFF
--- a/src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
+++ b/src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
@@ -125,12 +125,13 @@ public class BaseFormParserForJavaRosa implements Serializable {
 
   private static void redirectOutput() {
     File jrLogFile = new File(new StorageLocation().getBriefcaseFolder(), ".briefcase-javarosa.log");
+    log.info("Redirecting javarosa output to {}", jrLogFile);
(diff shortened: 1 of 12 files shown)


lognaturel:
I think there are a couple things we may want to adjust as reports come in and we see how much info is actually in the exceptions but this is a clear improvement.

PR #432, opened 2018-04-03
Upgrade JR dependency to version 2.9.0

### LINKED ISSUE #431: Upgrade JR to 2.9.0

ggalmazor:
Closes #431

Compiled and launched Briefcase, and tried to export forms

Runs all tests

N/A

I had to remove the override of `XFormParserWithBindEnhancements.parseBind(Element e)`. It seems that it only did some efficiency tricks but it's not clear why that code was there.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ dependencies {
         exclude group: 'commons-logging'
     }
     compile group: 'net.sf.kxml', name: 'kxml2', version: '2.3.0'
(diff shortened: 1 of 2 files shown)


lognaturel:
😱 Unused imports!! ;)

ggalmazor:
Damn!

lognaturel:
As @dcbriccetti always says, the best code is deleted code. ✨

PR #446, opened 2018-04-09
Issue 401 Remove sentitive info from Sentry logs

ggalmazor:
Continues work related to #401 and improves the Sentry appender to toggle it depending on the tracking consent.

- Produced an export error by tampering a submission file
- Checked that disabling the tracking consent, the events don't reach Sentry
- Checked that re-enabling the tracking consent, the events reach Sentry

Regarding the anonymization of the `ConnectTimeoutException`:
Since we don't have control of this `ConnectTimeoutException`, we only can bypass the exception's message to avoid logging the host information that comes on it.

Regarding the tracking consent:
Sentry doesn't have a global toggle but it provides a callbacks system to inspect every Event and decide if it has to reach Sentry or not.

In this case, I'm taking advantage of this callback to check the tracking pref to prevent sending any Event if the users disables tracking.

None.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/Launcher.java
+++ b/src/org/opendatakit/briefcase/Launcher.java
@@ -45,7 +45,7 @@ public class Launcher {
     if (!appPreferences.hasKey(BRIEFCASE_TRACKING_CONSENT_PROPERTY))
       appPreferences.put(BRIEFCASE_TRACKING_CONSENT_PROPERTY, TRUE.toString());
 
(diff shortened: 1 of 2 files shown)


ggalmazor:
@lognaturel, @yanokwa, I've added a commit to this PR that will make Sentry check the tracking pref to decide if it sends the event or not. If a user disables tracking, no events will reach Sentry.

PR #448, opened 2018-04-10
Clearer language about analytics

yanokwa:
Add even more precise language about what we are collecting and why


lognaturel:
This sounds good to me. Is https://docs.opendatakit.org/security-privacy/ still ok?

PR #457, opened 2018-04-28
Improve export operation code

ggalmazor:
This PR adds a new implementation of the Export process that solves some of the problems the old process presented, especially when exporting a large number of submissions. The main problems were:
- Contingency issues on shared state that performed badly when exporting many forms
- Saturation of the EventBus with progress events

Thanks to the changes this new process introduces (mainly by using Java8 Streams), this also paves the way for follow-up PRs with error management and user feedback improvements.

The old export process has been preserved and it can be toggled with a new GUI configuration setting and CLI flag. The new export process is enabled by default unless users configure it otherwise.

The commit https://github.com/opendatakit/briefcase/pull/457/commits/b4dc0b4b2cdf816ee96290a24ce1ef8017d6047d a comprehensive, outside-in test suite that compares outputs from the old and the new processes to verify that we haven't broken any feature:
- Basic exporting of plain and encrypted forms, with all supported data types
- Exporting forms with cascading and non-cascading repeat groups
- Exporting with a date range submission filter
- Exporting forms with media attachments, both plain and encrypted, covering various corner cases
- The overwrite files feature

    For more info on this, [check the test files](https://github.com/ggalmazor/briefcase/tree/improve_export_performance/test/java/org/opendatakit/briefcase/export) and [the form files they use](https://github.com/ggalmazor/briefcase/tree/improve_export_performance/test/resources/org/opendatakit/briefcase/export)

I have also tested manually the feature toggle on the GUI and the new CLI flag to run the old version of the export process.

#### Notes about performance

My machine is an i7 with 6 cores (12 threads), it has 32 GB of memory and I'm running the tests on an nvme hard drive.

Looking to the time it takes to export a form with 50K submissions, I get mixed results that I can't quite explain. In my machine, the old implementation consistently produces the result CSVs in ~17 seconds and the new algorithm dances around that value, sometimes being slower than that.

This is a list of changes I've tried to pinpoint a source for this inconsistent durations:
- Use the original submission file reading and sorting code to avoid any problem derived from using Java NIO2 or Streams for that.
- Reduce per-submission created number of objects by removing wrapping objects like Optionals and using primitives downstream, to reduce garbage collection problems
- Start Briefcase with a bigger minimum heap memory settings to reduce resizing during garbage collection
- Pass the OutputStreamWriter instances downstream to directly write in the files where any value is produced, instead of producing the values, collecting and writing them in the terminal operation.

None of this changes made the new implementation run consistently faster than the old one by themselves.

I'm curious about performance metrics on systems with fewer cores and memory. Any feedback will be very welcome.

[This is the dataset I've used](https://drive.google.com/open?id=1czxbJk3dIw3CY1AGlTlYozwyyKLNK_q8). You can decompress that into your Briefcase storage directory.

This is a total revamping of the export process. Progressively evolving the old one was considered, but it was too challenging to deal with a procedural style codebase which heavily relies on side effects to produce results.

Almost all the code of the new implementation started as a copy&paste of the old one, and then it was refactored using the outside-in tests as a baseline.

Going forward, there is a lot of room for design and performance improvements. The resulting code base should be more declarative and rely on immutable, side-effect free code as much as possible, which is easier to understand and refactor.

Even though the test suite has a really high coverage metrics both on the old and new processes, it is possible that some use case slipped our safety net. That's why we provide a feature toggle to reduce the impact of any issue, should this happen.

Nope.


src/org/opendatakit/briefcase/export/ExportProcessTracker.java
```diff
@@ -0,0 +1,61 @@
...
+  long exported;
+  private int lastReportedPercentage = 0;
+
+  public ExportProcessTracker(FormDefinition form, long total) {
+    this.form = form;
+    this.total = total;
+  }
+
+  public void incAndReport() {
+    this.exported++;
```
dcbriccetti: You sure you want the “this.”?

src/org/opendatakit/briefcase/export/SubmissionMetaData.java
```diff
@@ -0,0 +1,154 @@
...
+   * <p>
+   * The value gets mapped to an {@link OffsetDateTime}
+   */
+  Optional<OffsetDateTime> getSubmissionDate() {
+    if (submissionDate == null)
+      submissionDate = root.getAttributeValue("submissionDate")
+          .map(iso8601DateTime -> {
+            // Sometimes we can get just an ISO8601-ish string lacking of
+            // the minutes part of the offset block. Fix that.
+            return iso8601DateTime.length() - iso8601DateTime.lastIndexOf("+") == 3
```
dcbriccetti: This assumes everybody is *ahead* of UTC. The parse fails with `2018-01-15T00:45:36.127-08`, for example.

src/org/opendatakit/briefcase/export/SubmissionMetaData.java
```diff
@@ -0,0 +1,154 @@
...
+                ? iso8601DateTime + ":00"
+                : iso8601DateTime;
```
dcbriccetti: That doesn’t work. How about this?
```
            return iso8601DateTime.length() - iso8601DateTime.lastIndexOf(":") == 3
                ? iso8601DateTime
                : iso8601DateTime + ":00";
```

src/org/opendatakit/briefcase/export/SubmissionMetaData.java
```diff
@@ -0,0 +1,154 @@
...
+          })
+          .map(OffsetDateTime::parse);
```
dcbriccetti: Do you need to catch DateTimeParseException from parse? How confident are you the date strings will be in the right format?

src/org/opendatakit/briefcase/export/Model.java
```diff
@@ -0,0 +1,252 @@
...
+
+  /**
+   * Main constructor for {@link Model} that takes a {@link TreeElement} as its root.
+   */
+  Model(TreeElement model) {
+    this.model = model;
+  }
+
+  /**
+   * Method that lets iterate and (flat)map over all the children {@link Model} members of
```
dcbriccetti: Hey, let’s iterate! :-)
I suggest,
~~Method that lets~~ Iterates and (flat)maps over ...

src/org/opendatakit/briefcase/export/Model.java
```diff
@@ -0,0 +1,252 @@
...
+   * @return a {@link List} of type T
+   */
+  <T> List<T> flatMap(Function<Model, List<T>> mapper) {
+    return children().stream()
+        .flatMap(child -> mapper.apply(child).stream())
+        .collect(toList());
+  }
+
+  /**
+   * Method that lets iterate over all the children {@link Model} members of
```
dcbriccetti: ~~Method that lets iterate~~ Iterates over

src/org/opendatakit/briefcase/export/XmlElement.java
```diff
@@ -0,0 +1,306 @@
...
+    List<String> names = Arrays.asList(namesArray);
+    return findElement(names.get(0))
+        .map(e -> e.findElements(names.subList(1, names.size()).toArray(new String[]{})))
+        .orElse(Collections.emptyList());
+  }
+
+  /**
+   * Returns the value inside this element.
+   * <p>
+   * If no value is found, this method will throw.
```
dcbriccetti: Throw requires an object. I suggest “heave” or “upchuck”, which don’t. Heh heh.

src/org/opendatakit/briefcase/export/SubmissionMetaData.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.Optional;
+import org.opendatakit.briefcase.model.ParsingException;
+import org.opendatakit.briefcase.reused.Optionals;
+
+/**
+ * This class holds a form's metadata. Instances of this class are
+ * generated while parsing submissions and they are used to read
+ * each value only once.
+ * <p>
+ * All its members are lazily evaluated to avoid unnecessary parsing
```
dcbriccetti: .

src/org/opendatakit/briefcase/export/XmlElement.java
dcbriccetti: You should say what it will throw. Throw requires an object.

dcbriccetti:
I don’t understand why we’d want to keep the old export process.

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -19,6 +19,9 @@
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
 import static org.opendatakit.briefcase.export.ExportForms.buildCustomConfPrefix;
+import static org.opendatakit.briefcase.export.ExportOutcome.ALL_EXPORTED;
```
dcbriccetti: This is one of the special cases where I’d argue for the *. We’re using all three.

dcbriccetti:
Not part of this PR, but something you added in February. Do you get these helpful notices from IDEA? I usually resolve this one. Fewer fields is better.
screen shot 2018-05-12 at 4 09 13 pm

src/org/opendatakit/briefcase/export/ExportToCsv.java
```diff
@@ -0,0 +1,153 @@
...
+      EventBus.publish(ExportEvent.successForm(formDef, (int) exportTracker.total));
+
+    if (exportOutcome == SOME_SKIPPED)
+      EventBus.publish(ExportEvent.partialSuccessForm(formDef, (int) exportTracker.exported, (int) exportTracker.total));
+
+    if (exportOutcome == ALL_SKIPPED)
+      EventBus.publish(ExportEvent.failure(formDef, "All submissions have been skipped"));
+
+    return exportOutcome;
+  }
```
dcbriccetti: If the above is your personal preference, fine, but you might consider 
```
switch (exportOutcome) {
  case ALL_EXPORTED:
    EventBus.publish(ExportEvent.successForm(formDef, (int) exportTracker.total));
    break;
  case SOME_SKIPPED:
    EventBus.publish(ExportEvent.partialSuccessForm(formDef, (int) exportTracker.exported, (int) exportTracker.total));
    break;
  case ALL_SKIPPED:
    EventBus.publish(ExportEvent.failure(formDef, "All submissions have been skipped"));
    break;
}
```
or even
```
Object event = null;
switch (exportOutcome) {
  case ALL_EXPORTED:
    event = ExportEvent.successForm(formDef, (int) exportTracker.total);
    break;
  case SOME_SKIPPED:
    event = ExportEvent.partialSuccessForm(formDef, (int) exportTracker.exported, (int) exportTracker.total);
    break;
  case ALL_SKIPPED:
    event = ExportEvent.failure(formDef, "All submissions have been skipped");
    break;
}
EventBus.publish(event);
```

src/org/opendatakit/briefcase/export/CipherFactory.java
```diff
@@ -0,0 +1,119 @@
...
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import org.apache.commons.codec.binary.Base64;
+import org.opendatakit.briefcase.model.CryptoException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Handles the initialization of Cipher objects for the decryption of submission
```
dcbriccetti: `{@link Cipher}`, to be consistent with the `@link`s in doc comments below.

src/org/opendatakit/briefcase/export/CipherFactory.java
```diff
@@ -0,0 +1,119 @@
...
+  /**
+   * Return the next {@link Cipher} instance. This method has side-effects and will
+   * change the initialization vector, which will affect the next call to this method.
+   *
+   * @throws CryptoException
+   */
+  Cipher next() {
+    try {
+      ++ivSeedArray[ivCounter % ivSeedArray.length];
+      ++ivCounter;
```
dcbriccetti: I’m assuming you didn’t write
`++ivSeedArray[ivCounter++ % ...` because you thought it would be too hard to understand. Some don’t know the difference between pre- and post-increment?

yanokwa:
@dcbriccetti The idea behind keeping the old export was for safety and give people a chance to use the other export if they were getting unexpected results. I guess they can always use the old binary...

dcbriccetti:
I see. I suggest we remove it. It complicates the review (a little), and means it has to be tested and maintained (and maybe even deprecated) going forward.

src/org/opendatakit/briefcase/export/CipherFactory.java
```diff
@@ -0,0 +1,119 @@
...
+ * and media files.
+ */
+final class CipherFactory {
+  private static final Logger log = LoggerFactory.getLogger(CipherFactory.class);
+  private static final int IV_BYTE_LENGTH = 16;
+  private final SecretKeySpec symmetricKey;
+  private final byte[] ivSeedArray;
+  private int ivCounter = 0;
+
+  private CipherFactory(String instanceId, byte[] symmetricKeyBytes) throws CryptoException {
```
dcbriccetti: I don’t see consistency in when you use the `throws CryptoException` clause among the methods in this class with throw it.

src/org/opendatakit/briefcase/export/CipherFactory.java
```diff
@@ -0,0 +1,119 @@
...
+      log.error(msg, e);
+      throw new CryptoException(msg + " Cause: " + e);
+    }
+  }
+
+  /**
+   * Factory that initializes a new {@link CipherFactory} for the given instance ID,
+   * encryption key and private key values.
+   *
+   * @throws CryptoException
```
dcbriccetti: These are missing the description (under what circumstances the exception would be thrown).
https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#throws

src/org/opendatakit/briefcase/export/CipherFactory.java
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Handles the initialization of Cipher objects for the decryption of submission
+ * and media files.
+ */
+final class CipherFactory {
+  private static final Logger log = LoggerFactory.getLogger(CipherFactory.class);
+  private static final int IV_BYTE_LENGTH = 16;
+  private final SecretKeySpec symmetricKey;
+  private final byte[] ivSeedArray;
+  private int ivCounter = 0;
```
dcbriccetti: It’s not clear what “iv” is

src/org/opendatakit/briefcase/export/CipherFactory.java
```diff
@@ -0,0 +1,119 @@
...
+    symmetricKey = new SecretKeySpec(symmetricKeyBytes, "AES/CFB/PKCS5Padding");
+    try {
+      MessageDigest md = MessageDigest.getInstance("MD5");
+      md.update(instanceId.getBytes("UTF-8"));
+      md.update(symmetricKeyBytes);
+      byte[] messageDigest = md.digest();
+      ivSeedArray = new byte[IV_BYTE_LENGTH];
+      for (int i = 0; i < IV_BYTE_LENGTH; ++i) {
+        ivSeedArray[i] = messageDigest[(i % messageDigest.length)];
+      }
```
dcbriccetti: Might be good to explain why we create the seed array this way.

src/org/opendatakit/briefcase/export/CipherFactory.java
```diff
@@ -0,0 +1,119 @@
...
+  private int ivCounter = 0;
+
+  private CipherFactory(String instanceId, byte[] symmetricKeyBytes) throws CryptoException {
+    symmetricKey = new SecretKeySpec(symmetricKeyBytes, "AES/CFB/PKCS5Padding");
+    try {
+      MessageDigest md = MessageDigest.getInstance("MD5");
+      md.update(instanceId.getBytes("UTF-8"));
+      md.update(symmetricKeyBytes);
+      byte[] messageDigest = md.digest();
+      ivSeedArray = new byte[IV_BYTE_LENGTH];
```
dcbriccetti: You may have explained before why you do it this way, but why not initialize in the field declaration?

src/org/opendatakit/briefcase/export/CsvMapper.java
```diff
@@ -0,0 +1,334 @@
...
+    String[] fields = Optional.ofNullable(element.getValue())
+        .map(v -> v.split(" "))
+        .orElseThrow(() -> new BriefcaseException("No value on geopoint field"));
+
+    List<Pair<String, String>> values = new ArrayList<>();
+    values.add(Pair.of(element.fqn() + "-Latitude", Optional.ofNullable(fields[0]).orElse(null)));
+    values.add(Pair.of(element.fqn() + "-Longitude", Optional.ofNullable(fields[1]).orElse(null)));
+    values.add(Pair.of(element.fqn() + "-Altitude", Optional.ofNullable(fields[2]).orElse(null)));
+    values.add(Pair.of(element.fqn() + "-Accuracy", Optional.ofNullable(fields[3]).orElse(null)));
+    return values;
```
dcbriccetti: I didn’t like the duplication here so I tried my hand at Java 8 streams. Would you consider something like this?
```
    String[] tags = new String[] {"Latitude", "Longitude", "Altitude", "Accuracy"};
    return IntStream.range(0, tags.length).mapToObj(i ->
        Pair.of(element.fqn() + "-" + tags[i], Optional.ofNullable(fields[i]).orElse(null))
    ).collect(Collectors.toList());
```
screen shot 2018-05-12 at 5 33 41 pm

src/org/opendatakit/briefcase/export/DataType.java
```diff
@@ -0,0 +1,60 @@
...
+ * the License.
+ */
+package org.opendatakit.briefcase.export;
+
+import java.util.stream.Stream;
+
+/**
+ * This enum encodes JavaRosa data types as described in {@link org.javarosa.core.model.Constants}
+ */
+enum DataType {
```
dcbriccetti: I think we should declare this enum in JavaRosa, even if we don’t convert JR code to use the enum right away. I’ll make such a PR if I don’t hear disagreement.

src/org/opendatakit/briefcase/export/ExportEvent.java
```diff
@@ -0,0 +1,53 @@
...
+    String statusLine = String.format("Exported %s%% of the submissions", base100Percentage);
+    return new ExportEvent(form.getFormId(), statusLine);
+  }
+
+  public static ExportEvent start(FormDefinition form) {
+    return new ExportEvent(form.getFormId(), "Export operation has started");
+  }
+
+  public static ExportEvent end(FormDefinition form, long exported) {
+    return new ExportEvent(form.getFormId(), String.format("Export operation has ended", exported));
```
dcbriccetti: Gotta love IDEA.
screen shot 2018-05-12 at 9 15 09 pm

src/org/opendatakit/briefcase/reused/UncheckedFiles.java
```diff
@@ -0,0 +1,271 @@
...
+    try {
+      return FileUtils.checksumCRC32(file.toFile());
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    }
+  }
+
+  public static boolean isInstanceDir(Path dir) {
+    return Files.isDirectory(dir)
+        && dir.getFileName().toString().matches("^uuid[\\w]{8}-[\\w]{4}-[\\w]{4}-[\\w]{4}-[\\w]{12}$")
```
dcbriccetti: This is a little scary. Will it break if the naming convention changes? Should this test reside closer to where the convention is implemented?
It would be helpful to at least add a comment with an example of an instance directory so readers can match it up with the regex if they want to.

src/org/opendatakit/briefcase/reused/UncheckedFiles.java
```diff
@@ -0,0 +1,271 @@
...
+        && Files.exists(dir.resolve("submission.xml"));
+  }
+
+  public static boolean isSubmissionFile(Path path) {
```
dcbriccetti: Is this method called?

src/org/opendatakit/briefcase/export/FieldMapper.java
```diff
@@ -0,0 +1,63 @@
...
+   * @return a {@link FieldMapper} instance
+   * @see FieldMapper#simpleMapper(Function, int)
+   */
+  static FieldMapper simpleMapper(Function<XmlElement, List<Pair<String, String>>> mapper) {
+    return simpleMapper(mapper, 1);
+  }
+
+  /**
+   * {@link FieldMapper} factory that will just apply some concrete {@link XmlElement} mapper
+   * assuming that there is a fixed amount of key-value pairs to be returned.
```
dcbriccetti: Sometimes the amount/number distinction is not worth fussing about, but here I think you should say “fixed number” instead.
https://www.google.com/search?q=amount+number

src/org/opendatakit/briefcase/export/CsvMapper.java
```diff
@@ -0,0 +1,334 @@
...
+
+  @SuppressWarnings("checkstyle:OverloadMethodsDeclarationOrder")
+  private static List<Pair<String, String>> empty(String fqn) {
+    return empty(fqn, 1);
+  }
+
+  private static List<Pair<String, String>> formattedDate(XmlElement element, DateFormat formatter) {
+    String value = element.getValue();
+    return singletonList(Pair.of(element.fqn(), value.isEmpty() ? null : formatter.format(parseDate(value))));
+  }
```
dcbriccetti: This fails when the date tag has no body. Here’s a guess about fixing it, but maybe you have some better chaining construct, or a different solution altogether.
screen shot 2018-05-12 at 10 01 07 pm

dcbriccetti: Oh, I see for binary you do this:
```
    if (!element.hasValue())
      return empty(element.fqn());
```

src/org/opendatakit/briefcase/export/CsvMapper.java
```diff
@@ -0,0 +1,334 @@
...
+
+  private static List<Pair<String, String>> dateTime(XmlElement element) {
+    return formattedDate(element, getDateTimeInstance());
+  }
+
+  private static List<Pair<String, String>> geopoint(XmlElement element) {
+    String[] fields = Optional.ofNullable(element.getValue())
+        .map(v -> v.split(" "))
+        .orElseThrow(() -> new BriefcaseException("No value on geopoint field"));
```
dcbriccetti: This also throws an exception if element is empty, presumably not completing the export.

src/org/opendatakit/briefcase/export/ExportEvent.java
```diff
@@ -0,0 +1,65 @@
...
+public final class ExportEvent {
+  private final String formId;
+  private final String statusLine;
+
+  private ExportEvent(String formId, String statusLine) {
+    this.formId = formId;
+    this.statusLine = statusLine;
+  }
+
+  public static ExportEvent progress(FormDefinition form, double percentage) {
```
dcbriccetti: I recommend some changes here to improve the presentation of these messages.

screen shot 2018-05-13 at 10 41 48 am

Let’s pluralize correctly. 

“Export operation” is longer and no better than “Export”.  Same for “Successfully exported” and “exported”, and the latter allows the whole message to fit in the GUI on my computer, as you can see below.

screen shot 2018-05-13 at 10 40 40 am

I’m confused by the “Not exported yet”.

dcbriccetti:
Here’s my branch off your branch where I’ve tried out the changes I am recommending. Help yourself. (Just wondering what “help yourself” sounds like to people not familiar with the idiom.)

src/org/opendatakit/briefcase/export/SubmissionMetaData.java
dcbriccetti: This works better:
```
  Optional<OffsetDateTime> getSubmissionDate() {
    if (submissionDate == null)
      submissionDate = root.getAttributeValue("submissionDate")
          .map(SubmissionMetaData::regularizeDateTime)
          .map(OffsetDateTime::parse);
    return submissionDate;
  }

  /** Fixes ISO8601-ish strings not in this form: 2018-05-13T17:32:57±00:00 */
  private static String regularizeDateTime(String iso8601DateTime) {
    // 2018-04-26T08:58:20.525Z
    if (iso8601DateTime.endsWith("Z")) // Replace the Z with a zero offset
      return iso8601DateTime.substring(0, iso8601DateTime.length() - 1) + "+00:00";

    // 2018-05-13T17:32:57+00  Add the minutes if not present
    return iso8601DateTime + (iso8601DateTime.length() - iso8601DateTime.lastIndexOf(":") == 3 ? "" : ":00");
  }
```

src/org/opendatakit/briefcase/export/CipherFactory.java
```diff
@@ -0,0 +1,119 @@
...
+   * @throws CryptoException
+   */
+  static CipherFactory from(String instanceId, String base64EncryptedKey, PrivateKey privateKey) {
+    try {
+      Cipher pkCipher;
+      pkCipher = Cipher.getInstance("RSA/NONE/OAEPWithSHA256AndMGF1Padding");
+      pkCipher.init(Cipher.DECRYPT_MODE, privateKey);
+      byte[] encryptedSymmetricKey = Base64.decodeBase64(base64EncryptedKey);
+      byte[] decryptedKey = pkCipher.doFinal(encryptedSymmetricKey);
+      return new CipherFactory(instanceId, decryptedKey);
```
dcbriccetti: 11:38:52.334 [main] ERROR org.opendatakit.briefcase.util.FileSystemUtils - Error decrypting:1524644764247.jpg.enc
java.security.InvalidKeyException: Illegal key size
⋮
	at org.opendatakit.briefcase.util.CipherFactory.getCipher(CipherFactory.java:78)
	at org.opendatakit.briefcase.util.EncryptionInformation.getCipher(EncryptionInformation.java:59)
	at org.opendatakit.briefcase.util.FileSystemUtils.decryptFile(FileSystemUtils.java:455)

test/java/org/opendatakit/briefcase/export/ExportToCsvScenario.java
```diff
@@ -0,0 +1,295 @@
...
+          createDirectories(submissionDir);
+
+          // We will copy every file we find that is prefixed with the
+          // name of the form we're installing, ignoring PEM files and templates
+          walk(path.getParent())
+              .filter(file -> file.getFileName().toString().startsWith(formName + "_"))
+              .filter(file -> !file.getFileName().toString().endsWith(".pem"))
+              .filter(file -> !file.getFileName().toString().endsWith(".tpl"))
+              .forEach(file -> {
+                Path target = submissionDir.resolve(file.getFileName().toString().substring(formName.length() + 1));
```
dcbriccetti: I find this more readable. You?
```
.filter(file -> {
  String name = file.getFileName().toString();
  return name.startsWith(formName + "_") && !name.endsWith(".pem") && !name.endsWith(".tpl");
}).forEach(file -> {
```

test/java/org/opendatakit/briefcase/export/ExportToCsvScenario.java
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  private static Optional<String> readInstanceId(String formName) {
+    return maybeGetPath(formName + "_submission.xml")
+        .filter(Files::exists)
+        .map(ExportToCsvScenario::readInstanceId);
+  }
+
+  private static String readInstanceId(Path path) {
+    return readAllLines(path).get(0).split("instanceID=\"")[1].split("\"")[0];
```
dcbriccetti: Keep in mind that some forms are gigantic. Why not just read the first line?
```
    return readOneLine(path).split("instanceID=\"")[1].split("\"")[0];
```
screen shot 2018-05-13 at 1 03 45 pm

src/org/opendatakit/briefcase/export/SubmissionParser.java
```diff
@@ -0,0 +1,187 @@
...
+    Path instancesDir = formDir.resolve("instances");
+    if (!Files.exists(instancesDir) || !Files.isReadable(instancesDir))
+      return Stream.empty();
+
+    return list(instancesDir).parallel()
+        // Normally, we would use the Streams API to filter out files that
+        // are not contained in the given date range, but parsing is expensive
+        // enough to make it all in one pass and use Optionals to account
+        // for that
+        .map(path -> parseSubmission(path.resolve("submission.xml"), dateRange, isEncrypted, privateKey))
```
dcbriccetti: I had an error here:

java.io.UncheckedIOException: java.nio.file.FileSystemException: /private/tmp/ODK Briefcase Storage/forms/Independent Monitoring/instances/.DS_Store/submission.xml: Not a directory

Here is one solution:
```
        // for that
        .filter(path -> !path.toFile().isHidden())
        .map(path -> parseSubmission(path.resolve("submission.xml"), dateRange, isEncrypted, privateKey))
```

dcbriccetti:
Regarding performance, I can send you a YourKit snapshot if you like. We have a license, if you weren’t aware. Here’s my little system:

  Model Name:	MacBook Pro
  Model Identifier:	MacBookPro11,3
  Processor Name:	Intel Core i7
  Processor Speed:	2.8 GHz
  Number of Processors:	1
  Total Number of Cores:	4
  L2 Cache (per Core):	256 KB
  L3 Cache:	6 MB
  Memory:	16 GB

ggalmazor:
Thanks, @dcbriccetti, it's an impressive review! I'm trying to limit my brain bandwidth to one hard task at a time and I'm currently finishing some other stuff. I hope I can retake this PR during the next days and process your review.

src/org/opendatakit/briefcase/ui/export/ExportPanel.java
ggalmazor: Yup. That would improve the code...

src/org/opendatakit/briefcase/export/CipherFactory.java
ggalmazor: That's a mistake! Fixing it! :)

src/org/opendatakit/briefcase/export/CipherFactory.java
ggalmazor: This is just a copy of `org.opendatakit.briefcase.util.CipherFactory`. I honestly don't know why encryption is dealt with this way... I'll rescue some comments from the original, just in case.

src/org/opendatakit/briefcase/export/CipherFactory.java
ggalmazor: That would help a lot, but I honestly don't know why this is done that way :P

src/org/opendatakit/briefcase/export/CipherFactory.java
ggalmazor: Can you confirm that you're using Java 8u161 or greater?

src/org/opendatakit/briefcase/export/CipherFactory.java
ggalmazor: Most of the code in this class is a mystery to me... :/

src/org/opendatakit/briefcase/export/CsvMapper.java
ggalmazor: I see that it's not consistent how I deal with empty elements through the code at this point... Let's discuss how it should behave when dealing with empty elements and then I'll change this commit.

We need to discuss these data types:

| Data type | Expected output |
| --- | --- |
| text (generic field) | an empty string |
| date & time types | 1 empty string |
| geopoint | 4 empty strings |
| binary | 1 empty string |
| null (non repeating group) | should throw? |
| null (repeating group) | should throw? |

src/org/opendatakit/briefcase/export/DataType.java
ggalmazor: That's great, thanks!

ggalmazor: Has the a new version of JR including this enum been released yet?

src/org/opendatakit/briefcase/export/ExportToCsv.java
ggalmazor: But my if expression takes also into account the number of processed submissions... How can we express that in a `switch`?

src/org/opendatakit/briefcase/export/DataType.java
lognaturel: Yes! Still in snapshot but you can use it. The version is 2.10.0-SNAPHOT

src/org/opendatakit/briefcase/export/SubmissionParser.java
ggalmazor: Something doesn't seem right... Could you copy the full trace here, please?

I don't understand how you're getting that error there and what has to do with `.DS_Store` beind hidden...

src/org/opendatakit/briefcase/reused/UncheckedFiles.java
ggalmazor: > This is a little scary. Will it break if the naming convention changes? Should this test reside closer to where the convention is implemented?

That's a very good point! I'll add a comment and we can open an issue to fix this design flaw after we get this merged. This affects the pulling process where these directories are created

test/java/org/opendatakit/briefcase/export/ExportToCsvScenario.java
ggalmazor: I'd prefer to leave them separate. I'll give those predicates a name to see if it's more clear that way.

src/org/opendatakit/briefcase/export/DataType.java
ggalmazor: Ugh... possibly newbie question here, sorry...

I'm getting an error after updating the JR version in the build file:
```
   > Could not find org.opendatakit:opendatakit-javarosa:2.10.0-SNAPHOT.
     Searched in the following locations:
         https://repo.maven.apache.org/maven2/org/opendatakit/opendatakit-javarosa/2.10.0-SNAPHOT/opendatakit-javarosa-2.10.0-SNAPHOT.pom
         https://repo.maven.apache.org/maven2/org/opendatakit/opendatakit-javarosa/2.10.0-SNAPHOT/opendatakit-javarosa-2.10.0-SNAPHOT.jar
         https://oss.sonatype.org/content/groups/public/org/opendatakit/opendatakit-javarosa/2.10.0-SNAPHOT/opendatakit-javarosa-2.10.0-SNAPHOT.pom
         https://oss.sonatype.org/content/groups/public/org/opendatakit/opendatakit-javarosa/2.10.0-SNAPHOT/opendatakit-javarosa-2.10.0-SNAPHOT.jar
```

src/org/opendatakit/briefcase/export/ExportProcessTracker.java
```diff
@@ -0,0 +1,52 @@
...
+    this.total = total;
+  }
+
+  void incAndReport() {
+    this.exported++;
+    Integer percentage = (int) (exported * 100 / total);
+    if (percentage > lastReportedPercentage && percentage % 10 == 0) {
+      EventBus.publish(ExportEvent.progress(form, percentage.doubleValue() / 100));
+      lastReportedPercentage = percentage;
+    }
```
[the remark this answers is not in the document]
ggalmazor: I think it won't produce the same values because dividing `int / double` produces an `int` by truncating decimals.

ggalmazor:
> (About the old implementation) I see. I suggest we remove it. It complicates the review (a little), and would have to be tested and maintained (and maybe even deprecated) going forward.

The idea is to remove it after some time (maybe a pair of releases) after the new one is a little more battle tested.

src/org/opendatakit/briefcase/export/DataType.java
lognaturel: 🤔Really? I just tried 
```
   compile(group: 'org.opendatakit', name: 'opendatakit-javarosa', version: '2.10.0-SNAPSHOT') {
        exclude group: 'org.slf4j'
    }
```
and it worked. Maybe some kind of caching? Clean and build again?

You will need to exclude slf4j but that's not what the error was about.

ggalmazor: Thanks! That did the trick :)

Why exclude slf4j, though?

src/org/opendatakit/briefcase/export/ExportProcessTracker.java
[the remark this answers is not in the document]
dcbriccetti: scala> 5 / 2
res0: Int = 2

scala> 5 / 2D
res1: Double = 2.5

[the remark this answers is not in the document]
ggalmazor: Ooh, alright!

dcbriccetti:
>The idea is to remove it after some time (maybe a pair of releases) after the new one is a little more battle tested.

As Yaw said, users can change to the previous version of Briefcase if they have problems.

I don’t care very much about this.

ggalmazor:
I don't have a strong opinion about this. If @yanokwa changes his opinion we can remove the old one.

src/org/opendatakit/briefcase/export/DataType.java
lognaturel: I believe that if the versions in JR and Briefcase get out of sync the build will fail if it's included transitively from JR. They match now so I suppose we can leave it in and make sure to keep the versions in sync. I'd tend to exclude it so we don't have to deal with cryptic errors in the future.

dcbriccetti:
>I don't have a strong opinion about this. If @yanokwa changes his opinion we can remove the old one.

You mean if @yanokwa  _maintains_ his position? He seemed to agree that including both wasn’t needed.

yanokwa:
I discussed the removal with @lognaturel and she doesn't feel strongly about it. It is very easy to roll back and this saves us some maintenance, so let's do what @dcbriccetti suggested and remove the crutch, @ggalmazor!

ggalmazor:
OK, now there no place outside the test where the old implementation is used.

Next, I'll record the expected outputs to convert the GoldenMaster tests in place into QA tests that don't exercise the old implementation.

ggalmazor:
I've converted the tests but now I'm having some problems with time zones... I'll continue tomorrow :P

ggalmazor:
@yanokwa, @dcbriccetti, I think it's ready now.

src/org/opendatakit/briefcase/export/ExportToCsv.java
dcbriccetti: The switch on the enum with no default makes it explicit that we are addressing all three mutually exclusive conditions. For what that’s worth.

src/org/opendatakit/briefcase/export/CipherFactory.java
dcbriccetti: Sorry, I thought you wrote that.

src/org/opendatakit/briefcase/export/CipherFactory.java
dcbriccetti: Let’s forget that, then. And anywhere else I thought you wrote it and then I asked for something.

src/org/opendatakit/briefcase/export/CsvMapper.java
dcbriccetti: Thanks for that, and for leading us into the past future with Java 8 functional features

src/org/opendatakit/briefcase/export/CipherFactory.java
dcbriccetti: JRE: 1.8.0_152-release-1136-b20 x86_64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
macOS 10.12.6

test/java/org/opendatakit/briefcase/export/ExportToCsvScenario.java
dcbriccetti: OK

src/org/opendatakit/briefcase/export/SubmissionParser.java
dcbriccetti: Browse your data with Finder (are you using MacOS?) and it will create the .DS_Store directories.

src/org/opendatakit/briefcase/export/Model.java
```diff
@@ -0,0 +1,253 @@
...
+
+  /**
+   * Main constructor for {@link Model} that takes a {@link TreeElement} as its root.
+   */
+  Model(TreeElement model) {
+    this.model = model;
+  }
+
+  /**
+   * Iterates the children of this instance and returns the flatmapped result of mapping
```
dcbriccetti: I think it needs to be “iterate over the children”.
https://english.stackexchange.com/questions/23023/iterate-vs-iterate-over

src/org/opendatakit/briefcase/export/XmlElement.java
dcbriccetti: >will throw an exception

Touché!

src/org/opendatakit/briefcase/export/ExportToCsv.java
dcbriccetti: Using else/if’s would get you partway there. 
I don’t care much about this one.

src/org/opendatakit/briefcase/export/CsvMapper.java
dcbriccetti: I don’t know. Can you just do what the previous export did?

test/java/org/opendatakit/briefcase/export/ExportToCsvScenario.java
dcbriccetti: I like that less. The separate function is cool, but why not pass the extension in as an argument?

src/org/opendatakit/briefcase/reused/UncheckedFiles.java
dcbriccetti: That’s a good comment.

dcbriccetti:
I’m glad we agreed on so many of my suggested changes. What’s left to resolve?
- empty tags
- key error
- not exported yet
- hidden folders

ggalmazor:
Thanks, @dcbriccetti ! It's a lifesaver to have you onboard for these huge changes :)

**empty tags**

I'll throw in a new test focused on testing this to verify it works as expected.

**key error**

My best bet for this one is that you're running Briefcase with a JVM older than 1.8.0_161. Can you check that? Previous versions need additional config to work with the cryptographic library we're using.

I can recommend http://sdkman.io to manage JVM and other SDK versions in your computer.

**not exported yet**

This is used when the form hasn't been exported by the user yet. After the first export, the last export date is shown there.

Is there something you would change here?

**hidden folders**

I still don't understand how that trace is related to having that hidden folder. Could you check the contents of that `submissions.xml` file? is that actually a valid submission?

Anyway, I'll add a filter that excludes any directory that doesn't look like a legit instance directory.

ggalmazor:
Regarding empty tags, this is what the old implementation will produce:

| Type               | Output            | Example                   | 
| ------------------ | ----------------- | ------------------------- |
| string             | `""`              | `something,"",something`  |
| int                | `""`              | `something,"",something`  |
| decimal            | `""`              | `something,"",something`  |
| date               | empty column      | `something,,something`    |
| time               | empty column      | `something,,something`    |
| date_time          | empty column      | `something,,something`    |
| geopoint           | 4 x empty columns | `something,,,,,something` |
| geotrace           | `""`              | `something,"",something`  |
| geoshape           | `""`              | `something,"",something`  |
| barcode            | `""`              | `something,"",something`  |

Does this seem right to you, @lognaturel, @yanokwa, @dcbriccetti? I think numeric types should be encoded without double quotes...

ggalmazor:
I've fixed how we handle empty tags. Both `<field></field>` and `<field/>` should work as defined on the table from my last comment.

I've also added the filter that will process only submissions inside correct instance folders.

src/org/opendatakit/briefcase/export/ExportEvent.java
dcbriccetti: Both of these statements can’t be true:
screen shot 2018-05-17 at 1 41 11 pm

dcbriccetti:
>I still don't understand how that trace is related to having that hidden folder. 

map(path -> parseSubmission(path.resolve("submission.xml"), dateRange, isEncrypted, privateKey))

You were trying to open "submission.xml" in every directory in `instances`. There is no such file in the .DS_store directories.
```
dbl:/tmp/ODK Briefcase Storage/forms/All widgets/instances $ ls -a
.                                        .DS_Store
..                                       uuide8042ecc-9966-4d6f-9db3-2a5b2af3d96d
```

You’ve solved it with your new filter.

src/org/opendatakit/briefcase/export/CsvMapper.java
```diff
@@ -0,0 +1,352 @@
...
+    switch (dataType) {
+      case GEOPOINT:
+      case DATE:
+      case TIME:
+      case DATE_TIME:
+        return true;
+      default:
+        return false;
+    }
+  }
```
dcbriccetti: I find this much simpler. You?

screen shot 2018-05-17 at 2 17 58 pm

ggalmazor:
> You were trying to open "submission.xml" in every directory in instances. There is no such file in the .DS_store directories.

But the trace said "Not a directory"... I agree that now it's solved... Let's move on :P

dcbriccetti:
>But the trace said "Not a directory"... I agree that now it's solved... Let's move on :P

That puzzled me, too. I think the error message is misleading.

dcbriccetti:
Remaining items:

me: key error
@yanokwa @lognaturel: verify correct handling of nulls
@ggalmazor: contradictory “not exported yet”

What else?

src/org/opendatakit/briefcase/export/CipherFactory.java
ggalmazor: OK, with that JVM you should follow these complicated and boring instructions here: https://docs.opendatakit.org/encrypted-forms/#configuration

**OR**

Install JVM 1.8.0u161 or a greater version :)

ggalmazor:
@dcbriccetti, @yanokwa, @lognaturel, I think there is nothing pending to be changed now.

We could discuss some incoherence I've found on how the old implementation dealt with empty tags - see https://github.com/opendatakit/briefcase/pull/457#issuecomment-389788696 - but, at least, we can be sure that the new implementation does just the same. We could change that on further PRs if we want.

yanokwa:
Agreed with @ggalmazor that we should file an issue for the empty values and discuss that there. Matching the old implementation makes it good enough to test.

@ggalmazor/dcbriccetti can either of you explain the issue with the contradictory 'not exported yet'. Explain Like I'm Five.

ggalmazor:
> can either of you explain the issue with the contradictory 'not exported yet'. Explain Like I'm Five.

@dcbriccetti detected a problem on the UI: the column for "last export date" wasn't being updated after exporting a form, therefore, the status was telling "Successfully exported" and the last export date was "Not exported yet", which is contradictory.

What was happening was that I changed some code related to events and subscriptions and forgot to adapt one method in `ExportForms.appendStatus()` responsible for updating the last export date when we detected that a form had been successfully exported.

Now that the method works correctly, the last export date is updated and the UI doesn't show contradictory information anymore.

ggalmazor:
Thanks for all your feedback, @yanokwa, and @dcbriccetti! I know that reviewing these monster PRs requires a huge effort.

kkrawczyk123:
My efficiency results are not very impressive. On my ubuntu machine export of 50k submissions takes 1 minute, I see “Export has started” immediately, then % of progress in Export Status column and then amount of exported submissions. (I have 16GB of ram memory there). The same activity looks terrible on my Mac mini (8gb of ram) and takes like forever on my Windows VM (4gb of ram). I wasn’t even able to notice any % status of progress, and on Windows it took like 5 minutes for “Export has started” to show up. And it started, started and froze my briefcase ;( so after 30 minutes of waiting export still didn’t finish. I have some file in my export directory after nearly an hour I have 31800 submissions in that file.

I have noticed that progress bar is visible only for few seconds then it disappears (On all three systems right after Export has started appears in Export status on all 3 systems and when bulk export also) Is this an intended behavior? 

I have already mentioned it when progress bar was added when you do not add pem key file location when export encrypted form error message shows, export doesn’t start but the progress bar is active and it doesn’t want to stop to dance. (all three systems) 

yanokwa:
@kkrawczyk123 when it froze on Windows, did you get any error messages? Does the behavior change if you try the export on the CLI?

ggalmazor:
I know this is a pain to test, but it would be great to know how the master branch deals with the same amount of data.

ggalmazor:
Also, the expected behavior would be that the UI doesn't block at any time, the progress bar is active all the time and we get progress update every 10%. 

For long exports like these, it could be better to get progress updates each 1%, maybe.

yanokwa:
@kkrawczyk123 is it easy for you to try the efficiency test on master, so we can compare?

ggalmazor:
> I have already mentioned it when progress bar was added when you do not add pem key file location when export encrypted form error message shows, export doesn’t start but the progress bar is active and it doesn’t want to stop to dance. (all three systems)

That's a bug! I'm on it :)

ggalmazor:
@kkrawczyk123, I've fixed the error with the PEM file you have reported. I've also reviewed and improved how the UI locks itself while exporting to prevent user interaction while an export is happening.

I've also made some small changes to how the export process is launched which could improve the situation with lots of submissions.

kkrawczyk123:
## Branch vs. Master

### Round 1st:
Export of 50k submissions.

**Branch Results:**
_Ubuntu:_ less than 1 minute
_MacOs_: after 17 minutes Unexpected error, exported submissions 40k
screenshot from 2018-05-29 12-01-18

_Windows:_ the same error message as on MacOS after 15 minutes, exported submissions: 29k
screenshot from 2018-05-29 10-18-14

**Master Results:**
_Ubuntu:_ the same results as on the branch, less than minute
_MacOS:_ success after 8 minutes
_Windows:_ success after 15 mintes

### Round 2nd:
Bulk export of 50k submissions and other forms and submissions.

**Branch Results:**
_Ubuntu_: 1 minute 
_MacOS:_ the same error message after 5 minutes
_Windows_: the same error message after 8 minutes

**Master Results:**
_Ubuntu:_ same as on the branch, like 1 minute
_MacOS:_ success after 8 minutes
_Windows:_ success after 20 minutes
(amount of forms could be different on all platforms but still, success is a success)

### Round 3rd:
Cmd export.

**Branch Results:**
_Ubuntu:_ the same result in time meaning but prefix -f didn't worked
_MacOS:_ process didn't fnished, error in cmd response
_Windows_: process didn't fnished, error in cmd response
Content of cmd response:
10:52:52.822 [main] ERROR org.opendatakit.common.cli.Cli - Error
java.lang.OutOfMemoryError: null
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(Unknown Source)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(Unknown Source)
        at java.lang.reflect.Constructor.newInstance(Unknown Source)
        at java.util.concurrent.ForkJoinTask.getThrowableException(Unknown Source)
        at java.util.concurrent.ForkJoinTask.reportException(Unknown Source)
        at java.util.concurrent.ForkJoinTask.invoke(Unknown Source)
        at java.util.stream.Nodes.collect(Unknown Source)
        at java.util.stream.ReferencePipeline.evaluateToNode(Unknown Source)
        at java.util.stream.AbstractPipeline.evaluate(Unknown Source)
        at java.util.stream.SortedOps$OfRef.opEvaluateParallel(Unknown Source)
        at java.util.stream.AbstractPipeline.opEvaluateParallelLazy(Unknown Source)
        at java.util.stream.AbstractPipeline.sourceSpliterator(Unknown Source)
        at java.util.stream.AbstractPipeline.evaluate(Unknown Source)
        at java.util.stream.ReferencePipeline.forEachOrdered(Unknown Source)
        at org.opendatakit.briefcase.export.ExportToCsv.export(ExportToCsv.java:114)
        at org.opendatakit.briefcase.operations.Export.export(Export.java:88)
        at org.opendatakit.briefcase.operations.Export.lambda$static$2(Export.java:55)
        at org.opendatakit.common.cli.Cli.lambda$run$4(Cli.java:126)
        at java.lang.Iterable.forEach(Unknown Source)
        at org.opendatakit.common.cli.Cli.run(Cli.java:123)
        at org.opendatakit.briefcase.Launcher.main(Launcher.java:79)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
        at java.lang.reflect.Method.invoke(Unknown Source)
        at org.eclipse.jdt.internal.jarinjarloader.JarRsrcLoader.main(JarRsrcLoader.java:58)
Caused by: java.lang.OutOfMemoryError: GC overhead limit exceeded
        at java.util.stream.StreamSupport.stream(Unknown Source)
        at java.util.stream.Stream.concat(Unknown Source)
        at org.opendatakit.briefcase.export.XmlElement.lambda$flatten$6(XmlElement.java:227)
        at org.opendatakit.briefcase.export.XmlElement$$Lambda$120/1529060733.apply(Unknown Source)
        at java.util.stream.ReferencePipeline$7$1.accept(Unknown Source)
        at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(Unknown Source)
        at java.util.stream.AbstractPipeline.copyInto(Unknown Source)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source)
        at java.util.stream.StreamSpliterators$WrappingSpliterator.forEachRemaining(Unknown Source)
        at java.util.stream.Streams$ConcatSpliterator.forEachRemaining(Unknown Source)
        at java.util.stream.ReferencePipeline$Head.forEach(Unknown Source)
        at java.util.stream.ReferencePipeline$7$1.accept(Unknown Source)
        at java.util.ArrayList$ArrayListSpliterator.tryAdvance(Unknown Source)
        at java.util.stream.ReferencePipeline.forEachWithCancel(Unknown Source)
        at java.util.stream.AbstractPipeline.copyIntoWithCancel(Unknown Source)
        at java.util.stream.AbstractPipeline.copyInto(Unknown Source)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source)
        at java.util.stream.FindOps$FindOp.evaluateSequential(Unknown Source)
        at java.util.stream.AbstractPipeline.evaluate(Unknown Source)
        at java.util.stream.ReferencePipeline.findFirst(Unknown Source)
        at org.opendatakit.briefcase.export.XmlElement.findElement(XmlElement.java:113)
        at org.opendatakit.briefcase.export.CsvMapper.lambda$null$12(CsvMapper.java:194)
        at org.opendatakit.briefcase.export.CsvMapper$$Lambda$141/1536031937.accept(Unknown Source)
        at java.util.ArrayList.forEach(Unknown Source)
        at org.opendatakit.briefcase.export.Model.forEach(Model.java:72)
        at org.opendatakit.briefcase.export.CsvMapper.lambda$getRepeatCsvLine$13(CsvMapper.java:190)
        at org.opendatakit.briefcase.export.CsvMapper$$Lambda$139/461160828.accept(Unknown Source)
        at java.util.ArrayList.forEach(Unknown Source)
        at org.opendatakit.briefcase.export.CsvMapper.getRepeatCsvLine(CsvMapper.java:187)
        at org.opendatakit.briefcase.export.ExportToCsv.lambda$null$3(ExportToCsv.java:102)
        at org.opendatakit.briefcase.export.ExportToCsv$$Lambda$131/371439501.accept(Unknown Source)
        at java.util.ArrayList.forEach(Unknown Source)

**Master Results:**
_Ubuntu_: the same result in time meaning and prefix -f worked very well
_MacOS:_ export succeeded in 2 minutes, prefix -f worked very well
_Windows:_ export succeeded in 7 minutes, prefix -f worked very well


I confirm that bug connected with pem key file has been fixed and also that UI look while export is ongoing has been improved :)

kkrawczyk123:
One more UI think that I've noticed and forgot about in the previous comment, on Ubuntu and Windows scroll overlaps a bit details buttons on MacOS everything looks good, screenshots of ubuntu and windows attached: 
exportwin
export

ggalmazor:
Thanks, @kkrawczyk123. This information is very valuable! :)

ggalmazor:
We knew that the new code requires more memory to process everything. In my setup, it was approximately a 125% of the memory that the master branch needs.

It shouldn't take as much as 8GB of memory... Could you maybe close running programs and see how much free memory you have on your mac before starting and try again the new branch version?

kkrawczyk123:
Briefcase was the only running program on Mac when I was testing  - I use it only in this project so I could have opened only browser and some folders but it shouldn't have a significant impact on that.

ggalmazor:
Great :) then I guess we only would have to know how much free memory is left before starting the export.

ggalmazor:
@kkrawczyk123 regarding the scroll problem, I think it should be filed as a separate issue (we haven't changed the UI layout on this PR). Could you create it, please?

kkrawczyk123:
As u wish I will report :)

ggalmazor:
@kkrawczyk123, I've improved the memory usage... Could you test this PR again, please?

ggalmazor:

kkrawczyk123:
I was able to export with success on all 3 systems, results:

### Ubuntu:
**export:** less than 1 minute
**cmd export:** less than 1 minute

### MacOS:
**export:** 11 minutes
**cmd export:** 1 minute

### Windows:
**export:** 22 minutes
**cmd export:** 12 minutes.

The only  one issue that I've noticed is that -f prefix doesn't work on this pr on on 3 systems when exportig using cmd. 

ggalmazor:
Thanks, @kkrawczyk123! I'll fix the `-f` for next monday. I think I removed it by accident since the filename can't be set from the UI.

dcbriccetti:
briefcase in 100 mb

This code now runs with a max heap of only 100 MB. Nice work! There are other optimizations that we could make (not parsing each form twice), but I don’t think they need to be made now.

ggalmazor:
@kkrawczyk123, I've fixed the `-f` cli param :)

kkrawczyk123:
I confirm.

Issue #442, opened 2018-04-05
Improve pushing forms behavior


ggalmazor:
This issue is related to #420 

#### Software versions 
Briefcase v1.10.0

#### Problem description
Pushing forms to Aggregate always sends a blank form, which can be a waste of time and bandwidth with forms with media file attachments.

#### Expected behavior
- [ ] Briefcase should check against the list of forms installed on Aggregate and send the blank form only if it doesn't find it.
- [ ] If a user needs to send it anyway (maybe because there is a newer version of it), Briefcase should provide means to force send it:
  - [ ] On the UI
  - [ ] On the CLI

ggalmazor:
@opendatakit-bot label "needs discussion"

yanokwa:
The expected behavior is clear for everything but the GUI. I'm not even sure we should enable this on the GUI, but I'd be interested to hear how you think it should work.

ggalmazor:
I think we could add a new checkbox to enable "force-pushing" on the form to configure an Aggregate server as the push target with a text that explains that we won't send a blank form unless we don't find it there (we need to explain that anyway).

ggalmazor:
@opendatakit-bot label "contributor friendly"

ggalmazor:
I've updated the issue description to warn about its relation to the ongoing PR #405. It would be better to wait for that PR to be merged before starting to work on this issue.

dcbriccetti:
What’s the command line syntax to do a push?

ggalmazor:
The syntax would be `--push_aggregate` or `-psha` and some required and optional args:
```
Params for -psha operation:
  -id,--form_id <arg>                 Form ID
  -p,--odk_password <arg>             ODK Password
  -sd,--storage_directory <arg>       Briefcase storage directory
  -u,--odk_username <arg>             ODK Username
  -url,--aggregate_url <arg>          Aggregate server URL
Optional params for -psha operation:
  -fsb,--force_send_blank             Force sending the blank form to the Aggregate instance
```

dcbriccetti:
`--push_aggregate --form_id All%20widgets --storage_directory /tmp/ODK%20Briefcase%20Storage`

`2018-05-16 09:39:40,897 [main] ERROR o.o.a.p.BaseFormParserForJavaRosa - Launch Failed: Unrecognized option: --push_aggregate`

ggalmazor:
Hi Dave! It's running OK for me. Some thoughts:
[...]

PR #461, opened 2018-05-08
Issue 442 improve cli push

ggalmazor:
Addresses #442:
- Avoid sending the blank form if Briefcase detects that it's already present in the destination server
- Add a new CLI flag to let users decide if they want to force sending the form anyway

While we implement the rest of #442, which involves changes on the GUI, pushing from it always sends the blank forms.

- Added tests (unit and integration) for the HTTP interaction parts
- Manually tested the CLI and GUI to verify the intended behavior:
  - CLI outputs some message telling if it's sending the blank form or skipping it. Just run the CLI op with and without the `-fsb` flag
  - GUI should add the same lines as the CLI to the push details when used in combination with `-fsb`

Most of the changes in this PR are quite straightforward but I had to choose between using the existing HTTP interaction code or implementing it new.

I've chosen to implement a new `Http` port to avoid some of the problems we have already detected on the current code:
- It has side-effects that can be unpredictable effects, especially when combined with authentication
- It has a low level of abstraction, which can be verbose, and hard to follow and change
- Whatever code based on it is inherently harder to test

None.

Yep: https://github.com/opendatakit/docs/issues/703

### DIFF
--- a/src/org/opendatakit/briefcase/push/RemoteServer.java
+++ b/src/org/opendatakit/briefcase/push/RemoteServer.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C) 2018 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.opendatakit.briefcase.push;
+
+import java.net.URI;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Stream;
+import org.opendatakit.briefcase.reused.http.Credentials;
+import org.opendatakit.briefcase.reused.http.Query;
+
+/**
+ * This class represents a remote Aggregate server and it has some methods
+ * to query its state.
+ */
+public class RemoteServer {
+  private final URI baseUri;
+  private final Optional<Credentials> credentials;
+
+  private RemoteServer(URI baseUri, Optional<Credentials> credentials) {
+    this.baseUri = baseUri;
+    this.credentials = credentials;
+  }
+
+  /**
+   * Factory that creates a new {@link RemoteServer} to be used when it requires
+   * authentication.
+   */
+  public static RemoteServer authenticated(String baseUri, String username, String password) {
+    return new RemoteServer(
+        URI.create(baseUri),
+        Optional.of(new Credentials(username, password))
+    );
+  }
+
+  /**
+   * Factory that creates a new {@link RemoteServer} to be used when it doesn't
+   * require authentication.
+   */
+  public static RemoteServer normal(String baseUri) {
+    return new RemoteServer(URI.create(baseUri), Optional.empty());
+  }
+
+  /**
+   * Returns a new {@link Query} to check if the given form ID is present in the
+   * remote server or not.
+   *
+   * @return true if the form is present on the remote server, false otherwise
+   */
+  public Query<Boolean> containsFormQuery(String formId) {
+    return buildBaseQuery()
+        .resolve("/formList")
+        .map((String body) -> Stream.of(body.split("\n"))
+            .anyMatch(s -> s.contains("<formID>" + formId + "</formID>")));
+  }
+
+  private Query<String> buildBaseQuery() {
+    return new Query<>(baseUri, credentials, Function.identity());
+  }
+}
--- a/src/org/opendatakit/briefcase/reused/http/Credentials.java
+++ b/src/org/opendatakit/briefcase/reused/http/Credentials.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright (C) 2018 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.opendatakit.briefcase.reused.http;
+
+import java.util.Objects;
+
+/**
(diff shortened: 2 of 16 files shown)


test/java/org/opendatakit/briefcase/push/RemoteServerTest.java
```diff
@@ -0,0 +1,59 @@
...
+package org.opendatakit.briefcase.push;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+import org.opendatakit.briefcase.reused.http.FakeHttp;
+import org.opendatakit.briefcase.reused.http.Query;
+
+public class RemoteServerTest {
```
yanokwa: I don't think this test is valid and I think I led you astray with my curl command. You'll need to pass in a "X-OpenRosa-Version: 1.0" header..

`curl --header "X-OpenRosa-Version: 1.0" https://opendatakit.appspot.com/formList`

Your output will look like this:
```xml
<xforms xmlns="http://openrosa.org/xforms/xformsList">
<xform><formID>CascadingSelect</formID>
<name>Cascading Select Form</name>
<majorMinorVersion>2012072302</majorMinorVersion>
<version>2012072302</version>
<hash>md5:29a492009d35c1515886ddd307521819</hash>
<downloadUrl>https://opendatakit.appspot.com/formXml?formId=CascadingSelect</downloadUrl>
</xform>
<xform><formID>widgets</formID>
<name>Widgets</name>
<majorMinorVersion></majorMinorVersion>
<version></version>
<hash>md5:0eacc6333449e66826326eb5fcc75749</hash>
...
```

src/org/opendatakit/briefcase/push/RemoteServer.java
```diff
@@ -0,0 +1,74 @@
...
+ * the License.
+ */
+
+package org.opendatakit.briefcase.push;
+
+import java.net.URI;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Stream;
+import org.opendatakit.briefcase.reused.http.Credentials;
```
yanokwa: What is the reason for the `reused` namespace?

test/java/org/opendatakit/briefcase/push/RemoteServerTest.java
ggalmazor: OK, I've added the header and changed the test.

test/java/org/opendatakit/briefcase/push/RemoteServerTest.java
```diff
@@ -0,0 +1,71 @@
...
+            "<name>Some Form</name>\n" +
+            "<majorMinorVersion>2012072302</majorMinorVersion>\n" +
+            "<version>2012072302</version>\n" +
+            "<hash>md5:29a492009d35c1515886ddd307521819</hash>\n" +
+            "<downloadUrl>https://some.server.com/formXml?formId=some-form</downloadUrl>\n" +
+            "</xform>\n" +
+            "</xforms>"
+    );
+    RemoteServer server = RemoteServer.authenticated("https://some.server.com", "administrator", "aggregate");
+    assertThat(http.execute(server.containsFormQuery("some-form")), is(true));
```
dcbriccetti: You can ignore this, @ggalmazor, because I know I can’t persuade you, but this 'assert that x is true' instead of 'assertTrue x' bothers me even more than `if (boolean == true)`. OK, carry on.

ggalmazor: ¯\\\_(ツ)_/¯

src/org/opendatakit/briefcase/util/ServerUploader.java
```diff
@@ -204,7 +212,14 @@ public boolean uploadFormAndSubmissionFiles(List<FormStatus> formsToTransfer) {
...
       boolean outcome;
-      outcome = uploadForm(formToTransfer, briefcaseFormDefFile, briefcaseFormMediaDir);
+      boolean existsAlready = http.execute(server.containsFormQuery(formToTransfer.getFormDefinition().getFormId()));
+      if (forceSendBlank || !existsAlready)
+        outcome = uploadForm(formToTransfer, briefcaseFormDefFile, briefcaseFormMediaDir);
+      else {
+        formToTransfer.setStatusString("Skipping form upload to remote server because it already exists", true);
+        EventBus.publish(new FormStatusEvent(formToTransfer));
+        outcome = true;
+      }
```
dcbriccetti: Below this: Bitwise **and** with booleans‽  Also:
screen shot 2018-05-15 at 12 10 13 pm

src/org/opendatakit/briefcase/push/RemoteServer.java
ggalmazor: Yes, the nuance being that you don't start off by creating stuff on `reused`. You move stuff there from a business layer when you need to use it from another business layer, thus it becomes _reused_

dcbriccetti: Is this part of some well-known software architecture scheme?

src/org/opendatakit/briefcase/util/ServerUploader.java
```diff
@@ -204,7 +212,14 @@ public boolean uploadFormAndSubmissionFiles(List<FormStatus> formsToTransfer) {
       File briefcaseFormMediaDir = FileSystemUtils.getMediaDirectoryIfExists(briefcaseLfd.getFormDirectory());
 
       boolean outcome;
-      outcome = uploadForm(formToTransfer, briefcaseFormDefFile, briefcaseFormMediaDir);
+      boolean existsAlready = http.execute(server.containsFormQuery(formToTransfer.getFormDefinition().getFormId()));
```
dcbriccetti: If you inline `existsAlready` or make it a function (lazy val in Scala), you can avoid the RPC when `forceSendBlank` is true, because the `if` will short circuit.

ggalmazor: It didn't occur to me making it lazy. Thanks for the suggestion!

ggalmazor:
> Is this meant to come out twice?

You're getting the stdout CLI feedback message and the log lines, which are being redirected to stdout by logback. Changing the logback config will produce a different console output.

src/org/opendatakit/briefcase/reused/http/Credentials.java
```diff
@@ -0,0 +1,59 @@
...
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    Credentials that = (Credentials) o;
+    return Objects.equals(username, that.username) &&
+        Objects.equals(password, that.password);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(username, password);
```
dcbriccetti: I see you’re using `Objects.hash`. Would `Objects.deepEquals` be useful, too?

ggalmazor: I don't know about that. I override `.equals` and `.hash` to make Value Objects. Any member of a Value Object must be a primitive or another Value Object for this to work, therefore, `.equals` would and should be enough.

kkrawczyk123:
### Tested with success!

PR #463, opened 2018-05-15
Issue 366 push pull ui improvements

### LINKED ISSUE #366: Unexpected behavior related to the connection settings dialog

ggalmazor:
Closes #366 
This PR follows the work started by @mayank8318 in #405

This PR also migrates the Pull and Push tabs' UI layout to IntelliJ's GUI Designer.

The new tab implementations mirror the same structures and abstractions already in place for the export tab.

The new Aggregate connection dialog handles better any error scenarios produced by wrong user input or unexpected server responses. Any error will be gracefully reported to the user, letting her accept the configuration only after all necessary checks have been made and data has been validated.

During the development of this PR, I detected some problems with the way the Export tab updates its forms table when new forms are pulled. Since I had to change things to make it work on the push tab anyway, I fixed that on the export tab as well.

I had to change the way the pull and push tabs reacted to events because they shared a common transfer event classes. This made the push tab incorrectly react to events generated from the pull tab and viceversa.

Run automated tests and tested the changes manually to the UI and the CLI. 
- When testing pulling/pushing using an Aggregate server as a source, tested it both with and without credentials (changing security configuration on the server to allow/forbid listing and uploading for the anonymous user)
- Also tested the new Aggregate connection dialog with wrong URLs, invalid URLs and empty fields to test the different error outcomes.
- Tested to pull forms from an Android device through a USB connection.

This PR will require a full QA sweep before releasing, though.

We could have made a small fix on the previous connection testing code and UI layouts, but I discarded that for the following reasons:
- The HTTP interaction code relies on global state and side effects and changing that could have potentially affected all other operations, making this PR riskier.
- Duplicating and modifying the previous HTTP interaction code was an option but taking the chance for improving the design and getting rid of the global state and side-effects made more strategic sense.
- Changing the previous UI layout would have been very difficult (we also had the issue of vertically aligning the progress bar, and decided not to fix it for the same reason)
- Migrating the layout to IntelliJ's GUI Designer was long overdue anyway

I can't think of any.

Nope.


ggalmazor:
This is missing some stuff. Please, don't review yet

ggalmazor:
@yanokwa, @lognaturel, I think this one is ready for review now.

yanokwa:
@ggalmazor  Did you get a chance to test with local file systems?

I also seem to have an issue where if I auth against the server successfully, and put in bad credentials, I can still get a form list.

ggalmazor:
> Did you get a chance to test with local file systems?

Yeah, I used an Android phone connected to my computer with an USB cable, selected the odk folder, and pulled some forms successfully.

> Also, I seem to have an issue where if I auth against the server successfully, and put in bad credentials, I can still get a form list.

If the anonymous user has enough priviledges to list and upload, it doesn't matter that you send wrong credentials, because Aggregate ignores that. Briefcase can only react to an HTTP 401 response from the server. In this case, you can leave the credentials empty.

I could add an extra check for when the user adds credentials (maybe sending them to the login servlet) but I don't know if artificially failing something that would otherwise work makes sense. What do you think, @yanokwa?

src/org/opendatakit/briefcase/ui/reused/source/RemoteServerDialog.java
```diff
@@ -0,0 +1,72 @@
...
+    this.form = form;
+
+    this.form.onConnect(server -> {
+      Response<Boolean> response = server.testPull(http);
+      if (response.isSuccess()) {
+        triggerConnect(server);
+        form.hideDialog();
+      } else {
+        String error = response.isRedirection()
+            ? "Redirection detected.\n\nPlease, review the connection parameters and try again."
```
dcbriccetti: Better without commas after these “please”es.

src/org/opendatakit/briefcase/ui/reused/source/Source.java
```diff
@@ -0,0 +1,197 @@
...
+import org.opendatakit.briefcase.model.FormStatus;
+import org.opendatakit.briefcase.model.TerminationFuture;
+import org.opendatakit.briefcase.reused.BriefcaseException;
+import org.opendatakit.briefcase.reused.RemoteServer;
+import org.opendatakit.briefcase.reused.http.Http;
+import org.opendatakit.briefcase.ui.reused.FileChooser;
+import org.opendatakit.briefcase.util.FileSystemUtils;
+import org.opendatakit.briefcase.util.TransferAction;
+
+public interface Source<T> {
```
dcbriccetti: Doc comment?

src/org/opendatakit/briefcase/operations/ClearPreferences.java
```diff
@@ -37,6 +39,8 @@ private static void clear() {
     Stream.of(
         PullTransferPanel.class,
         PushTransferPanel.class,
+        PullPanel.class,
```
dcbriccetti: I like this use of Stream

dcbriccetti:
Connection refused causes hang.
Exception in thread "AWT-EventQueue-0" java.io.UncheckedIOException: org.apache.http.conn.HttpHostConnectException: Connect to localhost:8080 [localhost/127.0.0.1, localhost/0:0:0:0:0:0:0:1, localhost/fe80:0:0:0:0:0:0:1%1] failed: Connection refused (Connection refused)

dcbriccetti:
It seems I have to configure Aggregate once each for both push _and_ pull.

dcbriccetti:
This may have nothing to do with this PR, but ...
“Success” is not a sentence and shouldn’t have a period. And wouldn’t “Pulled” be a better word for a pull status?
screen shot 2018-05-22 at 9 12 00 pm

src/org/opendatakit/briefcase/ui/pull/components/PullFormsTableView.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  private static TableCellRenderer cellWithButton() {
+    return (table, value, isSelected, hasFocus, row, column) -> {
+      JButton button = (JButton) value;
+      button.setOpaque(true);
+      button.setBackground(isSelected ? table.getSelectionBackground() : table.getBackground());
+      return button;
+    };
+  }
```
dcbriccetti: I’d like your opinion on this alternative. Is it simpler or clearer?

screen shot 2018-05-22 at 9 34 08 pm

ggalmazor:
> It seems I have to configure Aggregate once each for both push and pull.

That is correct. Sources of each tab are managed independently

src/org/opendatakit/briefcase/ui/pull/components/PullFormsTableView.java
ggalmazor: That is much much much better! Thanks

ggalmazor:
> Connection refused causes hang.
> Exception in thread "AWT-EventQueue-0" java.io.UncheckedIOException: org.apache.http.conn.HttpHostConnectException: Connect to localhost:8080 [localhost/127.0.0.1, localhost/0:0:0:0:0:0:0:1, localhost/fe80:0:0:0:0:0:0:1%1] failed: Connection refused (Connection refused)
> 
> Timeout causes a hang.
> Exception in thread "AWT-EventQueue-0" java.io.UncheckedIOException: org.apache.http.conn.HttpHostConnectException: Connect to 192.168.1.50:80 [/192.168.1.50] failed: Operation timed out (Connection timed out)

Could you copy the stacktrace, please? Also, how did you produce the refused connection scenario? (I think I know how to reproduce the timeout)

ggalmazor:
@dcbriccetti, I think I've solved all the issues now.

I've added some code to handle connection problems and I've added a progress bar to the dialog to give the user feedback should the connection test go longer than expected.

**I've set the timeout to 10 secs**. Do you think this value is OK, @yanokwa?

src/org/opendatakit/briefcase/reused/OptionalProduct.java
```diff
@@ -0,0 +1,128 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.opendatakit.briefcase.reused;
+
+import java.util.Optional;
+
+/**
+ * This class represents the type-safe Application of an arbitrary amount of {@link Optional} values.
```
dcbriccetti: For discrete measurements, use number instead of amount. What amount of milk do you want? How many games did you win.

src/org/opendatakit/briefcase/reused/OptionalProduct.java
```diff
@@ -0,0 +1,128 @@
...
+ * <p>
+ * When all its members are present, it behaves like a n-arity version of an {@link Optional} value, being able to
+ * map and flatMap its contents.
+ * <p>
+ * When any of its members is missing, it behaves like a {@link Optional#empty()}
+ */
+public interface OptionalProduct {
```
dcbriccetti: Are there libraries for Java that we could be using instead of writing all this?

ggalmazor: I only know https://arrow-kt.io/docs/typeclasses/applicative/

src/org/opendatakit/briefcase/reused/OptionalProduct.java
ggalmazor: Thanks for the explanation :)

lognaturel:
A 10 second time-out seems good to me @ggalmazor.

@dcbriccetti thanks so much for your careful look through. The review is more to see if there are any opportunities to solve problems more efficiently or to e.g. leverage libs to help and you're well positioned to identify those things.

Agreed QA will be super important here. I'm wondering whether it might be more helpful to merge some of these big PRs and do a thorough QA pass and file issues for any problems that emerge rather than doing QA for each PR. What do you think @kkrawczyk123 @ggalmazor @dcbriccetti?

dcbriccetti:
Merge and QA seems good.--

Sent from a hand computer

ggalmazor:
Closing in favor of #474

kkrawczyk123:

PR #466, opened 2018-05-18
Issue 462 pull form from filesystem

### LINKED ISSUE #462: Add option to pull an individual form from disk
This issue is related to #37 

This repo has a separate UI dialog that can be built individually to produce the [ODK Form Uploader](https://docs.opendatakit.org/form-uploader/) binary, which is used by users to upload big forms to Aggregate.

Our users can't use Briefcase for that because it lacks the option to "install" an individual form directly from their hard-drive.

We could provide a new Pull source option to "install" a form from a hard-drive location and, by doing so, Briefcase could replace the Form Uploader binary, letting us simplify the ODK tool-stack.

My proposed behavior goes like this:
1. The user selects a new "Individual form in computer" option from the "Pull data from" drop-down.
2. A file choosing dialog pops up and the user selects the form file.
3. Briefcase checks the form file and verifies that any required attachment is present.
    - If everything is OK, the form is listed on the forms table (it'd be the only form listed)
    - If Briefcase detects any problem, a dialog with feedback is shown

From this point on, the user can normally pull (install) the form just like from any other currently supported source and push it to an Aggregate instance using the Push tab.

ggalmazor:
Closes #462 

This PR is a follow-up to #463, which ends with commit https://github.com/opendatakit/briefcase/pull/466/commits/b4d9f3889504a9b04d42ad33d423509434bbabb4

- Added new automated tests
- Manually tested the GUI by installing a simple form and the sample Birds form from my computer
- After installing a form, pushed the form to Aggregate

After #463, it's relatively easy to add a new Source for pulling or pushing forms.

Nope.

Yes, [here is the link](https://github.com/opendatakit/docs/issues/711)


ggalmazor:
I've been heavily rebasing this branch and its parent and I'm noticing some changes in the order in which some commits are been applied here.

In my local branch log, #db257b3 is applied after #ec0bda6, not after #5a76ea9.
Something similar happens with #bf39afd, which should be applied after #3e07cd2.

@lognaturel, you are my git guru :) any thoughts? At the end of the day, I think it doesn't matter but I'd love to know how's this happening.

I'm thinking that I might need to remove the branch in my fork and push it again. Maybe that will get the log in the same order as in my local repo...

lognaturel:
Huh. Are you saying that when you do a `git log` locally you see a different order from here? I've never seen that! Sounds like it's not a huge deal so I would tend to not worry about it for now, rebase on master once the commits from #463 are merged and then address any issues that may still exist.

Sorry I don't have more answers! #gurufail

ggalmazor:
Yup! See for yourself:

`git log --pretty=oneline`
image

> Sounds like it's not a huge deal so I would tend to not worry about it for now, rebase on master once the commits from #463 are merged and then address any issues that may still exist.

That makes sense, I'll do that. 

Thanks for your feedback! I'm more relaxed seeing that, even if we can't explain what happened, it doesn't seem important and that we agree on the way to go with this ;)

PR #468, opened 2018-05-18
Export date range inclusive

neoliteconsultant:
Closes #


src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -56,438 +56,438 @@
 
 @SuppressWarnings("checkstyle:MethodName")
 public class ConfigurationPanelForm extends JComponent {
-  public JPanel container;
```
ggalmazor: You have change indentation on this file from 2 spaces to 4 spaces. Could you revert that, please? If you're using IntelliJ, there is a code style definition file in `config/codestyle/codestyle.xml`. You can import that to ensure this doesn't happen.

src/org/opendatakit/briefcase/operations/Export.java
```diff
@@ -15,24 +15,6 @@
  */
 package org.opendatakit.briefcase.operations;
 
-import static java.time.format.DateTimeFormatter.ISO_DATE_TIME;
```
ggalmazor: It seems like this changeset is only about imports ordering. Please, drop this commit.

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
neoliteconsultant: Sure. I will do that

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.form
```diff
@@ -31,7 +31,7 @@
           <gridbag weightx="0.0" weighty="0.0"/>
         </constraints>
         <properties>
-          <text value="Start date (inclusive)"/>
+          <text value="  Start date (inclusive)"/>
```
ggalmazor: Why these spaces? If it's to "hack" text alignment, maybe you can try to align the text to the right instead? You may need to assign the text some fixed witdh for that to work.

neoliteconsultant: Noted. I will remove the spaces

neoliteconsultant:
I have squashed my commits

lognaturel:
Thanks, @neoliteconsultant! It looks like the squash didn't quite work as expected because I'm seeing a bunch of commits here. You can read more about how to squash at https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

It's not a big deal in this case because a committer can squash on merge. That's typically what we do for 1-commit pull requests anyway so that there's a linear history and no merge commits (this is a slightly controversial practice). So if in doubt, maybe you can practice doing interactive rebases on a practice branch and not worry about this one. The diff is exactly what's desired which is what really matters. 👍

neoliteconsultant:
Thanks  @lognaturel.

ggalmazor:
Hi @neoliteconsultant!

It seems that your branch has "branched off" from our master branch and now it has a lot of different commits...

I've downloaded your fork and I think it would be enough to create a new fresh branch from opendatakit's repo's latest master branch, cherrypick your commit c9a2b0c, push it to your fork and create a new PR.

To do the cherry pick you can execute this command: `git cherry-pick c9a2b0c05c9dc5767d58c53cf8faed127a24c40a`.

neoliteconsultant:
@ggalmazor sure I will do that

neoliteconsultant:
@ggalmazor I have created a branch 'inclusive_label_export_date' and initiated a new pull request as suggested

kkrawczyk123:
I think that it doesn't work in End Date context. It looks good with Start Date but if I want to export some submissions from today I have to choose tomorrow as End Date - master has the same behavior - I thought this would change in this pr so correct me if I am wrong. 
The same behavior on all three systems: Ubuntu, MacOS and Windows.

ggalmazor:
@kkrawczyk123, I've checked the source code and the range is inclusive on both ends. If you get less submissions than expected maybe there is a bug :P

kkrawczyk123:
What is expected then? I had 3 submissions from today.

ggalmazor:
if you set the end date to today, you should get those 3 submissions exported.

kkrawczyk123:
This is exactly what I was expecting but I set yesterday's date as start date and today's as an end date, but only yesterday's submissions are exported. Today's are exported when I set tomorrow as an end date.

ggalmazor:
Alright! Then we should open a new issue for this :)

kkrawczyk123:
Testing it in #477 

lognaturel:
Closing in favor of #477.

Issue #465, opened 2018-05-17
Remove code of subapps


ggalmazor:
After #462 there is no real need for continuing to maintain the 3 extra subapps Briefcase has:
- Charset converter
  - https://forum.opendatakit.org/t/pulldata-function-in-odk/4849 confirms that we can get rid of this.
- Form uploader
  - #462 lets install an individual form from the filesystem and then push it to Aggregate.
- Clear preferences
  - The `-c` CLI operation does just that 

Any of the features of these apps are already supported by the Briefcase main UI and CLI.

Removing these classes and all their related artifacts would make further coding and reviewing work much easier.

We also reduce the number of apps to release to just one app each time we release a new version of Briefcase.

PR #470, opened 2018-05-22
Issue 469 migrate settings tab ui

### LINKED ISSUE #469: Migrate the settings tab to IntelliJ's GUI Designer
After #463, the only remaining tab that doesn't use IntelliJ's GUI Designer is the Settings tab. Migrating this tab could let us clean almost all the remaining UI code and reuse other components that we are already using on the rest of tabs.

ggalmazor:
Closes #469 

This PR is based on #466 and its first commit is https://github.com/opendatakit/briefcase/pull/470/commits/f83f2c3ec674113fb6c9f90f9f39ac2dab7dfb28

This PR migrates the Settings Tab to IntelliJ's GUI Designer. It also improves the management of the briefcase dir (storage location) and how the UI locks and unlocks itself depending on whether the storage location has been defined or not.

This PR also paves the way for #465, since most of the "old" legacy code is only used on the three alternative sub-apps in Briefcase.

Run automated tests and manually tested Briefcase behavior. During the development, I've worked with both the old and the new tab at the same time and I've been able to compare their behavior to be sure the new does what's expected.

This is a code improvement that will let us continue the work on Briefcase more comfortably.

None that I can think of.

Nope.


yanokwa:
@kkrawczyk123 This has already been merged, but did not go through QA. Can you please do a thorough test and file any issues you find? It also includes the changes from #466.

kkrawczyk123:
Everything looks good to me. I just want to ask about one small detail - Host field was filled with "127.0.0.1" by default before and now it is not but there is still a default port value - @ggalmazor should it stay like that?

ggalmazor:
I'll add the default value to the proxy host field into #466

PR #471, opened 2018-05-23
Issue 465 remove subapps

### LINKED ISSUE #465: Remove code of subapps

ggalmazor:
Closes #465 

This PR is based on #470 and its first commit is https://github.com/opendatakit/briefcase/commit/5b5945251e0d7f7d50f117bca2e85cc3aca99097

Run automated tests and manually browsed the app.

This PR only removes unused stuff and all the changes have been made using safe refactors.

Nope.

Nope


src/org/opendatakit/briefcase/util/FormCache.java
```diff
@@ -1,51 +1,59 @@
...
+  public static FormCache from(Path briefcaseDir) {
+    Path cacheFile = briefcaseDir.resolve(CACHE_FILE_NAME);
+    if (exists(cacheFile))
+      try (InputStream in = Files.newInputStream(cacheFile);
+           ObjectInputStream ois = new ObjectInputStream(in)) {
+        Map<String, String> pathToMd5Map = (Map<String, String>) ois.readObject();
+        Map<String, BriefcaseFormDefinition> pathToDefinitionMap = (Map<String, BriefcaseFormDefinition>) ois.readObject();
+        return new FormCache(cacheFile, pathToMd5Map, pathToDefinitionMap);
+      } catch (IOException | ClassNotFoundException e) {
+        throw new BriefcaseException("Can't deserialize form cache file", e);
```
dcbriccetti: 2018-05-24 11:17:43,765 [main] ERROR o.o.a.p.BaseFormParserForJavaRosa - Failed to launch GUI
org.opendatakit.briefcase.reused.BriefcaseException: Can't deserialize form cache file
    at org.opendatakit.briefcase.util.FormCache.from(FormCache.java:44)
Caused by: java.io.InvalidClassException: org.opendatakit.briefcase.model.BriefcaseFormDefinition; local class incompatible: stream classdesc serialVersionUID = 3811217215466839668, local class serialVersionUID = 3138267492033119752

Consider: Don’t throw the exception, ignore the incompatible cache file, and create a new one.

PR #473, opened 2018-05-24
Document the export CSV format

ggalmazor:
This PR adds a documentation file with information about the CSV format Briefcase uses to export submissions.

### DIFF
--- a/docs/export-format.md
+++ b/docs/export-format.md
@@ -1,0 +1,146 @@
+# Briefcase export CSV format
+
+## Format parameters
+
+| Parameter               | Value | Description |
+| ----------------------- | ----- | ----------- |
+| Separator               | `,`   | Comma character |
+| Quoted                  | `"`   | Double quote character. See the [Values block columns](#values-block-columns) section to know which fields get quoted |
+| Header                  | Yes   | Every file will have a header with the respective column names |
+| Empty columns           | Yes   | See the [Values block columns](#values-block-columns) section to know which fields get empty columns |
+| Encoding of null values | No    | Nulls are empty columns and, sometimes, empty strings |
+
+## Main output file
+
+The main output file has values from any non-repeating group field found at the topmost level of a form's model and other metadata.
+
+### Structure of the file:
+
+| Column name    | Required | Quoted | Description |
+| -------------- | -------- | ------ | ----------- |
+| SubmissionDate | Yes      | Yes    | See the [Values block columns](#values-block-columns) section to learn how dates are encoded |
+| Values block   | Yes      |        | Holds columns with the values of topmost model fields and non-repeating groups |
+| KEY            | Yes      | No     | Holds the `instanceID` of each submission |
+| isValidated    | No       | No     | It's only present if the form is encrypted. It holds `true` if the cryptographic signature has been validated, `false` otherwise |
+
+## Repeat output file
+
+Each repeat group will produce an output CSV file.
+
+### Structure of the file:
+
+| Column name         | Required | Description |
+| ------------------- | -------- | ----------- |
+| Values block        | Yes      | Holds columns with the values of the repeat group's fields |
+| PARENT_KEY          | Yes      | Identifies the parent row |
+| KEY                 | Yes      | Identifies this row |
+| SET-OF-{GROUP NAME} | Yes      | Identifies the group this row belongs to |
+
+### `PARENT_KEY`
+
+|                  | Pattern         | Example                                               |
+| ---------------- | --------------- | ----------------------------------------------------- |
+| Top group        | `{INSTANCE ID}` | `uuid:00000000-0000-0000-0000-000000000000`           |
+| Descendant group | `{PARENT KEY}`  | `uuid:00000000-0000-0000-0000-000000000000/group1[1]` |
+
+### `KEY`
+
+|                  | Pattern                                  | Example                                                         |
+| ---------------- | ---------------------------------------- | --------------------------------------------------------------- |
+| Top group        | `{INSTANCE ID}/{GROUP NAME}[{ORDERING}]` | `uuid:00000000-0000-0000-0000-000000000000/group1[1]`           |
+| Descendant group | `{PARENT KEY}/{GROUP NAME}[{ORDERING}]`  | `uuid:00000000-0000-0000-0000-000000000000/group1[1]/group2[1]` |
+
+### `SET-OF-{GROUP NAME}`
+
+|                  | Pattern                      | Example                                                      |
+| ---------------- | ---------------------------- | ------------------------------------------------------------ |
+| Top group        | `{INSTANCE ID}/{GROUP NAME}` | `uuid:00000000-0000-0000-0000-000000000000/group1`           |
+| Descendant group | `{PARENT KEY}/{GROUP NAME}`  | `uuid:00000000-0000-0000-0000-000000000000/group1[1]/group2` |
+
+## Values block columns
+
+The values block has a column for each non-repeating field.
+
+### Repeating groups
+
+Repeating groups will produce a column named using the format `SET-OF-{GROUP NAME}` and their value will correspond to the same column in its own output file.
+
+### Non-repeating groups
+
+The fields of a non-repeating group replace their parent's column in a CSV file. Their column names will be composed to be able to track the group to which they belong. Example:
+
+- Lets consider the following model:
+
+    ```xml
+    <model>
+      <field1/>
+      <field2>
+        <field3/>
+        <field4/>
+      </field2>
+    </model>
+
+    ```
+- The output CSV file will have these columns: `field1,field2-field3,field2-field4`
+- The `field3` and `field4` show up in place of the `field2`, which is not present in the file.
+- These fields get a column name composed of the name of their parent `field2` and their own name, separated by a `-` (hyphen character).
+
+This will happen in a cascading manner, if there are non-repeating groups inside non-repeating groups.
+
+A very common example of this is the `<meta>` block, which is often used to add the `instanceID` to the submission. This field will get into the CSV inside a column named `meta-instanceID`.
+
+### Geopoint fields
+
+The fields of type `GEOPOINT` will produce a column for each one of their `Latitude`, `Longitude`, `Altitude`, and `Accuracy` components. The column names will be composed like a non-repeating group, with the name of the model field and each component name, separated by a `-` (hyphen character). Example: `some_place-Latitude`.
+
+### Date and time fields
+
+These fields get encoded following the user's locale settings and Java's [MEDIUM style pattern](https://docs.oracle.com/javase/7/docs/api/java/text/DateFormat.html).
+
+### Non-empty value codification
+
+| Type               | Quoted | Example csv                                                                                    |
+| ------------------ | ------ | ---------------------------------------------------------------------------------------------- |
+| text               | Yes    | `something,"some text",something`                                                              |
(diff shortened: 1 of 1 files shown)


docs/export-format.md
```diff
@@ -0,0 +1,158 @@
...
+
+| Column block | Description |
+| ------------ | ----------- |
+| Fixed block  | Holds a column for the submission date |
+| Values block | Holds columns with the values of topmost model fields and non repeating groups |
+| Fixed block  | Holds a column for the submission key and, when the form is encrypted, a content validation flag |
+
+### Fixed block columns
+
+| Column name    | Optional | Quoted | Description |
```
yanokwa: For optional, I was expecting the inverse. That is, I want to know what things are Required.

docs/export-format.md
```diff
@@ -0,0 +1,158 @@
+# Briefcase export CSV format
+
+## Format parameters
+
+| Parameter               | Value | Description |
+| ----------------------- | ----- | ----------- |
+| Separator               | `,`   | Comma character |
+| Quoted                  | `"`   | Double quote character. See the [Values block columns](#values-block-columns) section to know which fields get quoted |
+| Header                  | Yes   | Every file will have a header with the respective column names |
+| Empty columns           | Yes   | See the "Values block columns" section to know which fields get empty columns |
```
yanokwa: Add a link for Values block columns

docs/export-format.md
```diff
@@ -0,0 +1,158 @@
...
+| Encoding of null values | No    | Nulls are just empty columns and, sometimes, empty strings |
+
+## Main submission output file
+
+The main submission out file has values from any non repeating group field found at the topmost level of a form's model and other metadata.
+
+### Structure of the file:
+
+| Column block | Description |
```
yanokwa: I would drop this table and consolidate the information in the next table.

docs/export-format.md
```diff
@@ -0,0 +1,158 @@
...
+| Fixed block  | Holds a column for the submission date |
+| Values block | Holds columns with the values of topmost model fields and non repeating groups |
+| Fixed block  | Holds a column for the submission key and, when the form is encrypted, a content validation flag |
+
+### Fixed block columns
+
+| Column name    | Optional | Quoted | Description |
+| -------------- | -------- | ------ | ----------- |
+| SubmissionDate | No       | Yes    | Holds the submission date. See the [Values block columns](#values-block-columns) section to learn how it is encoded |
+| KEY            | No       | No     | Holds the `instanceID` of each submission |
```
yanokwa: I would add another row below SubmissionDate that represents the values rows.

docs/export-format.md
```diff
@@ -0,0 +1,158 @@
...
+| PARENT_KEY          | No       | Identifies the parent row |
+| KEY                 | No       | Identifies this row |
+| SET-OF-{GROUP NAME} | No       | Identifies the group this row belongs to |
+
+### The parent key `PARENT_KEY`
+
+|                        | Pattern                                                                            | Example                                               |
+| ---------------------- | ---------------------------------------------------------------------------------- | ----------------------------------------------------- |
+| Topmost level pattern  | `{INSTANCE ID}`                                                                    | `uuid:00000000-0000-0000-0000-000000000000`           |
+| Pattern of descendants | `{RECURSIVE PARENT KEY}/{PARENT GROUP NAME}[{PLACE OF PARENT AMONG ITS SIBLINGS}]` | `uuid:00000000-0000-0000-0000-000000000000/group1[1]` |
```
yanokwa: s/{PLACE OF PARENT AMONG ITS SIBLINGS}/{ORDERING}/g

docs/export-format.md
```diff
@@ -0,0 +1,158 @@
...
+| ------------------- | -------- | ----------- |
+| PARENT_KEY          | No       | Identifies the parent row |
+| KEY                 | No       | Identifies this row |
+| SET-OF-{GROUP NAME} | No       | Identifies the group this row belongs to |
+
+### The parent key `PARENT_KEY`
+
+|                        | Pattern                                                                            | Example                                               |
+| ---------------------- | ---------------------------------------------------------------------------------- | ----------------------------------------------------- |
+| Topmost level pattern  | `{INSTANCE ID}`                                                                    | `uuid:00000000-0000-0000-0000-000000000000`           |
```
yanokwa: Maybe parent pattern and children pattern?

docs/export-format.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Fixed block columns
+
+| Column name    | Optional | Quoted | Description |
+| -------------- | -------- | ------ | ----------- |
+| SubmissionDate | No       | Yes    | Holds the submission date. See the [Values block columns](#values-block-columns) section to learn how it is encoded |
+| KEY            | No       | No     | Holds the `instanceID` of each submission |
+| isValidated    | Yes      | No     | It's only present if the form is encrypted. It holds `true` if the cryptographic signature has been validated, `false` otherwise |
+
+## Repeat output file
```
yanokwa: I wonder if it's clearer to Title this "Child output file" and reduce then mention repeat in the follow on paragraph?

docs/export-format.md
yanokwa: It's not obvious what RECURSIVE PARENT KEY means. Can you give me an example and maybe I could come up with a different wording?

docs/export-format.md
```diff
@@ -0,0 +1,158 @@
...
+
+A very common example of this is the `<meta>` block, which is often used to add the `instanceID` to the submission. This field will get into the CSV inside a column named `meta-instanceID`.
+
+### Geopoint fields
+
+The fields of type `GEOPOINT` will produce a column for each one of their `Latitude`, `Longitude`, `Altitude`, and `Accuracy` components. The column names will be composed like a non repeating group, with the name of the model field and each component name, separated by a `-` (hyphen character). Example: `some_place-Latitude`.
+
+### Date and time fields
+
+These fields get encoded following the user's locale settings and using the [MEDIUM style pattern](https://docs.oracle.com/javase/7/docs/api/java/text/DateFormat.html).
```
yanokwa: s/using the/Java's/

docs/export-format.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Non empty value codification
```
yanokwa: s/non empty/non-empty/g

docs/export-format.md
```diff
@@ -0,0 +1,158 @@
...
+
+| Parameter               | Value | Description |
+| ----------------------- | ----- | ----------- |
+| Separator               | `,`   | Comma character |
+| Quoted                  | `"`   | Double quote character. See the [Values block columns](#values-block-columns) section to know which fields get quoted |
+| Header                  | Yes   | Every file will have a header with the respective column names |
+| Empty columns           | Yes   | See the "Values block columns" section to know which fields get empty columns |
+| Encoding of null values | No    | Nulls are just empty columns and, sometimes, empty strings |
+
+## Main submission output file
```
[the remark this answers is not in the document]
ggalmazor: Are you sure about this change? All throughout the code the concept that is used is "main" and "repeat" and parent might be misleading since a repeat group can be parent of a nested repeat group, and this second group can be also parent of a third nested group.

docs/export-format.md
ggalmazor: Let's say that there is 3 nested groups g1, g2, and g3.

```
g1 PARENT KEY uuid:00000000-0000-0000-0000-000000000000
g2 PARENT KEY uuid:00000000-0000-0000-0000-000000000000/g1[1]
g3 PARENT KEY uuid:00000000-0000-0000-0000-000000000000/g1[1]/g2[1]
```

Do you think we need to change the examples on the doc's table?

docs/export-format.md
ggalmazor: Same argument as before: a child can be a parent of some other field. I'm not sure about changing the semantics.

As an alternative, we could use "root" and "node", but I'd suggest changing the code to use the same language. These dissonances between the domain and the source code are confusing down the road.

docs/export-format.md
ggalmazor: Would `{PARENT'S PARENT KEY}` work? I know the word "recursive" can be daunting.

yanokwa: What about `grandparent`. I think your existing examples are good enough.

ggalmazor: I think you're right! I was concerned about  `GRANDPARENT KEY` maybe being misleading because maybe `PARENT KEY != KEY(parent)` but it turns out that it holds.

docs/export-format.md
```diff
@@ -0,0 +1,146 @@
...
+| ---------------------- | ---------------------------------------- | --------------------------------------------------------------- |
+| Topmost level pattern  | `{INSTANCE ID}/{GROUP NAME}[{ORDERING}]` | `uuid:00000000-0000-0000-0000-000000000000/group1[1]`           |
+| Pattern of descendants | `{PARENT KEY}/{GROUP NAME}[{ORDERING}]`  | `uuid:00000000-0000-0000-0000-000000000000/group1[1]/group2[1]` |
+
+### `SET-OF-{GROUP NAME}`
+
+|                        | Pattern                                       | Example                                                      |
+| ---------------------- | --------------------------------------------- | ------------------------------------------------------------ |
+| Topmost level pattern  | `{INSTANCE ID}/{GROUP NAME}`                  | `uuid:00000000-0000-0000-0000-000000000000/group1`           |
+| Pattern of descendants | `{RECURSIVE SET-OF}[{ORDERING OF THE GROUP}]` | `uuid:00000000-0000-0000-0000-000000000000/group1[1]/group2` |
```
yanokwa: s/RECURSIVE/GRANDPARENT/

docs/export-format.md
```diff
@@ -0,0 +1,146 @@
...
+
+| Parameter               | Value | Description |
+| ----------------------- | ----- | ----------- |
+| Separator               | `,`   | Comma character |
+| Quoted                  | `"`   | Double quote character. See the [Values block columns](#values-block-columns) section to know which fields get quoted |
+| Header                  | Yes   | Every file will have a header with the respective column names |
+| Empty columns           | Yes   | See the [Values block columns](#values-block-columns) section to know which fields get empty columns |
+| Encoding of null values | No    | Nulls are empty columns and, sometimes, empty strings |
+
+## Main submission output file
```
yanokwa: s/submission//

docs/export-format.md
```diff
@@ -0,0 +1,146 @@
...
+
+The main submission out file has values from any non-repeating group field found at the topmost level of a form's model and other metadata.
```
yanokwa: s/submission out/output/

ggalmazor:
I've studied a little bit more the code that generates the keys for the repeat files and I've simplified it (I've also updated the export efficiency PR).

I've changed the first column of those tables to "Nesting Level" and rows are "First" and "Following". What do you think?

yanokwa:
Nesting level language feels like a step back. First and following don't have the same vertical hierarchy to me as Top and descendants. The latter also has a family vibe to it that first the parent language we have in the doc.

ggalmazor:
OK, I've fixed the language to top/descendant group. Let me know if there's something left ;)

ggalmazor:
@yanokwa, @lognaturel, @matthew-white, I'm sorry I forgot about this... I've updated the PR with Matt's feedback.

yanokwa:
I this is good enough to merge. @clint-tseng if in your work on Central you run into questions, please file an issue so we can fix!

PR #477, opened 2018-05-29
Inclusive label export date

### LINKED ISSUE #456: Explain users that the export date range is inclusive
#### Software versions 
Briefcase v1.10.1

#### Problem description
As discussed in #159, the export date range start and end dates are inclusive both in the GUI and the CLI. We should explicitly tell the users about this.

N/A

#### Expected behavior
This is what I propose as a narrow solution:
- The labels of the date range fields on the GUI should read "(inclusive)"
- The CLI help messages of the date start and end params should read "(inclusive)"

#### Other information 
N/A

neoliteconsultant:
Closes #456 


src/org/opendatakit/briefcase/ui/export/components/ConfigurationDialogForm.java
```diff
@@ -28,6 +28,7 @@
 import javax.swing.JComponent;
 import javax.swing.JDialog;
 import javax.swing.JPanel;
+
```
ggalmazor: Remove this added empty line

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -327,14 +329,14 @@ private boolean confirmOverwriteFiles() {
     gbc.anchor = GridBagConstraints.EAST;
     container.add(pemFileLabel, gbc);
     startDateLabel = new JLabel();
-    startDateLabel.setText("Start date");
+    startDateLabel.setText("Start date (inclusive)");
```
ggalmazor: In this file, remove all changes except this one and the next

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
```diff
@@ -327,14 +329,14 @@ private boolean confirmOverwriteFiles() {
...
     gbc = new GridBagConstraints();
     gbc.gridx = 0;
     gbc.gridy = 2;
     gbc.anchor = GridBagConstraints.EAST;
     container.add(startDateLabel, gbc);
     endDateLabel = new JLabel();
-    endDateLabel.setText("End date");
+    endDateLabel.setText("End date (inclusive)");
```
ggalmazor: In this file, remove all changes except this one and the previous

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
neoliteconsultant: Which changes are you talking about here?

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanelForm.java
neoliteconsultant: Which changes are you talking about here?

ggalmazor: Take a look at the screenshot. You'll notice how there are some changes in the order of the imports and some added empty lines, which shouldn't be merged.

image

image

ggalmazor:
That would be a bug. Could you try the same export with #457? I think it will work as expected on that PR...

lognaturel:
I think it might be a timezone issue though I'm surprised Kasia is getting a problem since she's ahead of GMT. I'm behind GMT so it's June 1 for me but June 2 for submissions on Aggregate. If I set June 1 as the end date, I don't get my recent submission. If I set the end date as June 2, I do get my recent submission.

lognaturel:
Well hang on a moment, that doesn't totally make sense. `SubmissionDate` for that submission is `Jun 1, 2018 2:38:59 PM` which is in local time. Does the comparison convert to GMT? Or use a different day boundary than we expect?

ggalmazor:
I've written a quick test to check that both ends are inclusive after #457. In the test, I can ensure that briefcase runs in the same time zone as the submission dates, which is the only way to verify this particular behavior in a controlled scenario.

There are some issues to decide which submissions to include in the output file:
- Submission dates have an offset instead of a zone which doesn't tell us about a possible daylight saving time situation.
- Submissions of the same form can be generated in different time zones, which means that two simultaneous submissions can happen on different days.

In this scenario, we always will have the problem of deciding what to do with dates and time zones:
- We can move the submission date to the local time zone
- We can assume that the start/end dates are local to the submission's time zone.
- We can move the start/end dates to the submission's time zone

Any of these can change the submission date or the filters in a way that would make a submission fall from the output file.

My opinion is that there will always be a scenario that will produce the wrong output, regardless of how much effort we take on making Briefcase more intelligent, and letting the user configure the exact behavior feels like an overkill and that it would make Briefcase too complex and far more prone to human error.

I propose we do nothing.

kkrawczyk123:
@ggalmazor I tried the case with export today's submissions on #457 and it works as expected on all 3 systems.

ggalmazor:
> When I specify a start and end date, I expect that if the textual representation of that date is in the submission XML, it will be included in the output. Is that too naive?

That's the second option I gave. 

That option will be confusing when two simultaneous submissions have different offsets and they happen to fall into different local dates: 2016-06-04 22:00:00-04:00 it's the same moment as 2016-06-05 06:00:00+0200, but a start filter of 2016-06-05 will only take the last one even though they happened at the same time.

(I forgot to mention in my last comment that I think it's the best option)

yanokwa:
That scenario you suggest is unlikely because in general, submission date is set on the server, not on the device.

For me, having something that is easy to explain to users is preferred. I think saying that we match whatever the date (and not the time) was on the server is pretty easy to explain.

ggalmazor:
Alright! I just wanted to explain that all options have drawbacks. If we agree on the second behavior, I can change the date filter accordingly because right now I think we're not doing exactly that.

PR #487, opened 2018-06-04
Use a SwingWorker in RemoteServerDialog

dcbriccetti:
I hand-tested with a real Aggregate URL, a non-existing host (timeout), and an existing host with nothing listening on the port (connection refused).

SwingWorker exists for this scenario.

No

#### Does this change require updates to documentation? 
No


dcbriccetti:
>If only it used the Builder pattern and composition instead of extension! :D

You could write something pretty to wrap it. Or we could switch to Scala and use the functional interface to Swing. 

@lognaturel Merge, if you please?

lognaturel:
Fancy! Adding the `needs testing` tag so @kkrawczyk123 knows to pay close attention in her end-to-end testing.

kkrawczyk123:

PR #488, opened 2018-06-05
Add notice in RemoteServer dialog about account kinds and permissions

### LINKED ISSUE #483: No information about inability to log by Google account on Aggregate Server Configuration window.
#### Software versions 
Briefcase v1.10.x, Java v1.8.x, operating system Ubuntu, MacOS and Windows.

#### Problem description
Aggregate Server Configuration window doesn't contain information "Username cannot be a Google login; it must be an ODK Aggregate username with "Form Manager" permissions" as it used to before. 
It doesn't have to be the same message but it might be worth to inform users that they can't use their Google accounts. 

Click on Configure button.

ggalmazor:
Closes #483 

Manually tested on Ubuntu
Verified that the correct permission is shown in pull and push tabs
Verified that the shown permissions are indeed what is needed for pulling/pushing forms

It's the narrowest change I could come with

Nope

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/reused/source/RemoteServerDialog.java
+++ b/src/org/opendatakit/briefcase/ui/reused/source/RemoteServerDialog.java
@@ -68,9 +68,9 @@ public class RemoteServerDialog {
     );
   }
 
-  static RemoteServerDialog empty(RemoteServer.Test serverTester) {
+  static RemoteServerDialog empty(RemoteServer.Test serverTester, String requiredPermission) {
     return new RemoteServerDialog(
-        new RemoteServerDialogForm(),
+        new RemoteServerDialogForm(requiredPermission),
         serverTester
     );
   }
--- a/src/org/opendatakit/briefcase/ui/reused/source/RemoteServerDialogForm.form
+++ b/src/org/opendatakit/briefcase/ui/reused/source/RemoteServerDialogForm.form
@@ -9,7 +9,7 @@
     <children>
       <grid id="86f63" layout-manager="GridBagLayout">
         <constraints>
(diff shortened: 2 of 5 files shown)


dcbriccetti:
I think these should use an enum and not Strings, but I don’t want that to hold up the beta.
`String requiredPermission`

What’s that derogatory joke? _string_ typing
http://wiki.c2.com/?StringlyTyped

src/org/opendatakit/briefcase/ui/reused/source/RemoteServerDialog.java
```diff
@@ -68,9 +68,9 @@ private void showError(String error, String title) {
     );
   }
 
-  static RemoteServerDialog empty(RemoteServer.Test serverTester) {
+  static RemoteServerDialog empty(RemoteServer.Test serverTester, String requiredPermission) {
```
dcbriccetti: I think these should use an enum and not Strings, but I don’t want that to hold up the beta.
`String requiredPermission`

What’s that derogatory joke? _string_ typing
http://wiki.c2.com/?StringlyTyped

yanokwa:
**old**
old

**new**
new

The old is what I was expecting to see, basically a label. The new seems like a text box with grayed out text that cannot be edited. Any reason why we can't make it more of a label?

ggalmazor:
I had trouble getting the line breaks on a label and I went for a text pane, and somehow I got to remove the ugly white background but now I see that it's not consistent across platforms.

I'll try to figure out how to use a label instead.

ggalmazor:
I think I've fixed the text box. Will check on mac too

kkrawczyk123:

PR #495, opened 2018-06-06
Issue 479 reload cache after creating it

### LINKED ISSUE #479: During first briefcase run push and export list shouldn't depend on pull.
#### Software versions 
Briefcase v1.10.x, Java v1.8.x, operating system: Ubuntu, MacOS and Windows.

#### Problem description
User used briefcase before and his storage directory is full of forms. User runs new briefcase version. Server url is even saved and pull list is visible. Aggregate url on push tab is saved too but push list is empty and export list also - but they should depend on storage directory content and be loaded automatically as it used to be before. I think that user should be able to use pull, push and export functionalities separately and independently too. As there is no option to refresh all list I go to pull and pull and least one form. Then finally push and export lists are loaded. 
Push and export lists were restored when I run, close, run close and run and close briefcase every time I tried - but pull list wasn't restored all the time on Ubuntu so I had to click on reset button even if url was visible on pull tab I had to input connection credentials again and again. 

What do you think about use case like that @yanokwa @lognaturel @ggalmazor

ggalmazor:
Closes #479 

Written new tests that cover all the form cache's behavior.
Also tested manually:
1. From a clean state, no prefs, no forms in storage dir, use the official release of Briefcase to pull some forms
2. Launch this PR and see that all the expected forms are there

Also tested manually to set/clear to different storage locations on my computer to see that each time I got the expected form list

Also tested manually to pull new forms and existing forms to check that the push/export tabs where updated accordingly.

I could have patched a smaller change into the `FormCache` file, but its code was long overdue a review and some tests.

While I was writing the tests I could see how Briefcase was coupled to different parts of it and how some methods inside where doing too many things at once that were getting in the way of testing it properly.

After having a good coverage of the implementation I was able to refactor it to simplify the design of parts of Briefcase that depend on the form cache.

I also have reviewed all the event flow pulling new forms > updating the cache > updating UI elements to make it more coherent and reduce duplicated processes.

Nope.

Nope.


ggalmazor:
@lognaturel, could you review this one? I know you've had problems to use the new version and this PR should solve any transition problem

src/org/opendatakit/briefcase/util/FormCache.java
```diff
@@ -1,97 +1,149 @@
...
 
-  @Override
-  public void putFormFileFormDefinition(String filePath, BriefcaseFormDefinition definition) {
-    pathToDefinitionMap.put(filePath, definition);
+  public void update() {
+    briefcaseDir.ifPresent(path -> {
+      Set<String> scannedFiles = new HashSet<>();
+      list(path.resolve("forms"))
+          .map(this::getFormFilePath)
+          .peek(p -> scannedFiles.add(p.toString()))
```
dcbriccetti: What’s the benefit of using peek here?

src/org/opendatakit/briefcase/util/FormCache.java
```diff
@@ -1,97 +1,149 @@
...
-    Path cacheFile = briefcaseDir.resolve(CACHE_FILE_NAME);
-    if (exists(cacheFile))
-      try (InputStream in = Files.newInputStream(cacheFile);
+    FormCache formCache = empty();
+    formCache.setLocation(briefcaseDir);
+    return formCache;
+  }
+
+  @SuppressWarnings("unchecked")
+  public void setLocation(Path newBriefcaseDir) {
```
dcbriccetti: It looks like now when setting a location to an existing directory with a valid cache, that cache is rebuilt. Is that intended?

src/org/opendatakit/briefcase/util/FormCache.java
ggalmazor: None whatsoever. I'll remove that

src/org/opendatakit/briefcase/util/FormCache.java
ggalmazor: Yes, we can't be sure that the user hasn't changed the files inside. The only time we don't  want to rebuild it is on startup, which, now that you mention, I think I need to review

src/org/opendatakit/briefcase/util/FormCache.java
```diff
@@ -1,97 +1,146 @@
...
+            formDefByPath.put(form.toString(), new BriefcaseFormDefinition(form.getParent().toFile(), form.toFile()));
+            hashByPath.put(form.toString(), hash);
+          } catch (BadFormDefinition e) {
+            log.warn("Can't parse form file", e);
+          }
+        }
+      });
+      hashByPath.keySet().stream().filter(p -> !scannedFiles.contains(p)).collect(toList()).forEach(hashByPath::remove);
+      formDefByPath.keySet().stream().filter(p -> !scannedFiles.contains(p)).collect(toList()).forEach(formDefByPath::remove);
+      EventBus.publish(new CacheUpdateEvent());
```
dcbriccetti: Do you need the collect here? Here’s my rewrite, with the common predicate extracted.
```
Predicate<String> notScanned = p -> !scannedFiles.contains(p);
hashByPath.keySet().stream().filter(notScanned).forEach(hashByPath::remove);
formDefByPath.keySet().stream().filter(notScanned).forEach(formDefByPath::remove);
```

dcbriccetti: Is this a problem?
> If the map is modified while an iteration over the set is in progress (except through the iterator's own remove operation), the results of the iteration are undefined.

ggalmazor: Yeah, the collect is to prevent that error. Otherwise, you are removing elements while you iterate the lists

ggalmazor: It's not very elegant. I'm open to suggetsions ;)

dcbriccetti: Wow, look what IDEA just suggested.
```
hashByPath.keySet().removeIf(notScanned);
formDefByPath.keySet().removeIf(notScanned);
```
I didn’t check that it removes the elements from the map behind the key set view, but I think it does.

ggalmazor: That's neat! I'll add that

dcbriccetti: I checked:
```
Map<String, String> m = new HashMap<>();
m.put("a", "aa");
m.put("b", "bb");
System.out.println(m);
m.keySet().removeIf(k -> k.equals("a"));
System.out.println(m);

{a=aa, b=bb}
{b=bb}
```

ggalmazor: Oh my, that's very troubling... maps are super mutable.

lognaturel:
My review of the code was relatively surface but nothing jumped out at me as problematic. This behavior is inline with what I expect! The build failure is just an unneeded import.

lognaturel:
Oh, and the commit messages are really helpful for understanding the rationale of each change, thank you @ggalmazor.

src/org/opendatakit/briefcase/operations/ImportFromODK.java
```diff
@@ -48,13 +48,15 @@
...
+    FormCache formCache = FormCache.from(briefcaseDir);
+    formCache.update();
 
     List<FormStatus> forms = FileSystemUtils.getODKFormList(odkDir.toFile()).stream()
         .map(form -> new FormStatus(FormStatus.TransferType.GATHER, form))
         .collect(toList());
 
-    TransferFromODK.pull(BriefcasePreferences.buildBriefcaseDir(Paths.get(storageDir)), odkDir, forms);
+    TransferFromODK.pull(briefcaseDir, odkDir, forms);
+    // The form cache should update and save itself now
```
dcbriccetti: Not sure this comment will age well (provide value much in the future)

src/org/opendatakit/briefcase/util/FormCache.java
```diff
@@ -1,97 +1,147 @@
...
+        if (isFormNewOrChanged(form, hash)) {
+          try {
+            formDefByPath.put(form.toString(), new BriefcaseFormDefinition(form.getParent().toFile(), form.toFile()));
+            hashByPath.put(form.toString(), hash);
+          } catch (BadFormDefinition e) {
+            log.warn("Can't parse form file", e);
+          }
+        }
+      });
+      // Warning: Remove map entries by mutating the key set works because the key set is a view on the map
```
dcbriccetti: Let this be a trigger warning for all you functional purists. ;-)

src/org/opendatakit/briefcase/util/FormCache.java
```diff
@@ -122,8 +121,9 @@ public void update() {
           }
         }
       });
-      hashByPath.keySet().stream().filter(p -> !scannedFiles.contains(p)).collect(toList()).forEach(hashByPath::remove);
-      formDefByPath.keySet().stream().filter(p -> !scannedFiles.contains(p)).collect(toList()).forEach(formDefByPath::remove);
+      // Warning: Remove map entries by mutating the key set works because the key set is a view on the map
+      hashByPath.keySet().removeIf(key -> !scannedFiles.contains(key));
+      formDefByPath.keySet().removeIf(key -> !scannedFiles.contains(key));
```
dcbriccetti: Somehow I knew you would reject my suggestion to factor out the common predicate. We’re getting to know each other! :-)

kkrawczyk123:

PR #503, opened 2018-06-06
Fix pattern to check if form exists on remote server

### LINKED ISSUE #492: There is no difference between cmd push with or without -fsb prefix
#### Software versions 
Briefcase v1.10.x, Java v1.8.x, operating system, Ubuntu and MacOS, probably windows too

#### Problem description
There is no difference between cmd push with or without -fsb prefix
Form on the aggregate without -fsb should not be updated.

1. Push version 1 of form to aggregate by cmd
2. Change form version for 2, and push again - form should not be updated on aggregate (but it is)
3. Repeat step 2 with using -fsb prefix - there is no difference in behavior but it should be
The same situation occurs with updating media folder

#### Expected behavior
In step 2 form version should not be updated, only in step 3 form should be updated.

ggalmazor:
Closes #492

Manually tested UI and CLI push operations to verify that:
- The UI, it always pushes the blank form
- The CLI, it only pushes the blank form if the form doesn't exist or when `-fsb` is used

This is just a change to rescue the pattern fix, lost while merging branches. The previous pattern works when no JR headers are sent with the form list request.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/reused/RemoteServer.java
+++ b/src/org/opendatakit/briefcase/reused/RemoteServer.java
@@ -133,7 +133,7 @@ public class RemoteServer {
   public boolean containsForm(Http http, String formId) {
     return http.execute(Request.get(baseUrl, credentials)
         .resolve("/formList")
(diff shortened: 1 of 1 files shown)


kkrawczyk123:

PR #505, opened 2018-06-06
Fix date range validation to allow start & date to be the same

### LINKED ISSUE #496: The user is not able to export data from one day
#### Software versions 
Briefcase master
Java v1.8, 
operating system: Ubuntu, macOS, Windows

#### Problem description
The user is not able to export data from one day

1. The user is on the Export tab
2. The user sets start date and end date as the same date
3. The user is not able to click on the Export button

4.  Export button is not active although calendar allows selecting the same day.

#### Expected behavior
* The user is able to export data from one day

ggalmazor:
Closes #496

Manually tested that the export button gets enabled when choosing the same date in the start and end filter fields.

This is a quite straightforward fix to ensure that the date ranges work the way we want.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/export/ExportConfiguration.java
+++ b/src/org/opendatakit/briefcase/export/ExportConfiguration.java
@@ -244,7 +244,7 @@ public class ExportConfiguration {
   }
 
   private boolean isDateRangeValid() {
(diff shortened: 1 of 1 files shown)


kkrawczyk123:

PR #508, opened 2018-06-06
Cleaner wording on sources

yanokwa:
* Ran it on macOS and confirmed the UI

* We aren't pulling/pushing data necessarily
* Custom directory isn't as clear as what we are looking for
* Ditto with individual form

* We confuse users with the new language

* https://github.com/opendatakit/docs/issues/733


ggalmazor:
I like the wording much more now :)

kkrawczyk123:

PR #513, opened 2018-06-07
Improve error handling of the export process

ggalmazor:
This PR improves error handling of the export process
- By separating the generation of CSV lines from the writing of those lines in files, we can control any parsing/mapping problem and react by gracefully ignoring the error and logging/sending feedback events
- Parsing errors can happen while trying to read submission dates. Those submissions must be skipped and reported as individual failures
- We should prevent the export process from failing due to errors while parsing/mapping individual submissions

Before this, any parsing error would stop the export and show a cryptic error dialog to the user.

Tampered an individual submission to produce xml syntax errors and exported it.
Checked that:
- the tampered submission is not present in the output.
- other submissions are present in the ouptut
- the overall process finished and shows the number of exported submissions

It's the narrowest solution I could come up with

Nope.

Nope

### DIFF
--- a/src/org/opendatakit/briefcase/export/ExportEvent.java
+++ b/src/org/opendatakit/briefcase/export/ExportEvent.java
@@ -45,6 +45,14 @@ public final class ExportEvent {
     return new ExportEvent(form.getFormId(), String.format("Failure: %s", cause), false);
   }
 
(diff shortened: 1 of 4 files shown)


lognaturel:
It might be good to have a corrupt/malformed submission case in the release suite!

kkrawczyk123:
Tested on all three systems. 

PR #514, opened 2018-06-07
Issue 512 Fix remember passwords checkbox behavior

### LINKED ISSUE #512: Forms are not pulled when `Pull before export` option is selected
**EDIT!**
A new bug which can cause this problem:
1. The user opens Briefcase and **does not select** `remember the password`
2. The user connects to aggregate and pulls one form
3. The user goes to `Export` tab and chooses `Export directory`
4. The user exports this one form
5. `Pull before export` checkbox is **ACTIVE**

#### Software versions 
Briefcase v1.11.0
Operating system: Linux, Windows, macOS
**REGRESSION - worked on ODKBriefcasev1.10.1-151gf61342b**

#### Problem description
Forms are not pulled when `Pull before export` option is selected

1. The user pulls one form
2. The user exports this form
3. The user fills this form on Collect and sends to Aggregate - make sure that form is visible on Aggregate
4. The user selects `Pull before export` and clicks on `Export` button

 5. The new submission was not added to CSV - pull before export does not work

#### Expected behavior
The form is pulled and newest submissions are visible in export file

ggalmazor:
- Now it stays disabled after exporting if the remember passwords setting is not checked
- Now it is unchecked if the user revokes the consent to remember passwords to avoid seeing it checked and disabled

Closes #512

Manually tested in the UI

Passing the preference downstream is the narrowest fix I can come up with.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanel.java
+++ b/src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanel.java
@@ -18,6 +18,7 @@ package org.opendatakit.briefcase.ui.export.components;
 import java.util.ArrayList;
 import java.util.List;
 import org.opendatakit.briefcase.export.ExportConfiguration;
+import org.opendatakit.briefcase.export.PullBeforeOverrideOption;
 
 public class ConfigurationPanel {
   private final ExportConfiguration configuration;
@@ -96,7 +97,8 @@ public class ConfigurationPanel {
     onChangeCallbacks.forEach(Runnable::run);
   }
 
-  public void setEnabled(boolean enabled) {
+  public void setEnabled(boolean enabled, boolean savePasswordsConsent) {
+    form.changeMode(savePasswordsConsent);
     form.setEnabled(enabled);
   }
 
@@ -113,6 +115,8 @@ public class ConfigurationPanel {
   }
 
   public void savePasswordsConsentRevoked() {
+    form.setPullBefore(false);
+    form.setPullBeforeOverride(PullBeforeOverrideOption.INHERIT);
     form.changeMode(false);
   }
 }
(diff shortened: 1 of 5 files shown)


src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanel.java
```diff
@@ -96,7 +97,8 @@ private void triggerOnChange() {
     onChangeCallbacks.forEach(Runnable::run);
   }
 
-  public void setEnabled(boolean enabled) {
+  public void setEnabled(boolean enabled, boolean savePasswordsConsent) {
```
dcbriccetti: Does this method now do more than setting an enabled state?

kkrawczyk123:
Tested on all three systems. On all three I obtained the same results:
When I run briefcase - pull before export wasn't active - I go to remember password checkbox and set it - pull before export was enabled and worked correctly.
But I when I didn't set remember password checkbox - pull before export wasn't active - I exported some random form and it enabled pull before export - I was able to choose it - pull before export didn't work which is good that it did not work - but it can give the user the wrong impression so we must fix it.
@ggalmazor @ln @yanokwa 

src/org/opendatakit/briefcase/ui/export/components/ConfigurationPanel.java
ggalmazor: It just needs more information to set the enabled state. When the savePasswordsConsent is `false`, the remember password checkbox must be disabled.

I use the `setEnabled` method because it's present in all `JComponents` and it felt right, but in this case, it's more like a `lockUI` that prevents/enables user interaction.

ggalmazor:
Sorry, @kkrawczyk123! Long story short, I accidentally broke the fix while fixing a test. Now it should work as advertised.

PR #515, opened 2018-06-07
Improve data sent to analytics to track it more easily

ggalmazor:
We were having trouble verifying that the usage information was being sent to GA, mainly because we weren't filling some fields that one would normally fill if the app was a website (following the recommendation on the GA manuals).

This PR sets the document path and title to the tab the user opens to trick GA into thinking that the app is a website. This improves a lot the information available on GA about Briefcase usage.

Previously we were even having trouble seen realtime usage information, which on this PR seems to work better.

Manually launched Briefcase and verified on GA that we are sending data.

No other option for the moment.

Nope

Nope

### DIFF
--- a/src/org/opendatakit/briefcase/ui/reused/Analytics.java
+++ b/src/org/opendatakit/briefcase/ui/reused/Analytics.java
@@ -57,11 +57,21 @@ public class Analytics {
   }
 
   public void enter(String screenName) {
-    buildScreenViewHit().screenName(screenName).sessionControl("start").sendAsync();
+    buildScreenViewHit()
+        .documentTitle(screenName)
+        .documentPath(screenName.toLowerCase())
+        .screenName(screenName)
+        .sessionControl("start")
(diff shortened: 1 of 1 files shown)


kkrawczyk123:
@ggalmazor  None of my actions pull/push/export from all three operating systems were tracked on google analytics in real time events.

ggalmazor:
@kkrawczyk123 I can see all my tab changes real-time... Can you check again, please? Bear in mind that it has some delay and that the real-time dashboard is not 100% reliable.

pr_515

The most reliable dashboard to verify that GA integration works is the one that lists visited pages per version: https://analytics.google.com/analytics/web/#/report/content-pages/a91951913w156325748p157918353/explorer-table.secSegmentId=analytics.appVersion&explorer-table.plotKeys=%5B%5D&explorer-table-dataTable.sortColumnName=analytics.appVersion&explorer-table-dataTable.sortDescending=true/

PR #516, opened 2018-06-07
Issue 509 Fix the behavior of http proxy fields on settings tab

### LINKED ISSUE #509: The user needs to click two times to unselect `Use HPPT Proxy` if he manipulates the Port value
#### Software versions 
Briefcase v1.11
operating system: Ubuntu, macOS, Windows

#### Problem description
The user needs to click two times to unselect `Use HPPT Proxy` if he manipulates the Port value

1. The user selects checkbox `Use HTTP Proxy`
2. The user changes `Port` value
3. The user clicks to unselect the checkbox `Use HTTP Proxy`
4. The checkbox is not unselected. `Port` value is changed to default 8080.

#### Expected behavior
The checkbox is unselected when the user clicks on it. `Port` value is changed to default 8080 at the same time. It should work just like the Host.

ggalmazor:
Closes #509

Manually tested that the fields do what's expected.
Also verified that information is still saved between launches and the proxy fields are shown in the correct state,

Narrowest change

Nope.

Nope.


kkrawczyk123:
Tested on all three systems. 

PR #519, opened 2018-06-09
[WIP] Add Automation tab

emajj:
Closes #

Still a work in progress

Yes. It's not complete and has zero test coverage.


lognaturel:
Nice to see this, @emajj! 🎉 I see you've seen @ggalmazor's comments at https://opendatakit.slack.com/archives/C374LNDK9/p1528703370000276. Don't hesitate to push more commits for feedback!

src/org/opendatakit/briefcase/automation/AutomationConfiguration.java
```diff
@@ -0,0 +1,58 @@
+package org.opendatakit.briefcase.automation;
+
+import java.nio.file.Path;
+import java.util.Optional;
+
+public class AutomationConfiguration {
```
ggalmazor: Since we're not using this class I'd suggest to remove it but we know we will eventually need one. What we don't really know is if those will be the members it will hold.

I suggest we leave just the scriptLocation and remove the rest for now.

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class AutomationPanel {
+  public static final String TAB_NAME = "Automation";
+
+  private final AutomationPanelForm view;
+
+
+  public AutomationPanel(AutomationPanelForm view) {
+    AnnotationProcessor.process(this);
+    this.view = view;
```
ggalmazor: I was expecting here something like: 
```java
view.onGenerate((AutomationConfiguration config) -> {
  // The stuff from the action listener of the generateScriptButton in AutomationPanelForm
});
```

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -0,0 +1,138 @@
...
+  private JButton scriptDirChooseButton;
+  private JButton generateScriptButton;
+
+  public AutomationPanelForm() {
+    $$$setupUI$$$();
+    scriptDirChooseButton.addActionListener(__ ->
+        directory(container, fileFrom(scriptDirField))
+            .choose()
+            .ifPresent(file -> setScriptDir(Paths.get(file.toURI()))));
+    generateScriptButton.addActionListener(__ -> {
```
ggalmazor: The code on this block belongs to the AutomationPanel. The view should not know anything about how the scripts are generated. It should just let the panel subscribe to the "generate script" business event generate an `AutomationConfiguration` from the values the user has entered using the UI

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  private static Optional<File> fileFrom(JTextField textField) {
+    return Optional.ofNullable(textField.getText())
+        .filter(StringUtils::nullOrEmpty)
+        .map(path -> Paths.get(path).toFile());
+  }
+
+  private void generateScript(String fileName, String contents) {
+    File script = new File(scriptDirField.getText(), fileName);
```
ggalmazor: We should always use the Java NIO2 API. Change this with `Files.write(Path path, byte[] bytes, OpenOption... options)`

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -0,0 +1,138 @@
...
+      PrintWriter printWriter = new PrintWriter(script);
+      printWriter.print(contents);
+      printWriter.close();
+    } catch (FileNotFoundException e) {
+      System.err.println(e.getMessage());
+    }
+  }
+
+  private boolean isOnWindows() {
+    String osName = System.getProperty("os.name");
```
ggalmazor: There are methods in `FindDirectoryStructure` that do this. Check out `FindDirectoryStructure.isWindows()`

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -0,0 +1,138 @@
...
+    return Optional.ofNullable(textField.getText())
+        .filter(StringUtils::nullOrEmpty)
+        .map(path -> Paths.get(path).toFile());
+  }
+
+  private void generateScript(String fileName, String contents) {
+    File script = new File(scriptDirField.getText(), fileName);
+    try {
+      PrintWriter printWriter = new PrintWriter(script);
+      printWriter.print(contents);
```
ggalmazor: OK, but I think we can go a little bit further and write a script that exports to `/tmp` all the forms in the `FormCache` that actually works. 

Could you also check if we can give it execution permissions for Mac & Linux hosts?

src/org/opendatakit/briefcase/automation/AutomationConfiguration.java
emajj: Ok

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
emajj: Makes sense. But I thought the AutomationPanel will be come with a configuration object which will make the onGenerate method take a Runnable instead of a Consumer<AutomationConfiguration> that way the onGenerate event handler will rely on the configuration that came with it instead of the one been sent by the AutomationPanelForm and the signature will be 

```java
view.onGenerate(() -> {

});
```
Instead of

```java
view.onGenerate(config -> {

});
```

Similar to how it's done here https://github.com/opendatakit/briefcase/blob/master/src/org/opendatakit/briefcase/ui/export/ExportPanel.java#L87-L97
```java
form.onExport(() -> {
      form.setExporting();
      List<String> errors = getErrors();
      if (errors.isEmpty()) {
        new Thread(this::export).start();
      } else {
        analytics.event("Export", "Export", "Configuration errors", null);
        showErrorDialog(getForm().getContainer(), errors.stream().collect(joining("\n")), "Export errors");
        form.unsetExporting(appPreferences.getRememberPasswords().orElse(false));
      }
    });
```

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
emajj: Why can't this be 
```java
view.onGenerate(() -> {

});
```
Like how it's done in the export panel. See response to the question below for an explanation of this.

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
ggalmazor: I know this can be confusing. The export tab is the first we refactored and I tried some stuff that I've been improving on newer tab refactors. 

My current rationale about this kind of UI components is that they produce the conf object and let the controllers subscribe with a consumer that will use the conf to do stuff.

Newer tabs like settings, pull and push work that way.

That's why suggested to have an `onGenerate(Consumer<AutomationConfiguration> callback)`.

The export tab doesn't work that way and it's way more complex than the other tabs because of this.

Does that make sense?

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -2,51 +2,70 @@
...
+        generateScript("automation.sh", config);
     });
   }
 
-  private void generateScript(String scriptName, String commands) {
+  private void generateScript(String scriptName, AutomationConfiguration configuration) {
     Path scriptDirPath = configuration.getScriptLocation().orElseThrow(() -> new BriefcaseException("No script location directory defined"));
-    Path scriptFilePath = Paths.get(scriptDirPath.toString() + scriptName);
+    String storageDir = appPreferences.getBriefcaseDir().orElseThrow(() -> new BriefcaseException("")).getParent().toString();
+    String template = "java -jar briefcase.jar --export --form_id {0} --storage_directory " +
```
ggalmazor: Use a `{N}` param for every value that could/will change: jar file, storage dir and export directory. It's cleaner to leave in the template just what will never ever change and use `{N}` params for any moving part.

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -2,51 +2,70 @@
...
-        generateScript("automation.sh", "echo $JAVA_HOME");
+        generateScript("automation.sh", config);
     });
   }
 
-  private void generateScript(String scriptName, String commands) {
+  private void generateScript(String scriptName, AutomationConfiguration configuration) {
     Path scriptDirPath = configuration.getScriptLocation().orElseThrow(() -> new BriefcaseException("No script location directory defined"));
-    Path scriptFilePath = Paths.get(scriptDirPath.toString() + scriptName);
+    String storageDir = appPreferences.getBriefcaseDir().orElseThrow(() -> new BriefcaseException("")).getParent().toString();
```
ggalmazor: Use shorter `.orElseThrow(BriefcaseException::new)` if you don't want to add a message.

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -2,51 +2,70 @@
...
+    String template = "java -jar briefcase.jar --export --form_id {0} --storage_directory " +
+        storageDir + " --export_directory /tmp --export_filename {1}.csv";
     try {
-      Files.write(scriptFilePath, commands.getBytes());
-    } catch (IOException e) {
-      System.err.println(e.getMessage());
+      PrintWriter writer = new PrintWriter(scriptDirPath.toFile() + File.pathSeparator + scriptName);
```
ggalmazor: I think your NIO2 version didn't work because you weren't building the scriptFilePath correctly. You need to use the `Path.resolve()` method for that.

ggalmazor: Also, this version would weirdly use `:` as `File.pathSeparator` which produced a wrong path and made the whole this fail. With `Path.resolve()` you don't need to care about that.

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -2,51 +2,70 @@
...
+        Object[] params = new Object[]{form.getFormId(), form.getFormName()};
+        String command = MessageFormat.format(template, params);
+        writer.println(command);
+        writer.close();
+      }
+      writer.close();
+    } catch (FileNotFoundException e) {
+      e.printStackTrace();
     }
+
```
ggalmazor: So, I'm missing some things for Mac and Linux hosts:
- Missing the [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) in the first line
- Missing POSIX execution permissions (You can set those with `Files.setPosixFilePermissions` and `PosixFilePermissions.fromString()`)

Windows hosts don't need any of that as far as I know

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -2,51 +2,70 @@
...
-    this.configuration = configuration;
-
-    view.onSelectScriptDir(path -> configuration.setScriptLocation(path));
+    this.formCache = formCache;
+    this.appPreferences = appPreferences;
 
-    view.onGenerate(() -> {
+    view.onGenerate(config -> {
+      System.out.println(config.getScriptLocation());
       if (isWindows())
```
ggalmazor: You can use an if expression here: 
```java
generateScript(isWindows() ? "automation.bat" : "automation.sh", config);
```

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -25,8 +24,7 @@
   private JTextField scriptDirField;
   private JButton scriptDirChooseButton;
   private JButton generateScriptButton;
-
-  private List<Consumer<Path>> onSelectScriptDirCallbacks = new ArrayList<>();
+  private AutomationConfiguration automationConfiguration = AutomationConfiguration.empty();
```
ggalmazor: Having a conf object as a member of the view is how we solved this problem in the `ConfigurationPanel`, but it has turned out to be much better to build the object ad-hoc just before calling the callbacks in the action listener of the generate script button.

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -36,17 +34,13 @@ public AutomationPanelForm() {
...
   }
 
-  void onSelectScriptDir(Consumer<Path> callback) {
-    onSelectScriptDirCallbacks.add(callback);
-  }
-
-  void onGenerate(Runnable callback) {
-    generateScriptButton.addActionListener(__ -> callback.run());
+  void onGenerate(Consumer<AutomationConfiguration> callback) {
+    generateScriptButton.addActionListener(__ -> callback.accept(automationConfiguration));
```
ggalmazor: Remove the `automationConfiguration` member of this class and create a new `AutomationConfiguration` object inside the action listener using the current value of the UI fields.

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -36,17 +34,13 @@ public AutomationPanelForm() {
...
   }
 
   private void setScriptDir(Path path) {
     scriptDirField.setText(path.toString());
-    onSelectScriptDirCallbacks.forEach(consumer -> consumer.accept(path));
+    automationConfiguration.setScriptLocation(path);
```
ggalmazor: No need to do this anymore. Just set the field's value.

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -0,0 +1,97 @@
...
+    AnnotationProcessor.process(this);
+    this.view = view;
+    this.formCache = formCache;
+    this.appPreferences = appPreferences;
+
+    view.onGenerate(config -> {
+      generateScript(isWindows() ? "automation.bat" : "automation.sh", config);
+    });
+
+    view.onSetExportConfiguration(() -> {
```
ggalmazor: We shouldn't add this here, since the only callback we want to use from the view is `onGenerate()`

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -0,0 +1,144 @@
...
+            .choose()
+            .ifPresent(file -> setScriptDir(Paths.get(file.toURI()))));
+  }
+
+  void onSetExportConfiguration(Runnable runnable) {
+    changeDefaultConfigurationButton.addActionListener(__ -> runnable.run());
+  }
+
+  void onGenerate(Consumer<AutomationConfiguration> callback) {
+    AutomationConfiguration config = AutomationConfiguration.empty();
```
ggalmazor: The creation of the config object should happen inside the listener's block. Here it will be created just once, when `AutomationPanel` calls this method and won't get the changes from the user's interaction with the UI.

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.form
```diff
@@ -0,0 +1,89 @@
...
+          <gridbag weightx="0.0" weighty="0.0"/>
+        </constraints>
+      </vspacer>
+      <component id="a1863" class="javax.swing.JButton" binding="changeDefaultConfigurationButton" default-binding="true">
+        <constraints>
+          <grid row="3" column="2" row-span="1" col-span="1" vsize-policy="0" hsize-policy="3" anchor="0" fill="1" indent="0" use-parent-layout="false"/>
+          <gridbag weightx="0.0" weighty="0.0"/>
+        </constraints>
+        <properties>
+          <text value="Change default configuration"/>
```
ggalmazor: It should say "Set export configuration" when there is no configuration and change to "Edit export configuration" when the user has set it. We should also add a "Clear export configuration" to remove it.

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
emajj: Ok

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
emajj: That means we don't even need a call back for this right? Just a method in the view that sets the export configuration?

emajj: One question. This is supposed to update the AutomationConfiguration.exportConfiguration field but AutomationConfiguration is not field of this class. It's always created inside the onGenerate callback. So what do I do with ExportConfiguration returned by the dialog?

ggalmazor: That’s correct: no callback needed

ggalmazor: You can store the conf object in a member of the AutomationPanelForm, inside an Optional object

src/org/opendatakit/briefcase/automation/AutomationConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+  public AutomationConfiguration setScriptLocation(Path path) {
+    this.scriptLocation = Optional.ofNullable(path);
+    return this;
+  }
+
+  public Optional<ExportConfiguration> getExportConfiguration() {
+    return exportConfiguration;
+  }
+
+  public AutomationConfiguration setExportConfiguration(ExportConfiguration exportConfiguration) {
```
ggalmazor: I think you don't need this method. In fact, we want the conf to be immutable, so no setters should be allowed.

src/org/opendatakit/briefcase/automation/AutomationConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public static AutomationConfiguration empty() {
+    return new AutomationConfiguration(Optional.empty(), Optional.empty());
+  }
+
+  public Optional<Path> getScriptLocation() {
+    return scriptLocation;
+  }
+
+  public AutomationConfiguration setScriptLocation(Path path) {
```
ggalmazor: I think you don't need this method. In fact, we want the conf to be immutable, so no setters should be allowed.

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -0,0 +1,164 @@
...
+  private Optional<ExportConfiguration> exportConfiguration = Optional.empty();
+
+  public AutomationPanelForm() {
+    $$$setupUI$$$();
+    scriptDirChooseButton.addActionListener(__ ->
+        directory(container, fileFrom(scriptDirField))
+            .choose()
+            .ifPresent(file -> setScriptDir(Paths.get(file.toURI()))));
+    changeDefaultConfigurationButton.addActionListener(__ -> setExportConfiguration());
+    if (exportConfiguration.isPresent()) {
```
ggalmazor: This won't ever happen, so it can be removed.

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -0,0 +1,164 @@
...
+import org.opendatakit.briefcase.util.StringUtils;
+
+@SuppressWarnings("checkstyle:MethodName")
+public class AutomationPanelForm {
+  public JPanel container;
+  private JLabel scriptDirLabel;
+  private JTextField scriptDirField;
+  private JButton scriptDirChooseButton;
+  private JButton generateScriptButton;
+  private JButton changeDefaultConfigurationButton;
```
ggalmazor: Since we're also going to have the pull and push configurations, I think it would be better to rename this to `setExportConfigurationButton`

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -0,0 +1,164 @@
...
+            .ifPresent(file -> setScriptDir(Paths.get(file.toURI()))));
+    changeDefaultConfigurationButton.addActionListener(__ -> setExportConfiguration());
+    if (exportConfiguration.isPresent()) {
+      changeDefaultConfigurationButton.setText("Change export configuration");
+    }
+  }
+
+  private void setExportConfiguration() {
+    ConfigurationDialog dialog = ConfigurationDialog.from(exportConfiguration, true, BriefcasePreferences.getStorePasswordsConsentProperty());
+    dialog.onOK(config -> {
```
ggalmazor: You can contract this into `dialog.onOK(config -> exportConfiguration = Optional.ofNullable(config));`

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -0,0 +1,164 @@
...
+      exportConfiguration = Optional.ofNullable(config);
+    });
+    dialog.open();
+    if (exportConfiguration.isPresent()) {
+      changeDefaultConfigurationButton.setText("Change export configuration");
+    }
+  }
+
+  void onGenerate(Consumer<AutomationConfiguration> callback) {
```
ggalmazor: Perrrrrrrfect!

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -48,12 +48,12 @@ private void generateScript(String scriptName, AutomationConfiguration configura
     Path storageDir = appPreferences.getBriefcaseDir().orElseThrow(BriefcaseException::new).getParent();
     String template = "java -jar {0} --export --form_id {1} --storage_directory {2} --export_directory {3} --export_filename {4}.csv";
 
-    List<String> scriptLines = formCache.getForms()
+    List<String> scriptLines = forms.getSelectedForms()
```
ggalmazor: Looking good!

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -34,33 +38,76 @@
...
+  public AutomationPanel(AutomationPanelForm view, TransferForms forms, BriefcasePreferences tabPreferences, BriefcasePreferences appPreferences, FormCache formCache) {
     AnnotationProcessor.process(this);
     this.view = view;
     this.forms = forms;
     this.formCache = formCache;
     this.appPreferences = appPreferences;
 
+    // Read prefs and load saved remote server if available
+    this.pullSource = RemoteServer.readPreferences(tabPreferences).flatMap(view::preloadPullSource);
+    this.pullSource.ifPresent(source -> {
```
ggalmazor: Since the pull source should be always empty on creation, this block should be removed as well.

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -34,33 +38,76 @@
...
+
+    view.onPushSource(pushSource -> {
+      Source.clearAllPreferences(tabPreferences);
+      pushSource.storePreferences(tabPreferences, getStorePasswordsConsentProperty());
+      forms.load(pushSource.getFormList());
+      view.refresh();
+    });
+
+    // Read prefs and load saved remote server if available
+    this.pushSource = RemoteServer.readPreferences(tabPreferences).flatMap(view::preloadPushSource);
```
ggalmazor: Same comments as with the sentence in line 52. Remove

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -34,33 +38,76 @@
...
   }
 
   private void generateScript(String scriptName, AutomationConfiguration configuration) {
     Path scriptDirPath = configuration.getScriptLocation().orElseThrow(BriefcaseException::new);
     Path storageDir = appPreferences.getBriefcaseDir().orElseThrow(BriefcaseException::new).getParent();
-    String template = "java -jar {0} --export --form_id {1} --storage_directory {2} --export_directory {3} --export_filename {4}.csv";
+    String exportTemplate = "java -jar {0} --export --form_id {1} --storage_directory {2} --export_directory {3} --export_filename {4}.csv";
 
-    List<String> scriptLines = forms.getSelectedForms()
+    List<String> scriptLines = pullSource.orElseThrow(BriefcaseException::new).pullScriptLines(forms.getSelectedForms(), appPreferences);
```
ggalmazor: I have some comments about the mechanics that you've laid out here. A `Source` should not know anything about the automation scripts. Any transformation from values in memory to actual script lines should happen here, not in `Source`.

The way I see it, `Source` should return some `Map<String,String>` with its internal params, and then that map could be used to fill in template params here.

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -34,33 +38,76 @@
...
+    List<String> pushInstructions = pushSource.orElseThrow(BriefcaseException::new).pushScriptLines(forms.getSelectedForms(), appPreferences);
```
ggalmazor: Same from previous line here.

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -80,6 +127,7 @@ public static AutomationPanel from(Http http, BriefcasePreferences appPreference
     return new AutomationPanel(
         AutomationPanelForm.from(http, forms),
         forms,
+        BriefcasePreferences.forClass(AutomationPanel.class),
```
ggalmazor: This shouldn't be required on this class since we're removing all interaction with preferences.

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -89,6 +90,14 @@ void onPushSource(Consumer<Source<?>> callback) {
     pushSourcePanel.onSource(callback);
   }
 
+  public Optional<Source<?>> preloadPullSource(RemoteServer server) {
```
ggalmazor: Since we're removing the interaction with prefs, this method can be removed

src/org/opendatakit/briefcase/ui/automation/AutomationPanelForm.java
```diff
@@ -89,6 +90,14 @@ void onPushSource(Consumer<Source<?>> callback) {
...
+    return pushSourcePanel.preload(server);
+  }
+
+  public Optional<Source<?>> preloadPushSource(RemoteServer server) {
```
ggalmazor: Since we're removing the interaction with prefs, this method can be removed

src/org/opendatakit/briefcase/ui/reused/source/Source.java
```diff
@@ -190,6 +192,16 @@ static void clearAllPreferences(BriefcasePreferences prefs) {
    */
   String getDescription();
 
+  /**
+   * Returns pull instructions for forms
+   *
+   * @param forms {@link List} of forms to be pulled
+   * @return lines of instructions to pull
+   */
+  List<String> pullScriptLines(List<FormStatus> forms, BriefcasePreferences appPreferences);
```
ggalmazor: So, return a `Map<String, String>` instead of this? Definitely not a list of strings with java invocations in them

src/org/opendatakit/briefcase/ui/reused/source/Source.java
```diff
@@ -190,6 +192,16 @@ static void clearAllPreferences(BriefcasePreferences prefs) {
...
+
+  List<String> pushScriptLines(List<FormStatus> forms, BriefcasePreferences appPreferences);
```
ggalmazor: So, return a `Map<String, String>` instead of this? Definitely not a list of strings with java invocations in them

src/org/opendatakit/briefcase/ui/reused/source/Source.java
```diff
@@ -253,6 +265,48 @@ public String getDescription() {
       return server.getBaseUrl().toString();
     }
 
+    @Override
+    public List<String> pullScriptLines(List<FormStatus> forms, BriefcasePreferences appPreferences) {
```
ggalmazor: Change the implementation once we decide what should these methods return

src/org/opendatakit/briefcase/ui/reused/source/Source.java
```diff
@@ -253,6 +265,48 @@ public String getDescription() {
...
+              server.getBaseUrl(),
+              username,
+              password
+          ))
+          .collect(Collectors.toList());
+      return pullInstructions;
+    }
+
+    @Override
+    public List<String> pushScriptLines(List<FormStatus> forms, BriefcasePreferences appPreferences) {
```
ggalmazor: Change the implementation once we decide what should these methods return

src/org/opendatakit/briefcase/ui/reused/source/Source.java
```diff
@@ -332,6 +386,26 @@ public String getDescription() {
       return path.toString();
     }
 
+    @Override
+    public List<String> pullScriptLines(List<FormStatus> forms, BriefcasePreferences appPreferences) {
```
ggalmazor: Change the implementation once we decide what should these methods return

src/org/opendatakit/briefcase/ui/reused/source/Source.java
```diff
@@ -408,6 +482,16 @@ public String getDescription() {
       return String.format("%s at %s", form.getFormName(), path.toString());
     }
 
+    @Override
+    public List<String> pullScriptLines(List<FormStatus> forms, BriefcasePreferences appPreferences) {
```
ggalmazor: Change the implementation once we decide what should these methods return

src/org/opendatakit/briefcase/ui/reused/source/Source.java
```diff
@@ -408,6 +482,16 @@ public String getDescription() {
...
+      throw new BriefcaseException("Not yet implemented for FormInComputer");
+    }
+
+    @Override
+    public List<String> pushScriptLines(List<FormStatus> forms, BriefcasePreferences appPreferences) {
```
ggalmazor: Change the implementation once we decide what should these methods return

src/org/opendatakit/briefcase/ui/reused/source/Source.java
```diff
@@ -408,6 +482,16 @@ public String getDescription() {
       return String.format("%s at %s", form.getFormName(), path.toString());
     }
 
+    @Override
+    public List<String> pullScriptLines(List<FormStatus> forms, BriefcasePreferences appPreferences) {
+      throw new BriefcaseException("Not yet implemented for FormInComputer");
```
ggalmazor: Once we change what this method returns, maybe we won't need to throw an exception. Maybe we can return the internal state of this instance and let the calling site decide what to do with it.

src/org/opendatakit/briefcase/ui/reused/source/Source.java
```diff
@@ -408,6 +482,16 @@ public String getDescription() {
...
+    }
+
+    @Override
+    public List<String> pushScriptLines(List<FormStatus> forms, BriefcasePreferences appPreferences) {
+      throw new BriefcaseException("Not yet implemented for FormInComputer");
```
ggalmazor: Once we change what this method returns, maybe we won't need to throw an exception. Maybe we can return the internal state of this instance and let the calling site decide what to do with it.

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -34,33 +38,76 @@
...
+      forms.load(source.getFormList());
+      view.refresh();
+    });
+
+    view.onPullSource(pullSource -> {
+      this.pullSource = Optional.of(pullSource);
+      Source.clearAllPreferences(tabPreferences);
+      pullSource.storePreferences(tabPreferences, getStorePasswordsConsentProperty());
+      forms.load(pullSource.getFormList());
+      view.refresh();
```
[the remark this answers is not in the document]
emajj: thought when you connect to an aggregate server it's supposed to get the forms from there and hence update the table?

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -34,33 +38,76 @@
...
+    this.pullSource.ifPresent(source -> {
+      forms.load(source.getFormList());
+      view.refresh();
+    });
+
+    view.onPullSource(pullSource -> {
+      this.pullSource = Optional.of(pullSource);
+      Source.clearAllPreferences(tabPreferences);
+      pullSource.storePreferences(tabPreferences, getStorePasswordsConsentProperty());
+      forms.load(pullSource.getFormList());
```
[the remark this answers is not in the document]
emajj: I know in the automation tab we get the forms from formCache but doesn't that only read forms stored in the briefcase storage directory?

if we have to pull a form from aggregate, how do we then update the table after connecting to an aggregate server?

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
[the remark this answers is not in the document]
ggalmazor: Nope, that's the Pull tab's concern. Here we just want to enter data to create the automation script

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -117,6 +119,15 @@ private void generateScript(String scriptName, AutomationConfiguration configura
 
   }
 
+  private void writeBriefcaseJarTwoScript(List<String> scriptLines) {
```
ggalmazor: In this method, we're changing an input argument to have some sorts of output (we add elements to a list) but this is something we always want to avoid. 

You can avoid mutating the input argument by returning the line it produces, for example.

As a general rule, we always want our methods to behave like [referentially transparent functions](https://en.wikipedia.org/wiki/Referential_transparency?oldformat=true)

src/org/opendatakit/briefcase/ui/automation/TransferFormsTableView.java
```diff
@@ -58,8 +62,12 @@ public TransferFormsTableView(TransferFormsTableViewModel model, String[] header
     columns.getColumn(SELECTED_CHECKBOX_COL).setPreferredWidth(40);
     columns.getColumn(FORM_NAME_COL).setMinWidth(formNameDims.width + 25);
     columns.getColumn(FORM_NAME_COL).setPreferredWidth(formNameDims.width + 25);
-    columns.getColumn(STATUS_COL).setMinWidth(statusDims.width + 25);
-    columns.getColumn(STATUS_COL).setPreferredWidth(statusDims.width + 25);
+    columns.getColumn(PULL_COL).setMinWidth(pullDims.width + 25);
```
ggalmazor: I think that you could set the `minWidth` and `preferredWith` to a fixed value of `40`, like the checkbox in the first column.

ggalmazor:
Last commits are looking good. Thanks @emajj!

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
```diff
@@ -68,6 +68,7 @@ private void generateScript(String scriptName, AutomationConfiguration configura
     String briefcaseJar = isWindows() ? "%JAR%" : "$JAR";
     List<String> pullInstructions = forms.getSelectedForms()
         .stream()
+        .filter(FormStatus::isPullSelected)
```
ggalmazor: This three lines could be `forms.getPullForms()` or `forms.getSelectedForPulling()`

src/org/opendatakit/briefcase/model/FormStatus.java
```diff
@@ -28,6 +28,9 @@
   private String statusString = "";
   private final StringBuilder statusHistory = new StringBuilder();
   private boolean isSuccessful = true;
+  private boolean isPullSelected = false;
```
ggalmazor: This is a good idea, but this class is used by a myriad of other processes. Could you track the selection of forms in some other way? You could have some new methods and members in `TransferForms`, for example.

src/org/opendatakit/briefcase/ui/automation/AutomationPanel.java
ggalmazor: Same for the other blocks

src/org/opendatakit/briefcase/ui/automation/TransferFormsTableViewModel.java
```diff
@@ -79,11 +79,11 @@ public Object getValueAt(int rowIndex, int columnIndex) {
       case TransferFormsTableView.FORM_NAME_COL:
         return form.getFormName();
       case TransferFormsTableView.PULL_COL:
-        return false;
+        return form.isPullSelected();
```
ggalmazor: `forms.isPullSelected(form)`

src/org/opendatakit/briefcase/model/FormStatus.java
ggalmazor: It's kind of risky changing this form because it belongs to the part of Briefcase that hasn't been refactored yet.

emajj:
I've always been curious about what's left with the PR. I remember our last
chat you told me not to worry about a sqashing the commits and I thought
this was good to go.

PR #521, opened 2018-06-11
Fix (recover) missing pull/push/export operation GA events

ggalmazor:
Manually pulled/pushed/exported forms to see event hits in the realtime report in GA

This is just some code that was missed during some merge conflict resolution

Nope.

Nope.


ggalmazor:
@kkrawczyk123, this PR fixes sending the events on GA when pulling/pushing/exporting forms.

kkrawczyk123:
Verified on Ubuntu, MacOS and Windows with success!

Issue #522, opened 2018-06-12
Log file not created


lognaturel:
#### Software versions 
Briefcase v1.1.11.beta

#### Problem description
@aurdipas did not have a log file after running the beta: https://forum.opendatakit.org/t/odk-briefcase-v1-11-beta/13432/13

#### Steps to reproduce the problem
Java version "1.8.0_172"
Windows 7 enterprise Service pack 1

#### Expected behavior
A log file should be created.

ggalmazor:
This has to do with the build process: a configured `logback.xml` has to be placed in the `res` directory.

That logback conf file has to have a file appender configured to produce a `briefcase.log` file.

ggalmazor:
For a long time, I have been wanting to add a `.example` file with that. @kkrawczyk123 would benefit from that too for the QA tests.

ggalmazor:
I think this one can be closed :)

lognaturel:
👍Do you remember why you said "partially addresses" in #526?

ggalmazor:
Sure! I wanted to discuss this with you and @yanokwa before closing it just in case you wanted to do something more than having that example conf file on the repo.

lognaturel:
Got it! Then I think we can close. That file will be packaged as `logback.xml` for releases and so we should be able to ask users for their logs as needed.

PR #526, opened 2018-06-12
Add a rolling file appender to a logback conf example

ggalmazor:
Partially addresses #522 

Run Brierfcase activating the new appender and verified that the `briefcase.log` file was generated.
Changed date and verified that the file rolling policy worked as expected (generates one file per day).

This is the rolling file appender config explained in logback docs

Nope.

Nope.

### DIFF
--- a/res/logback.xml.example-release
+++ b/res/logback.xml.example-release
@@ -1,18 +1,24 @@
 <configuration>
-  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-    <encoder>
-      <pattern>%d [%thread] %-5level %logger{36} - %msg%n</pattern>
-    </encoder>
-  </appender>
-
-  <appender name="Sentry" class="io.sentry.logback.SentryAppender">
+  <appender name="SENTRY" class="io.sentry.logback.SentryAppender">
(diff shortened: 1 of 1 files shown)


lognaturel:
I didn't even look at the logback file because I thought the production one was already going to file! 🤦🏻‍♀️ 

This looks like the behavior we want. Do we want to log to stdout at all in this case? Do we want to perhaps just make this the default `logback.xml` file?

3gb is high for the limit. Hopefully it won't get to that in 30 days, but how about making it 500mb or even 100mb?

ggalmazor:
> Do we want to log to stdout at all in this case? 

You're right, I'll remove the appender from the example conf.

> Do we want to perhaps just make this the default logback.xml file?

I don't think that's a good idea. Different conf files are best suited while developing, debugging, testing... Making this the default would require to change it every time there is an upstream repo sync and one would have to be really careful not to commit any accidental change to it.

The general rule for conf files in the repo (also in Aggregate, mostly) is to ignore the actually required conf file and commit some example files for different scenarios.

> 3gb is high for the limit. Hopefully it won't get to that in 30 days, but how about making it 500mb or even 100mb?

Let's start with 100mb and change it if we get feedback about it.

kkrawczyk123:
**Tested with success!**

PR #527, opened 2018-06-12
Issue 491 review pull source saving conditions

### LINKED ISSUE #491: Form list is empty on Pull tab when the user restarts Briefcase
#### Software versions 
Briefcase master
Java v1.8
operating system: Linux, Windows
Problem is not visible on macOS

#### Problem description
EDIT: Problem is visible only when credentials are required.
Form list is empty on Pull tab when the user restarts Briefcase
The empty list was also on v1.10.1 but the user was not connected to Aggregate. 

Now, the user needs to reset and add URL and credentials again. 
1. The user opens Briefcase and connects to Aggregate
2. The user sees forms on the Pull tab

3. The user closes Briefcase
4. The user reopens Briefcase
5. The user is connected to aggregate but form list is empty

#### Expected behavior
Forms are visible

ggalmazor:
Closes #491

Pull tab verifications, starting with clean prefs (remember passwords unset):
- Set a Pull source with an Aggregate that requires credentials. Verify that the form list is filled.
- Close and reopen Briefcase. Verify that the source is unset and the form list is empty.
- Check the remember passwords setting.
- Set a Pull source with an Aggregate that requires credentials. Verify that the form list is filled.
- Close and reopen Briefcase. Verify that the source is set and the form list is filled.
- Uncheck the remember passwords setting. Verify that nothing changes on the pull tab.
- Close and reopen Briefcase. Verify that the source is unset and the form list is empty.
- Set a Pull source with an Aggregate that doesn't require credentials. Verify that the form list is filled.
- Close and reopen Briefcase. Verify that the source is set and the form list is filled.

Push tab verifications, starting with clean prefs (remember passwords unset):
- Set a Push target with an Aggregate that requires credentials.
- Close and reopen Briefcase. Verify that the target is unset.
- Check the remember passwords setting.
- Set a Push target with an Aggregate that requires credentials.
- Close and reopen Briefcase. Verify that the target is set.
- Uncheck the remember passwords setting. Verify that nothing changes on the push tab.
- Close and reopen Briefcase. Verify that the target is unset.
- Set a Push source with an Aggregate that doesn't require credentials.
- Close and reopen Briefcase. Verify that the target is set.

This is a quick fix to prevent confusing situations for the current release (June'18). On #491 a more involved solution is discussed.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/reused/RemoteServer.java
+++ b/src/org/opendatakit/briefcase/reused/RemoteServer.java
@@ -100,14 +100,22 @@ public class RemoteServer {
   }
 
   public void storePreferences(BriefcasePreferences prefs, boolean storePasswords) {
+    prefs.remove(AGGREGATE_1_0_URL);
+    prefs.remove(USERNAME);
     prefs.remove(PASSWORD);
-    prefs.put(AGGREGATE_1_0_URL, getBaseUrl().toString());
-    credentials.ifPresent(cs -> {
-      prefs.put(USERNAME, cs.getUsername());
-      if (storePasswords)
-        prefs.put(PASSWORD, cs.getPassword());
-    });
 
+    // We only save the Aggregate URL if no credentials are defined or
+    // if they're defined and we have the user's consent to save passwords,
+    // to avoid saving a URL that won't work without credentials.
+    if (!credentials.isPresent() || storePasswords)
+      prefs.put(AGGREGATE_1_0_URL, getBaseUrl().toString());
+
+    // We only save the credentials if we have the user's consent to save
+    // passwords
+    if (credentials.isPresent() && storePasswords) {
+      prefs.put(USERNAME, credentials.get().getUsername());
+      prefs.put(PASSWORD, credentials.get().getPassword());
+    }
   }
 
   public URL getBaseUrl() {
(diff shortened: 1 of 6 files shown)


kkrawczyk123:
Behavior on pull and push tab has been unified which is good and works as described. 

I think that we should consider two cases more which might be confusing for users:

- Turning off the Remember passwords works always no matter if you choose it before logging in or after that. Turning on Remember password options works only when it’s set before logging in. If you log first and then set Remember passwords option it won’t work even if the checkbox will be set. So I think that it is possible that user encounter situation like that: log in, check Remember Password and expect that connection will be saved - but it is not. It might be confusing as checkbox has been chosen.

-   When credentials are not required from aggregate but set anyway - target is not saved after relaunching briefcase and it can be a little bit confusing too, as the behavior is exactly the same as for aggregate which requires logging in and target are saved only when Remember password is chosen. So it can be difficult for the user to differentiate those behaviors.

@lognaturel, @yanokwa and @ggalmazor what are your impressions about that?

yanokwa:
Hi @kkrawczyk123, since this PR has been merged, can you please file these are separate issues so we can discuss them. And if you have ideas on what the behavior should be, include them in the issues.

yanokwa:
Hi @kkrawczyk123, since this PR has been merged, can you please file these are separate issues so we can discuss them. And if you have ideas on what the behavior should be, include them in the issues.

PR #533, opened 2018-06-14
Issue 532 improve http proxy form fields

### LINKED ISSUE #532: Improve HTTP Proxy definition form fields
#### Software versions 
Briefcase v1.11.0 beta 0

#### Problem description
It seems like the form doesn't verify the user's input and it can produce an uncaught `IllegalArgumentException` when the user leaves the HTTP Proxy host field empty.

Leave the HTTP Proxy field empty and change focus.

#### Expected behavior
It should not trigger any processing of the HTTP Proxy when there is no host.
It should maybe show error feedback using a dialog.

ggalmazor:
Closes #532

Manually test the form's behavior by leaving the HTTP Proxy host field empty.
Verify that there are no errors logged.
Verify that after closing and reopening Briefcase, the HTTP Proxy has not been saved

This is the smallest solution to the issue.

I opted not to give feedback by popping an error dialog because Briefcase wasn't saving anything anyway and it felt too intrusive.

Nope

Nope


kkrawczyk123:
**Tested with success!**

PR #534, opened 2018-06-14
Issue 531 remove dupe server conf test code

### LINKED ISSUE #531: Remove duplication of server conn params tests
With the PR #474 we introduced a side-effect free way of testing server connection params. The new test uses an HTTP interaction layer with improved error handling and feedback.

There are still some places in Briefcase that use the old `ServerConnectionTest`, which could be replaced with the new tests.

ggalmazor:
Closes #531 

Manually launched push/pull ops with correct and wrong conf params
Verified that all wrong params error feedback outputs are given.

This reduces code complexity by removing duplication.

Nope.

Nope.


kkrawczyk123:
**Tested with success!**

PR #538, opened 2018-06-19
Issue 398 cli pull in parallel

### LINKED ISSUE #398: CLI should provide a way to enable and disable experimental downloads
#### Software versions 
Briefcase v1.9.0

#### Problem description
I use Briefcase on the CLI on a server and I would like to enable the experimental download (because it's super fast). I can't do that easily because it's off by default and there is no way to turn it on without installing a full GUI on my server. 

#### Expected behavior
I'd expect to be able to toggle some of these settings from the CLI. Something like `--parallel_submission=true`. 

Alternatively, I'd like to be able to configure settings on a machine with a GUI and copy those settings to the machine without a GUI.

ggalmazor:
Closes #398

This PR adds a new CLI param `-pp` or `--parallel_pull` to enable pulling forms in parallel on the CLI. Before this, the CLI would use whatever saved preference the user would have set using the UI's Settings tab.

This PR also changes UI code to decouple it from global state - which was harmful in many ways - wherever we needed to know if the "pull in parallel" setting was set or not.

Pulled a form with the UI with pull in parallel setting checked and unchecked. Verified on the logs that Briefcase would only use 1 thread if unchecked.
Pulled a form with the CLI adding and omitting the new `-pp` flag. Verified on the logs that Briefcase would only use 1 thread if the flag was omitted.

This a pretty straightforward change: break coupling to global state by replacing it with injected values. Then inject values from a new source.

Almost all the changes have been made using safe refactoring tools in IntelliJ.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
+++ b/src/org/opendatakit/briefcase/model/BriefcasePreferences.java
@@ -16,8 +16,6 @@
 
 package org.opendatakit.briefcase.model;
 
-import java.io.IOException;
-import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.Security;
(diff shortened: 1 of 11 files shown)


lognaturel:
Looks good to me. Would be great to get a QA pass on this before merging since I don't think there's a rush (@ggalmazor is there?).

kkrawczyk123:
### Tested with success!
Verified on Ubuntu, MacOS and Windows. I obtained the same results. Checked log files and did some efficiency testing to compare pulling with and without -pp prefix. Pulling with -pp prefix is twice timed shorter than pulling without -pp prefix.

kkrawczyk123:

PR #545, opened 2018-06-27
Issue 544 make submissions directory filter less restrictive

### LINKED ISSUE #544: Export to CSV results in empty CSV
#### Software versions 
Briefcase v1.11.0

#### Problem description
Exporting to CSV in v1.11.0 results in a CSV with just the headers.

* Clear prefs
* Import from Collect odk storage folders
* Export to CSV

#### Expected behavior
Export CSV has more than header rows.

#### Other information 
Briefcase v1.10 works great, so this is a bug introduced in v1.11

ggalmazor:
Closes #544

This PR changes the filter Briefcase uses when listing the contents of the instances folder of a form to retrieve the submissions to be exported. Before this change, we would expect a particular format that would leave out forms using arbitrary instanceIDs and forms pulled from ODK Collect directories.

Now, the filter will reject only non-directory entries and hidden directories (directories starting with a `.` character are hidden in mac and linux computers)

- Pull a form with submissions from an ODK Collect directory and export it. 
- Artificially add a `.DS_Store` directory inside the instances folder of the pulled form and add an empty `submission.xml` file in it.
- Verify that all submissions are present and no errors are logged

I've also improved the unit test suite that covers the directory filter with new test cases for this.

This is the narrowest fix I could come with

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/reused/UncheckedFiles.java
+++ b/src/org/opendatakit/briefcase/reused/UncheckedFiles.java
@@ -218,11 +218,9 @@ public class UncheckedFiles {
 
   public static boolean isInstanceDir(Path dir) {
     return Files.isDirectory(dir)
-        // Instance directories follow the pattern "uuid01234567-0123-0123-0123-012345678901"
-        // ("uuid" followed by a UUID string)
-        // This is decided when forms are pulled and written into the storage directory
-        // TODO Extract into an artifact, test, and use everywhere where this convention is used
-        && dir.getFileName().toString().matches("^uuid[\\w]{8}-[\\w]{4}-[\\w]{4}-[\\w]{4}-[\\w]{12}$")
+        // Ignore hidden mac/linux hidden folders
+        && !dir.getFileName().toString().startsWith(".")
+        // Check for presence of a submission.xml file inside
         && Files.exists(dir.resolve("submission.xml"));
   }
 
--- a/test/java/org/opendatakit/briefcase/reused/UncheckedFilesIsInstanceDirTest.java
+++ b/test/java/org/opendatakit/briefcase/reused/UncheckedFilesIsInstanceDirTest.java
@@ -70,7 +70,7 @@ public class UncheckedFilesIsInstanceDirTest {
   }
 
   @Test
-  public void a_linux_or_mac_hidden_directoryis_not_an_instance_directory_even_if_it_has_a_submission_dot_xml_file() {
+  public void a_linux_or_mac_hidden_directory_is_not_an_instance_directory_even_if_it_has_a_submission_dot_xml_file() {
     Path correctDir = createDirectories(tempDir.resolve(".DS_Store"));
     write(correctDir.resolve("submission.xml"), Stream.empty());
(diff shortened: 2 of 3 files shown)


yanokwa:
I think only you can launch a rebuild.

ggalmazor:
CircleCI complains about not having write permission...

ggalmazor:
Nevermind, I've just fixed a typo in one commit. That will launch the build again

test/java/org/opendatakit/briefcase/reused/UncheckedFilesIsInstanceDirTest.java
```diff
@@ -55,13 +55,6 @@ public void a_file_is_not_an_instance_directory() {
     assertThat(isInstanceDir(someFile), is(false));
   }
 
-  @Test
-  public void an_arbitrary_named_directory_is_not_an_instance_directory() {
-    assertThat(isInstanceDir(createDirectories(tempDir.resolve("cocotero"))), is(false));
```
yanokwa: Odd choice of variable name, but I guess this is a Spanish thing?

src/org/opendatakit/briefcase/reused/UncheckedFiles.java
```diff
@@ -218,11 +218,9 @@ public static long checksumOf(Path file) {
 
   public static boolean isInstanceDir(Path dir) {
     return Files.isDirectory(dir)
-        // Instance directories follow the pattern "uuid01234567-0123-0123-0123-012345678901"
-        // ("uuid" followed by a UUID string)
-        // This is decided when forms are pulled and written into the storage directory
-        // TODO Extract into an artifact, test, and use everywhere where this convention is used
-        && dir.getFileName().toString().matches("^uuid[\\w]{8}-[\\w]{4}-[\\w]{4}-[\\w]{4}-[\\w]{12}$")
+        // Ignore hidden max/linux hidden folders
```
yanokwa: s/max/mac

test/java/org/opendatakit/briefcase/reused/UncheckedFilesIsInstanceDirTest.java
ggalmazor: Are you referring to "cocotero"? I can change that to something more neutral. I use cocotero when I don't want to think about a name and it doesn't really matter but I try to change that after. This one slipped through.

ggalmazor: OK, changed it. 

By the way, cocotero means coconut tree.

kkrawczyk123:
Looks good on Ubuntu but MacOS and Windows don't allow me to create `.DS_Store` directory how can I cheat on them :) ?

ggalmazor:
> Looks good on Ubuntu but MacOS and Windows don't allow me to create .DS_Store directory how can I cheat on them :) ? 

I think there's no need to test that on windows, since the filter is especifically for hidden folders in mac/linux. As far as I know, Windows hosts don't create such folders that could get in the way and produce a failure.

kkrawczyk123:
Ok, sure.

PR #542, opened 2018-06-26
Issue 504 recover from moved storage dir failure

ggalmazor:
Closes #504

#### What has been done to verify that this works as intended?
Set the conditions that would produce an error by moving the already saved briefcase dir location to somewhere else and run Briefcase. Verified that I got the welcome message and no errors, as expected.

#### Why is this the best possible solution? Were any other approaches considered?
This is the narrowest fix to implement a solution discussed in #504 

#### Are there any risks to merging this code? If so, what are they?
Nope.

#### Does this change require updates to documentation? If so, please file an issue at https://github.com/opendatakit/docs/issues/new and include the link below.
Nope.

PR #549, opened 2018-06-28
Issue 547 feedback on cli ops about missing storage dir

### LINKED ISSUE #547: Avoid NoSuchFileException by building cache.ser if it doesn't exist
https://sentry.io/opendatakit/briefcase/issues/591991089/

```
UncheckedIOException: java.nio.file.NoSuchFileException: <root>/ODK Briefcase Storage/cache.ser
    at org.opendatakit.briefcase.reused.UncheckedFiles.createFile(UncheckedFiles.java:133)
    at org.opendatakit.briefcase.util.FormCache.setLocation(FormCache.java:80)
    at org.opendatakit.briefcase.util.FormCache.from(FormCache.java:54)
    at org.opendatakit.briefcase.operations.ImportFromODK.importODK(ImportFromODK.java:52)
    at org.opendatakit.briefcase.operations.ImportFromODK.lambda$static$1(ImportFromODK.java:42)
...
(9 additional frame(s) were not displayed)

NoSuchFileException: <root>/ODK Briefcase Storage/cache.ser
    at org.opendatakit.briefcase.reused.UncheckedFiles.createFile(UncheckedFiles.java:131)
...
(15 additional frame(s) were not displayed)

Error
```

ggalmazor:
Closes #547

This PR brings to the CLI the same filtering we introduced in #542 for the main UI window. The difference is that, while the UI recovers from a missing storage directory by resetting its saved location, the CLI now will exit on error status and give the user some feedback

This PR also reduces the log's verbosity by ignoring the stacktrace from an EOF error while reading the cache file, which we can't really do much about.

Run the four CLI ops with a storage directory argument pointing to a non-existing path in my filesystem. Verified that the CLI op fails and gives the expected feedback.

To test the change on the logs, truncate the cache file with `truncate -s0 cache.ser` (mac/linux). Then run a CLI op and verify that the logs get just one line explaining that there has been an EOF.

This is a straightforward fix, similar to what we do in the UI.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/operations/Common.java
+++ b/src/org/opendatakit/briefcase/operations/Common.java
@@ -15,6 +15,11 @@
  */
 package org.opendatakit.briefcase.operations;
 
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.opendatakit.briefcase.model.BriefcasePreferences;
+import org.opendatakit.briefcase.reused.BriefcaseException;
 import org.opendatakit.common.cli.Param;
 
 class Common {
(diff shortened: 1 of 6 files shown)


lognaturel:
Closing in favor of #554.

PR #552, opened 2018-06-28
Issue 551 ignore hidden dirs updating cache

### LINKED ISSUE #551: Ignore hidden folders while updating the Form Cache
https://sentry.io/opendatakit/briefcase/issues/580431067/

```
FileNotFoundException: /Users/nmcp/Documents/MutasaData/ODK Briefcase Storage/forms/.DS_Store/.DS_Store.xml (Not a directory)
    at org.opendatakit.briefcase.util.FileSystemUtils.getMd5Hash(FileSystemUtils.java:344)
    at org.opendatakit.briefcase.util.FormCache.lambda$null$1(FormCache.java:115)
    at org.opendatakit.briefcase.util.FormCache.lambda$update$2(FormCache.java:112)
    at org.opendatakit.briefcase.util.FormCache.update(FormCache.java:110)
    at org.opendatakit.briefcase.util.FormCache.onPullSuccess(FormCache.java:145)
...
(32 additional frame(s) were not displayed)

No File
```

ggalmazor:
Closes #551

It's not clear how to reproduce this issue. On Linux, creating a `.DS_Store` won't produce any error (maybe it's dependant on system and jvm version).

Nevertheless, before this PR, directories that weren't form dirs where implicitly being filtered out by the `foreach` block's code. Now an explicit filter is taking care of it, making the whole process more clear to new eyes.

This PR also forces an update of the cache after the cache file is created in the `FormCache` file. This will let Briefcase recover from situations where the cache file has been removed by the user or it has somehow become corrupt.

Tested on Linux by creating a `.DS_Store` dir inside the `forms` dir and removing the cache file. Verified that no errors were present on the logs and that all forms in the storage dir where shown in the export tab.

Same results on mac.

This is a straightforward fix to bring the same filtering ideas we use in other places.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/reused/UncheckedFiles.java
+++ b/src/org/opendatakit/briefcase/reused/UncheckedFiles.java
@@ -216,6 +216,15 @@ public class UncheckedFiles {
     }
   }
 
+  public static boolean isFormDir(Path dir) {
+    String dirName = dir.getFileName().toString();
+    return Files.isDirectory(dir)
+        // Ignore hidden mac/linux hidden folders
+        && !dirName.startsWith(".")
+        // Check for presence of the blank form
+        && Files.exists(dir.resolve(dirName + ".xml"));
+  }
+
   public static boolean isInstanceDir(Path dir) {
     return Files.isDirectory(dir)
         // Ignore hidden mac/linux hidden folders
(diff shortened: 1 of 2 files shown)


ggalmazor:
Sentry issue: [BRIEFCASE-98](https://sentry.io/opendatakit/briefcase/issues/579123037/)

(solved by this PR)

kkrawczyk123:
**Tested with success!**
Verified on Ubuntu.

PR #554, opened 2018-06-28
Issue 553 make cli less picky about storage dir

### LINKED ISSUE #547: Avoid NoSuchFileException by building cache.ser if it doesn't exist

### LINKED ISSUE #553: CLI shouldn't be so picky about the storage dir
#### Software versions 
Briefcase >v1.10.x

#### Problem description
We want to be rigorous and explicit, but previous versions were more tolerant. We chose to be restrictive because it felt wrong that Briefcase would go creating storage dirs in the user's filesystem.

It turns out that many users rely on this to pull forms using CLI ops and it could represent a burden for them to have actual full storage directories to be able to run those commands.

Briefcase CLI ops could create and prepare storage dirs instead of failing.

Run a CLI op with a `-sd` that would produce a non existing location for the briefcase dir.

#### Expected behavior
Gracefully recover instead of failing.

#### Other information 
N/A

ggalmazor:
Closes #553, closes #547 

This PR should be considered as an alternative behavior to the one proposed in  #549

Thir PR will imitate the old Briefcase behavior that will create the storage dir if the user adds a `-sd` arg that will produce a briefcase dir that doesn't exist.

Run all 4 CLI ops with an `-sd` arg that would produce a non-existent briefcase dir. Verify that user gets an error feedback message but the op completes. Verify that the briefcase dir exists afterward.

This is an alternative to failing, which is proposed in #549 
This also is what Briefcase actually did in <=1.9.0

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/operations/Common.java
+++ b/src/org/opendatakit/briefcase/operations/Common.java
@@ -15,6 +15,11 @@
  */
 package org.opendatakit.briefcase.operations;
 
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.opendatakit.briefcase.model.BriefcasePreferences;
+import org.opendatakit.briefcase.reused.UncheckedFiles;
 import org.opendatakit.common.cli.Param;
 
 class Common {
@@ -23,4 +28,13 @@ class Common {
   static final Param<String> ODK_USERNAME = Param.arg("u", "odk_username", "ODK Username");
   static final Param<String> ODK_PASSWORD = Param.arg("p", "odk_password", "ODK Password");
   static final Param<String> AGGREGATE_SERVER = Param.arg("url", "aggregate_url", "Aggregate server URL");
+
+  static Path getOrCreateBriefcaseDir(String storageDir) {
+    Path briefcaseDir = BriefcasePreferences.buildBriefcaseDir(Paths.get(storageDir));
+    if (!Files.exists(briefcaseDir)) {
+      System.err.println("The directory " + briefcaseDir.toString() + " doesn't exist. Creating it");
+      UncheckedFiles.createBriefcaseDir(briefcaseDir);
+    }
+    return briefcaseDir;
+  }
 }
--- a/src/org/opendatakit/briefcase/operations/Export.java
+++ b/src/org/opendatakit/briefcase/operations/Export.java
@@ -20,7 +20,6 @@ import static org.opendatakit.briefcase.export.ExportForms.buildExportDateTimePr
 import static org.opendatakit.briefcase.operations.Common.FORM_ID;
 import static org.opendatakit.briefcase.operations.Common.STORAGE_DIR;
 
-import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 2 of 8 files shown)


lognaturel:
I would be in favor of this approach. It feels more likely to get the user on the right track than #547. The only slight downside I'm seeing is the case I sometimes run into -- I put in `<root>/ODK Briefcase Storage` rather than just `<root>`. This behavior means that I then end up with two nested storage folders. But really that's pretty easy to recover from.

ggalmazor:
Yes, I think you're right. I don't like the idea of Briefcase creating storage directories left and right in the user's computer, but latest feedback makes me think it's a better tradeoff than crashing.

lognaturel:
Yes, absolutely. In fact, #549 won't work because often the reason we use the CLI is because we're running headless and launching the GUI is just not possible.

Do we think this should go through testing? I'm comfortable merging as-is so we can get another build out since we know it's affecting real users -- https://forum.opendatakit.org/t/issues-with-cli-for-odk-briefcase-v1-11-1/13774

In general, how about adding a case to the testing checklist to run the CLI without an ODK Briefcase dir to make sure that behavior is preserved?

yanokwa:
@kkrawczyk123 Can you please add test case for this?

kkrawczyk123:
@yanokwa sure I will.
I've tested it and works well when pulling from gae aggregate. I have a problem with using cmd with my tomcat instance, and I am not sure if it's connected to my instance or it's some aggregate's regression. 
@ggalmazor Could you please check on your local instance?

ggalmazor:
Thanks, @kkrawczyk123! I'll try local tomcat and see :)

kkrawczyk123:
After very long investigation with @ggalmazor I am no longer able to reproduce the issue that I had when using cmd and local aggregate's instance. It might have been connected to some server issue, or some unwanted changes in my briefcase directory or just storage directory.
Tested pulling when sd existed and when did not exist with gae and tomcat instances on Ubuntu , Mac and Windows. 
I added test cases to briefcase testing plan to be sure if situation above won't happen again or if will we have to figure out what it's caused by.

PR #561, opened 2018-06-30
WIP - Spike export submission date parser

ggalmazor:
This PR is just to compare notes and discuss export improvements and optimization strategies.

### DIFF
--- a/src/org/opendatakit/briefcase/export/ExportToCsv.java
+++ b/src/org/opendatakit/briefcase/export/ExportToCsv.java
@@ -16,39 +16,24 @@
 
 package org.opendatakit.briefcase.export;
 
-import static java.nio.file.StandardOpenOption.APPEND;
-import static java.nio.file.StandardOpenOption.CREATE;
-import static java.nio.file.StandardOpenOption.TRUNCATE_EXISTING;
-import static java.util.Comparator.comparing;
-import static org.opendatakit.briefcase.export.CsvMapper.getMainHeader;
-import static org.opendatakit.briefcase.export.CsvMapper.getMainSubmissionLines;
-import static org.opendatakit.briefcase.export.CsvMapper.getRepeatCsvLine;
-import static org.opendatakit.briefcase.export.CsvMapper.getRepeatHeader;
+import static java.util.stream.Collectors.groupingByConcurrent;
+import static java.util.stream.Collectors.reducing;
+import static java.util.stream.Collectors.toList;
 import static org.opendatakit.briefcase.export.ExportOutcome.ALL_EXPORTED;
 import static org.opendatakit.briefcase.export.ExportOutcome.ALL_SKIPPED;
 import static org.opendatakit.briefcase.export.ExportOutcome.SOME_SKIPPED;
+import static org.opendatakit.briefcase.export.SubmissionParser.getListOfSubmissionFiles;
+import static org.opendatakit.briefcase.export.SubmissionParser.parseSubmission;
 import static org.opendatakit.briefcase.reused.UncheckedFiles.createDirectories;
-import static org.opendatakit.briefcase.reused.UncheckedFiles.walk;
-import static org.opendatakit.briefcase.reused.UncheckedFiles.write;
-import static org.opendatakit.briefcase.util.StringUtils.stripIllegalChars;
 
-import java.nio.file.Files;
 import java.nio.file.Path;
-import java.time.LocalTime;
-import java.time.OffsetDateTime;
-import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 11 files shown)


ggalmazor:
@dcbriccetti, it would be great if you could review both commits.

ggalmazor:
The commit 579923bb555a4eca400f004c260c5553a63962e2 brings sorting and writing at the end of the process. 

It still doesn't pass all the tests (some rogue newlines and misteriously missing lines in the date range test) but it shows promising results (7 seconds on SSD, in contrast to the 21 seconds of the previous version)

dcbriccetti:
in UncheckedFiles, I suggest
```
  public static boolean isInstanceDir(Path dir) {
    return Files.isDirectory(dir) && !dir.toFile().isHidden()
        && Files.exists(dir.resolve("submission.xml"));
  }
```

src/org/opendatakit/briefcase/export/ExportToCsv.java
```diff
@@ -166,13 +110,86 @@ public static ExportOutcome export(FormDefinition formDef, ExportConfiguration c
...
+          )
+      );
+      return new OutputFile(
+          getMainHeader(formDefinition.getModel(), formDefinition.isFileEncryptedForm()),
+          output,
+          configuration.getOverwriteExistingFiles().orElse(false),
+          submissionMapper
+      );
+    }
+
```
dcbriccetti: I’m curious why you prefer the separate Function object (left) to the in-line form (right).

screen shot 2018-06-30 at 4 02 03 pm

dcbriccetti:
About counting the files in ExportToCsv.export:
- It takes a long time for the many instances. 15 seconds or more for me on SSD. Is knowing the count beforehand only for the progress information?

- Should we use try-with-resources?
```
  public static ExportOutcome export(FormDefinition formDef, ExportConfiguration configuration, boolean exportMedia) {
    long start = System.nanoTime();
    // Create an export tracker object with the total number of submissions we have to export
    long submissionCount;
    try (Stream<Path> dirStream = walk(formDef.getFormDir().resolve("instances"))) {
      submissionCount = dirStream.filter(UncheckedFiles::isInstanceDir).count();
    }
    ExportProcessTracker exportTracker = new ExportProcessTracker(formDef, submissionCount);
    exportTracker.start();
```

dcbriccetti:
Summary: looks great! Still runs in 300 MB for the near-50,000 instances export.
Changes I made while going through the code: https://github.com/dcbriccetti/briefcase/commit/00e0f42d2588b27f5fa849280358f35fb8b1d8f6
(all noted above)

ggalmazor:
Hi, @dcbriccetti!

Thanks for your feedback! I've just finished what I initially wanted to do with the code and now I'll review your commit.

Now the code passes all the tests and works well from the UI too. I get 7 seconds on SSD (300% performance increase) and 5' on a mechanical drive (200% performance increase). Now I'm wondering how does the new code perform memory-wise. I'll try to make sense of YourKit to figure it out, but I'll welcome your expert eyes on this too ;)

ggalmazor:
Just run a quick test with the old code and this PR is faster by almost half of time: 7 seconds vs 11 seconds in my machine (SSD drive)

ggalmazor:
@dcbriccetti,

I'm getting this memory usage graph with the current master version:

image

And this one with thir PR:

image

Does that make sense?

I think now we require a little more memory because now we are collecting the lines to sort them, instead of sorting paths to files, which seem lighter.

Also, I remember you did some tests constraining the memory the process could use... How did you do that?

dcbriccetti:
From my Gradle notes:
- Set JVM options for a run program
	tasks.withType(JavaExec) {
	    jvmArgs = ['-Xmx300m']
	}

ggalmazor:
With a 300MB of memory limit I'm getting this:
image

dcbriccetti:
More frequent garbage collection. Did you see that YourKit Profiler also shows you the GC pause durations? That’s nice to know.

ggalmazor:
I see!

47 pauses (unconstrained) vs 757 pauses (constrained). That makes sense

dcbriccetti:
The lengths of the pauses are important, too, of course. Really clamp down the memory and see what fun ensues! Eventually the JVM says, “The heck with this!”.

ggalmazor:
Now I'm wondering. Are these good results? I feel like they are: performance is better and it seems to work well under memory constrained scenarios.

Do you think we should test anything special before considering working on a PR with nice commits?

dcbriccetti:
This is ready to implement. As they would say in some parts of the U.S.: “We done good.”

ggalmazor:
Thanks, @dcbriccetti!

ggalmazor:
Closing in favor of #571

PR #563, opened 2018-07-02
Issue 558 Attach causing exception to domain exception

### LINKED ISSUE #558: Log cause of error
We're losing the cause exception for this error, thus, we are unable to know if it's a bug on our side or a third-party error.

PullFromServerException: Failure pulling forms from server. FormIds: ~redacted~

https://sentry.io/opendatakit/briefcase/issues/539689428/

```
PullFromServerException: Failure pulling forms from server. FormIds: ~redacted~
    at org.opendatakit.briefcase.util.TransferFromServer.pull(TransferFromServer.java:67)
    at org.opendatakit.briefcase.operations.PullFormFromAggregate.pullFormFromAggregate(PullFormFromAggregate.java:71)
    at org.opendatakit.briefcase.operations.PullFormFromAggregate.lambda$static$0(PullFormFromAggregate.java:45)
    at org.opendatakit.common.cli.Cli.lambda$run$4(Cli.java:119)
    at org.opendatakit.common.cli.Cli.run(Cli.java:116)
...
(7 additional frame(s) were not displayed)

Error
```

ggalmazor:
Closes #558

This PR doesn't change behavior.

This PR introduces a technical change to add information to an error log in order to try to know if there's a bug behind it or not.

It's not clear how to reproduce the error that would let us know if the change works, but it's a really straightforward change and it doesn't affect to any other part of Briefcase. It should be safe to merge. 

This PR follows the good practice to not losing the causing exception when wrapping it in a domain exception.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/util/PullFromServerException.java
+++ b/src/org/opendatakit/briefcase/util/PullFromServerException.java
@@ -22,6 +22,10 @@ import org.opendatakit.briefcase.model.FormStatus;
 import org.opendatakit.briefcase.reused.BriefcaseException;
 
 public class PullFromServerException extends BriefcaseException {
+  public PullFromServerException(List<FormStatus> forms, Throwable cause) {
+    super("Failure pulling forms from server. FormIds: " + forms.stream().map(f -> f.getFormDefinition().getFormId()).collect(joining(", ")), cause);
+  }
+
   public PullFromServerException(List<FormStatus> forms) {
     super("Failure pulling forms from server. FormIds: " + forms.stream().map(f -> f.getFormDefinition().getFormId()).collect(joining(", ")));
   }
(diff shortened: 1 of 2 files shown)


src/org/opendatakit/briefcase/util/PullFromServerException.java
```diff
@@ -22,6 +22,10 @@
 import org.opendatakit.briefcase.reused.BriefcaseException;
 
 public class PullFromServerException extends BriefcaseException {
+  public PullFromServerException(List<FormStatus> forms, Throwable cause) {
+    super("Failure pulling forms from server. FormIds: " + forms.stream().map(f -> f.getFormDefinition().getFormId()).collect(joining(", ")), cause);
```
dcbriccetti: Hmm, haven’t I seen this first argument somewhere before? :-)

ggalmazor: I'm thinking about extracting that construction to its own private method? Would that be acceptable?

ggalmazor:
I've extracted a method to reduce duplication.

PR #564, opened 2018-07-02
Issue 530 silence sentry logs

### LINKED ISSUE #530: Silence some of the error reports sent to Sentry
#### Software versions 
Briefcase v1.11.0 beta

#### Problem description
We have sent to Sentry many errors that are not really actionable (user input errors, third party service errors, etc.)

N/A

#### Expected behavior
Briefcase should only send error reports from possible/probable bugs only. Just actionable reports.

#### Other information 
N/A

ggalmazor:
Closes #530

This PR doesn't change behavior

This PR makes some technical changes to silence some of the reports that get to Sentry. Now, only non-caught exceptions that reach the Launcher (topmost class) should be sent as an error report to Sentry.

Verified that CLI ops and UI launch work as expected.
Introduced a syntactic error in a submission to produce an export error and verified that the error was logged but no error reports were sent to Sentry (it is caught and handled by the Export panel).
Forced a non-caught error by adding `throw new RuntimeException("cocotero")` to the last line on the `try` block of the `Cli.run()` method and verified that an error report was sent to Sentry (tracking consent must be checked in the Settings Tab).

This solution has been discussed in #530 as a good solution to reduce the number of Sentry reports we need to comb in search for bugs.

The implementation also improves the design by uncoupling the Cli from our domain's dependencies, as it should be.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/Launcher.java
+++ b/src/org/opendatakit/briefcase/Launcher.java
@@ -51,22 +51,8 @@ public class Launcher {
     if (!appPreferences.hasKey(BRIEFCASE_TRACKING_CONSENT_PROPERTY))
       appPreferences.put(BRIEFCASE_TRACKING_CONSENT_PROPERTY, TRUE.toString());
 
-    if (SENTRY_ENABLED) {
-      Sentry.init(String.format(
-          "%s?release=%s&stacktrace.app.packages=org.opendatakit&tags=os:%s,jvm:%s",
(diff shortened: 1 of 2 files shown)


yanokwa:
@ggalmazor There was a minor misspelling in the first commit. I fixed (s/catched/caught) and rebased.

kkrawczyk123:
**Tested wit success!**
Verified on Ubuntu.

PR #565, opened 2018-07-02
Issue 413 ui disable export media

### LINKED ISSUE #413: Provide an option to disable exporting media files on the UI
The CLI export operation supports disabling the export of media file attachments to submissions.

We should add a new export configuration option to disable on the UI as well.

ggalmazor:
Closes #413 

Tests passing.

Tested all combinations of the new setting with the UI, both as default and override confs. Verified that media files are exported only when the option is checked.

Verified that the export media checkbox is checked by default after cleaning prefs.

Verified that the export media checkbox state was saved after relaunching Briefcase (both default and override).

The changes on this PR imitate what's being done with other conf params.

We should refactor the export conf object and related UI components to use a more up-to-date approach like in the pull/push tabs that would dramatically reduce code duplication and complexity.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/export/ExportConfiguration.java
+++ b/src/org/opendatakit/briefcase/export/ExportConfiguration.java
@@ -70,7 +70,7 @@ public class ExportConfiguration {
   private Optional<Boolean> overwriteExistingFiles;
   private Optional<Boolean> exportMedia;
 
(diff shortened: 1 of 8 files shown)


kkrawczyk123:
I have noticed a minor issue reported in https://github.com/opendatakit/briefcase/issues/580

PR #566, opened 2018-07-02
Issue 479 reload cache button

### LINKED ISSUE #479: During first briefcase run push and export list shouldn't depend on pull.

ggalmazor:
Closes #479

This PR adds a new button to the Settings tab that lets the user reload the form cache. This is useful when the user needs to manually add or remove any form in Briefcase's current storage directory.

image

- Set a storage directory. 
- Check that a form is listed in the push and export tabs.
- Remove the form folder manually
- Click on the reload cache button (verify that a dialog confirms the operation)
- Check that the form is no longer present in the push and export tabs.

This PR introduces UI changes only. All the underlying mechanisms already existed. It's a fairly straightforward change.

Nope.

Nope.


yanokwa:
Hmm. I wonder if this would be simpler right below the Storage Directory folder. And we could call it `Refresh Cache` or maybe `Rebuild Cache`. What do you think of this UI, @lognaturel?

cache

lognaturel:
I don't have a strong preference between the two placement options. What about trying to come up with wording that would make sense to a non-technical user? Maybe something like "Reload forms from disk"?

yanokwa:
Yeah, I guess we have space, so `Reload forms from storage location`

ggalmazor:
Alright, this is what it looks like now: 

image

dcbriccetti:
I like this latest UI, except:
I’m not used to seeing JButtons stretched wide like that.
“The form cache has been successfully reloaded”: how about rebuilt instead of reloaded? Loading to me suggests bringing something from disk into memory.

src/org/opendatakit/briefcase/ui/settings/SettingsPanel.java
```diff
@@ -60,6 +61,10 @@ public SettingsPanel(SettingsPanelForm form, BriefcasePreferences appPreferences
       analytics.enableTracking(enabled, false);
     });
     form.onHttpProxy(appPreferences::setHttpProxy, appPreferences::unsetHttpProxy);
+    form.onReloadCache(() -> {
+      formCache.update();
+      JOptionPane.showMessageDialog(getContainer(), "The form cache has been successfully reloaded");
```
dcbriccetti: I prefer rebuilt to reloaded.

ggalmazor:
@yanokwa, @lognaturel, @dcbriccetti, I have no strong opinion about "rebuild" or "reload". Please, confirm what you want and I'll proceed :)

yanokwa:
I prefer rebuilt to reload.

ggalmazor:
OK, I've updated the settings tab with a troubleshooting section at the bottom.

kkrawczyk123:
**Tested with success!**

This is such a great functionality! The only one suggestion to functionality is button should be enabled when storage directory is chosen and disabled if it not. I have already discussed it with @ggalmazor and it will be fixed by https://github.com/opendatakit/briefcase/pull/578
And my suggestion to GUI is reported in https://github.com/opendatakit/briefcase/pull/578

PR #570, opened 2018-07-09
Fix logger class name

dcbriccetti:
A classic copy-and-paste error


ggalmazor:
Thanks, @dcbriccetti!! It seems that you forgot to remove the unused import. Sorry!

dcbriccetti:
A classic newbie error! (Mine)

PR #571, opened 2018-07-09
Issue 569 export performance improvement

### LINKED ISSUE #569: Improve performance of the export operation
#### Software versions 
Briefcase v1.11.2, Java v1.8.171

#### Problem description
Due to the latest changes to the export operation's code, we've experienced some performance loss related to how Briefcase needs to hit the filesystem to get the submission files, parse their submission date and produce an ordered output.

Export a form with a large set of submissions like [this one](https://drive.google.com/open?id=1czxbJk3dIw3CY1AGlTlYozwyyKLNK_q8) with 50k submissions.

#### Expected behavior
Exporting forms should be at least as fast as in previous versions of Briefcase

In my system, the attached form gets currently exported in 23 seconds using an SSD drive and up to 10 minutes using a mechanical drive.

Previous versions of Briefcase would export it in 12 seconds.

#### Other information 
N/A

ggalmazor:
Closes #569 

This PR doesn't change Briefcase's behaviour.

This PR comes with technical improvements that make the export operation run faster (even faster than v1.9.0) with approximately the same memory consumption.

The changes from this PR have been discussed in #561 and have been developed with the support of the test suite that verifies the export results.

I've used a form with 50k submissions (attached to #569) to time the export operation and measure memory consumption.

Now the export operation runs under 9 seconds (in my computer) and allocates 2.8GB of memory (like v1.9.0). It also completes the operation in memory constrained scenarios (tested with -Xmx300m).

This PR changes the output sorting strategy: now we only sort the CSV lines of the main file, just before we need to write them to a file. By maximizing the number of operations we fit into unsorted Java8 streams we get the most of the system's computing and I/O power. 

Before this, we were focusing on getting the list of submissions files sorted by submission date, which forced a sorted stream operation from the beginning of the export operation and wasn't performing well.

No.

No.


yanokwa:
@ggalmazor Can you please solve the merge conflicts? I'm the worst at merging out of order 😭

kkrawczyk123:
Wow! I have to admit that it has been a huge performance improvement:
my results:
Ubuntu: 14 s
Mac: 32 s
Windows: 6 min (but it still is much faster than before that changes) and it is a vm.

lognaturel:
That's a great result, thanks @kkrawczyk123, @ggalmazor!

yanokwa:
@ggalmazor Accroding to https://github.com/opendatakit/briefcase/pull/457, the pre-rewrite performance (v1.10) was about 17 seconds. It's now about 9 seconds. That's about a 60% increase in performance. That sound correct? I need to confirm for when I write release notes...

ggalmazor:
Previous versions were somewhat unpredictable, especially on mechanical drives. Also, I don't really remember the details of the benchmarking rigs I used.

Just to be sure I've run a set of 10 exports with each version and the results are 8.7 seconds (master) vs 11.35 seconds (v.1.10.0). That's a 30% increase in performance.

Issue #556, opened 2018-06-29
Investigate why so many Duplicate keys while rebuilding indexes


ggalmazor:
https://sentry.io/opendatakit/briefcase/issues/512538494/

```
IllegalStateException: Duplicate key org.opendatakit.briefcase.model.FormStatus@5de5e95
    at org.opendatakit.briefcase.push.PushForms.rebuildIndex(PushForms.java:91)
    at org.opendatakit.briefcase.push.PushForms.<init>(PushForms.java:35)
    at org.opendatakit.briefcase.ui.push.PushPanel.from(PushPanel.java:104)
    at org.opendatakit.briefcase.ui.MainBriefcaseWindow.<init>(MainBriefcaseWindow.java:122)
    at org.opendatakit.briefcase.ui.MainBriefcaseWindow.launchGUI(MainBriefcaseWindow.java:81)
...
(20 additional frame(s) were not displayed)

Failed to launch GUI
```

ggalmazor:
Closing since this has been already analyzed and we have agreed on implementing a central database to solve the indexing of forms

PR #575, opened 2018-07-10
Issue 556 name as form index key

ggalmazor:
Partially addresses #556

This PR comes with some technical changes to avoid getting the duplicated keys error some users have been experiencing.

This PR also fixes two bugs:
- The pull form definition option wouldn't create an empty instances directory. This would produce an error when exporting a form pulled this way.
- The forms list for pull/push panels wouldn't rebuild indexes after loading a list of forms. This could produce inconsistencies between in memory and on screen data.

- Pull the forms in this package [forms.zip](https://github.com/opendatakit/briefcase/files/2180240/forms.zip)
  (There are two forms with the same form ID but different titles)
- Verify that both forms are shown in the push and export tables (`Simple Form` and `Simple Form 2`)
- Close Briefcase
- Remove the `cache.ser` file
- Open Briefcase
- Verify that both forms are shown in the push and export tables
- Verify that no error is logged during the whole process.

This still allows errors when Briefcase finds two forms with the same ID and title in different storage dir folders, which could only be possible by direct manual manipulation by the user.

Replacing form IDs as keys to identify forms in the storage dir seems the best option we currently have, without making Briefcase much more complex by using composite keys.

Nope, but any custom export configuration will be lost, as the pref keys were build using the form ID, not the form name.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/export/ExportForms.java
+++ b/src/org/opendatakit/briefcase/export/ExportForms.java
@@ -28,7 +28,6 @@ import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import org.opendatakit.briefcase.model.BriefcasePreferences;
 import org.opendatakit.briefcase.model.FormStatus;
-import org.opendatakit.briefcase.model.IFormDefinition;
 import org.opendatakit.briefcase.model.ServerConnectionInfo;
 
 public class ExportForms {
@@ -36,16 +35,16 @@ public class ExportForms {
   private static final String CUSTOM_CONF_PREFIX = "custom_";
   private final List<FormStatus> forms;
   private ExportConfiguration defaultConfiguration;
-  private final Map<String, ExportConfiguration> customConfigurations;
+  private final Map<String, ExportConfiguration> customConfigurationsByFormName;
   private final Map<String, LocalDateTime> lastExportDateTimes;
   private final Map<String, ServerConnectionInfo> transferSettings;
   private final List<BiConsumer<String, LocalDateTime>> onSuccessfulExportCallbacks = new ArrayList<>();
   private Map<String, FormStatus> formsIndex = new HashMap<>();
 
-  public ExportForms(List<FormStatus> forms, ExportConfiguration defaultConfiguration, Map<String, ExportConfiguration> configurations, Map<String, LocalDateTime> lastExportDateTimes, Map<String, ServerConnectionInfo> transferSettings) {
+  public ExportForms(List<FormStatus> forms, ExportConfiguration defaultConfiguration, Map<String, ExportConfiguration> customConfigurationsByFormName, Map<String, LocalDateTime> lastExportDateTimes, Map<String, ServerConnectionInfo> transferSettings) {
     this.forms = forms;
     this.defaultConfiguration = defaultConfiguration;
-    this.customConfigurations = configurations;
+    this.customConfigurationsByFormName = customConfigurationsByFormName;
     this.lastExportDateTimes = lastExportDateTimes;
     this.transferSettings = transferSettings;
     rebuildIndex();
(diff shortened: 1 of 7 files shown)


src/org/opendatakit/briefcase/ui/export/ExportPanel.java
```diff
@@ -125,13 +125,13 @@ private void updatePreferences() {
       preferences.putAll(form.getConfPanel().getConfiguration().asMap());
 
     // Clean all custom conf keys
-    forms.forEach(formId ->
-        preferences.removeAll(ExportConfiguration.keys(buildCustomConfPrefix(formId)))
+    forms.forEach(formStatus ->
```
dcbriccetti: `forEach` here is a misnomer, wouldn’t you say? I don’t expect a for each to first map to get a specific property.

src/org/opendatakit/briefcase/export/ExportForms.java
```diff
@@ -117,59 +111,59 @@ public FormStatus get(int rowIndex) {
     return forms.get(rowIndex);
   }
 
-  public void forEach(Consumer<String> callback) {
-    forms.stream().map(ExportForms::getFormId).forEach(callback);
+  public void forEach(Consumer<FormStatus> callback) {
```
dcbriccetti: Can we call this `forEachFormStatus`?

ggalmazor: Wouldn't that be a little redundant, since the signature explicitly expresses that this is a `forEach` of a list of `FormStatus` objects?

Also, the main purpose of this class is to wrap a list of `FormStatus` objects. I do the same pattern with other wrappers that have `.map()` and `.flatMap` methods. If we really need to add suffixes to clarify it's probably because the wrapper does too much stuff and should be split.

ggalmazor:
I'm closing this PR while we do some extra checks.

ggalmazor:
|Test|master|PR|
|---|---|---|
|Same formId<br/>Different title<br/>Test: pull “Form Definition”|Pulls both forms to the storage dir<br/>No errors are logged<br/>The second form is not listed in Push and Export tabs<br/>Briefcase fails to relaunch (dupe error)|Pulls both forms to the storage dir<br/>No errors are logged<br/>Both forms are listed in Push and Export tabs<br/>Briefcase relaunches without failing|
|Different formId<br/>Same title<br/>Test: pull from Aggregate|Pulls both forms but only the last prevails in the storage directory<br/>No errors logged|Fails to get the list of forms from Aggregate.<br/>No feedback is given but Briefcase can resume operation.<br/>Error is logged|
|Different formId<br/>Same title<br/>Test: pull “Form Definition”|Pulls both forms but only the last prevails in the storage directory<br/>No errors logged|Pulls both forms but only the last prevails in the storage directory<br/>No errors logged|
|Same formId<br/>Same title<br/>(manually copied in storage dir)<br/>Test: no cache file, relaunch Briefcase|Briefcase fails to relaunch (dupe error)|Briefcase fails to relaunch (dupe error)|

lognaturel:
Thanks, it's helpful to see the behavior evolution and good to know users are most likely experiencing what they "always" have, we just have better logging now.

While I agree this PR does improve the behavior, I'm still wondering whether it would be better to address this as part of the switch over to a proper database for managing state. I think when we do that, we can use a proper auto-incrementing, 100% unique primary key and use multiple fields (e.g. form_id, source, version or something like that) to determine matches. That would solve the problem more completely and mean users only have to adapt to one behavior change. This is an improvement but it means users who haven't had problems in the past might suddenly start getting some. For example, what if a user has forms `my_survey`/`My Survey` and `my_survey_cool_update`/`My Survey` on their server? This actually seems fairly likely to me.

ggalmazor:
OK! That is very reasonable. Let's sit on this one :)

kkrawczyk123:
After consulting with @ggalmazor it looks like work on it has been stopped so I am removing needs tetsing label.

PR #577, opened 2018-07-12
Issue 572 exploded choice lists

### LINKED ISSUE #572: Add CLI option to explode multiple choice select values on exports
Based on https://github.com/cims-bioko/briefcase/issues/1

> As a data analyst, I want briefcase to export multi-selects as boolean grids similar to KoboToolkit's analyzer export tool, so that I can analyze multi-select questions directly in a statistics tool (Stata, SPSS, etc) without a lot of pre-processing.
> 
> example

ggalmazor:
Closes #572

This PR adds a new export configuration parameter that lets explode choice list fields and get extra columns (one per select choice) with a `1` if the choice was selected or `0` otherwise.
- In the CLI, this feature can be activated with the new CLI param `--explode_choice_lists`, or `-ecl` in its short form. 
- In the UI, it is disabled by default and can't be activated yet.

A form can define choice list fields in two ways:
- Declare a nodeset's binding to be of `type="select"`
- Declare another binding type (`string`, for example) and use a `<select>` element in the form to input values for that field.

Added new tests that cover the new feature
Run the export CLI command with and without the new param and verified that the new columns were there.

We could have just implemented the first kind of choice list fields by adding a new `CsvFieldMapper` implementation but that wouldn't save us from all the adaptations we need to make to know about the actual select choices of a field.

Also, from a design point of view, this lets us introduce the missing concept of field decorators, which let change the output of a field, regardless of its type.

Nope.

Nope.


src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -296,6 +314,10 @@ public void ifExportMediaPresent(Consumer<Boolean> consumer) {
     exportMedia.ifPresent(consumer);
   }
 
+  public void isExplodeChoiceListPresent(Consumer<Boolean> consumer) {
```
dcbriccetti: It seems that this method and the one above are never called.

test/resources/org/opendatakit/briefcase/export/choice-lists.xml
```diff
@@ -0,0 +1,138 @@
+<?xml version="1.0"?>
+<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml">
+  <h:head>
+    <h:title>Choice lists</h:title>
+    <model>
+      <instance>
+        <data id="choice-lists">
+          <some_string/>
+          <!--<some_int/>-->
```
dcbriccetti: Why all the commented-out elements?

src/org/opendatakit/briefcase/export/ExplodeChoiceLists.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import org.opendatakit.briefcase.reused.Pair;
+
+class ExplodeChoiceLists {
+  static CsvFieldMapper decorate(CsvFieldMapper mapper) {
+    return (localId, workingDir, model, element, configuration) -> {
+      List<Pair<String, String>> output = new ArrayList<>();
+      output.addAll(mapper.apply(localId, workingDir, model, element, configuration).collect(toList()));
```
dcbriccetti: IDEA suggests this:
`List<Pair<String, String>> output = mapper.apply(localId, workingDir, model, element, configuration).collect(toList());`

src/org/opendatakit/briefcase/export/Model.java
```diff
@@ -248,4 +257,57 @@ private long size() {
...
+  }
+
+
+  public List<SelectChoice> getChoices() {
+    QuestionDef control = controls.get(fqn());
+    List<SelectChoice> staticChoices = Optional.ofNullable(control.getChoices()).orElse(emptyList());
+    return !staticChoices.isEmpty() ? staticChoices : control.getDynamicChoices().getChoices();
+  }
+
+  // TODO This should be defined in JavaRosa, like the DataType enum
```
dcbriccetti: If we put this here now, do you want to create an issue in JR to move it later? Or move it now?

src/org/opendatakit/briefcase/export/ExplodeChoiceLists.java
ggalmazor: I decided to ignore IntelliJ on this one. I think that it's more clear for any future reader to have an empty list and then some statements that add elements to it as a result of different processes.

src/org/opendatakit/briefcase/export/Model.java
ggalmazor: I'd rather not be blocked by a JR change. Let's do it later.

I've created an issue and I've assigned it to me to not forget about this. https://github.com/opendatakit/javarosa/issues/351

test/resources/org/opendatakit/briefcase/export/choice-lists.xml
ggalmazor: Ugh, I forgot to cleanup that file :P

ggalmazor:
Thanks for your review, @dcbriccetti!

I think I've solved all your concerns (at least the ones from this PR). I've also fixed some rogue COCOTEROs.

src/org/opendatakit/briefcase/export/ExplodeChoiceLists.java
dcbriccetti: OK, I can see that.

kkrawczyk123:
**Tested with success!**

kkrawczyk123:

yanokwa:
I finally got a chance to look at this, @ggalmazor.

* What would really help for all these PRs is a sample form with a few submissions on https://nafundi-test.appspot.com (use test/test) or https://sandbox.aggregate.opendatakit.org.
* I'm not sure `explode_choice_lists` is the best way to describe this for users. Ona calls this `split select multiple` and Kobo uses `split select multiple choice answers`. I think `split_select_multiples` is more user-friendly and consistent.
* Ona/Kobo default to True/False instead of 1/0 (not sure what Kobo does) and I'm curious why that choice was made? If there is a good reason, great, but if not, seems like we should just do the same thing.
* Can someone confirm that that this only changes the behavior of select multiples and it doesn't change select ones? I think that'd be unexpected.
* Despite what the PR description says, I think this probably requires a change to the docs, no?

ggalmazor:
> What would really help for all these PRs is a sample form with a few submissions on nafundi-test.appspot.com (use test/test) or sandbox.aggregate.opendatakit.org.

I've uploaded a pair of new forms `choice-lists` and `choice-lists-dynamic` and pushed a few submissions to nafundi-test.appspot.com.

> I'm not sure explode_choice_lists is the best way to describe this for users. Ona calls this split select multiple and Kobo uses split select multiple choice answers. I think split_select_multiples is more user-friendly and consistent.

I'll change it to `split_select_multiples`

> Ona/Kobo default to True/False instead of 1/0 (not sure what Kobo does) and I'm curious why that choice was made? If there is a good reason, great, but if not, seems like we should just do the same thing.

There is no standard way to encode booleans in CSV files, although apps commonly parse '', 0, 'false' and 'False' as false and 1, 'true', 'True' as true.

The good thing about 0/1 is that it uses 1 char as opposed to potentially 5, which can reduce the output file's size with lots of choices and rows.

> Can someone confirm that this only changes the behavior of select multiples and it doesn't change select ones? I think that'd be unexpected.

I've improved the tests to cover that.

> Despite what the PR description says, I think this probably requires a change to the docs, no?

I didn't do it because we're not currently explaining any export CLI op param. In any case, I think we should explain all params and add some use case examples too.

PR #578, opened 2018-07-12
Fix behavior of reload forms button

ggalmazor:
This PR changes the locking behavior of the reload forms button, which stayed enabled all the time, even when there was no storage dir set.

The new behavior is:
- Disable it by default
- Enable it when the storage directory is set
- Disable it again when the storage directory gets cleared

- Verify it is disabled on first launch after clearing prefs.
- Verify it is enabled after setting the storage dir
- Verify it gets disabled when clearing the storage dir

It's the most straighforward change I could come up with to support the behavior we want.

Nope.

Nope.


kkrawczyk123:
**Tested with success!**
Verified on Ubuntu, MacOS and WIndows.

PR #581, opened 2018-07-13
Issue 580 fix export custom conf export media

### LINKED ISSUE #580: Export media checkbox not saved on Override export configuration.
#### Software versions 
Briefcase v1.11.x, Java v1.x.x, operating system: Ubuntu, MacOS and Windows

#### Problem description
Export media checkbox not saved on Override export configuration. 

1. Uncheck "Export media files" on export tab
2. Click on individual's form override export configuration button 
3. Export media files on that windows is selected by default so click on OK button
4. Take a look on the export tab - individual's form override export configuration button remain black
5. Export that form - media should be exported but they are not
6. Go back to individual's form override export configuration an uncheck export media files and click on ok button.
7. Take a look on the export tab - individual's form override export configuration button now is green
8. Go back to individual's form override export configuration and click on clear all button
and try case: open override, uncheck, check and click ok - now override button is green and export of media works great. 

#### Expected behavior
The individual checkbox should be valid always independently on Overal Export media files on main export tab.

ggalmazor:
Closes #580 

This PR changes the behavior of the custom conf dialogs. Now, the user can choose to inherit whichever setting for export media is set in the default configuration panel or override it to enable or disable it.

These changes imitate what we did to solve the same situation with the "pull before export" setting.

- Wrote new tests for the resolution of effective export media conf param.
- Run cycles of setting some scenario, relaunching Briefcase and verifying that the scenario was correctly saved & restored:

  | Default | Custom | Resolved to |
  |---|---|---|
  |ON|INHERIT|ON|
  |ON|ON|ON|
  |ON|OFF|OFF|
  |OFF|INHERIT|OFF|
  |OFF|ON|ON|
  |OFF|OFF|OFF|

- Verified with the Birds form that media files where exported according to each scenario

This imitates previous solution to the same problem.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/export/ExportMediaOverrideOption.java
+++ b/src/org/opendatakit/briefcase/export/ExportMediaOverrideOption.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2018 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.opendatakit.briefcase.export;
+
+import java.util.Optional;
+import java.util.stream.Stream;
+
+public enum ExportMediaOverrideOption {
+  INHERIT("Use default", null), EXPORT_MEDIA("Export media", true), DONT_EXPORT_MEDIA("Do not export media", false);
+
+  private final String label;
+  private final Optional<Boolean> value;
+
+  ExportMediaOverrideOption(String label, Boolean value) {
+    this.label = label;
+    this.value = Optional.ofNullable(value);
+  }
+
+  public static ExportMediaOverrideOption from(String name) {
+    return Stream.of(values())
+        .filter(value -> value.name().equals(name))
+        .findFirst()
+        .orElseThrow(() -> new IllegalArgumentException("Unknown ExportMediaOverrideOption value " + name));
+  }
+
+  public static ExportMediaOverrideOption from(Optional<Boolean> maybeValue) {
+    return maybeValue.map(value -> value ? EXPORT_MEDIA : DONT_EXPORT_MEDIA).orElse(INHERIT);
+  }
+
+  public Optional<Boolean> asBoolean() {
+    return value;
+  }
+
+  @Override
+  public String toString() {
+    return label;
+  }
+}
--- a/src/org/opendatakit/briefcase/export/ExportMediaOverrideOption.java
+++ b/src/org/opendatakit/briefcase/export/ExportMediaOverrideOption.java
@@ -16,7 +16,6 @@
 package org.opendatakit.briefcase.export;
 
 import java.util.Optional;
-import java.util.stream.Stream;
 
 public enum ExportMediaOverrideOption {
   INHERIT("Use default", null), EXPORT_MEDIA("Export media", true), DONT_EXPORT_MEDIA("Do not export media", false);
@@ -27,17 +26,6 @@ public enum ExportMediaOverrideOption {
   ExportMediaOverrideOption(String label, Boolean value) {
     this.label = label;
     this.value = Optional.ofNullable(value);
-  }
-
-  public static ExportMediaOverrideOption from(String name) {
-    return Stream.of(values())
-        .filter(value -> value.name().equals(name))
-        .findFirst()
-        .orElseThrow(() -> new IllegalArgumentException("Unknown ExportMediaOverrideOption value " + name));
-  }
-
-  public static ExportMediaOverrideOption from(Optional<Boolean> maybeValue) {
-    return maybeValue.map(value -> value ? EXPORT_MEDIA : DONT_EXPORT_MEDIA).orElse(INHERIT);
   }
 
   public Optional<Boolean> asBoolean() {
--- a/test/java/org/opendatakit/briefcase/export/ExportConfigurationTest.java
+++ b/test/java/org/opendatakit/briefcase/export/ExportConfigurationTest.java
@@ -162,13 +162,15 @@ public class ExportConfigurationTest {
 
   @Test
   public void resolves_if_we_need_to_pull_depending_on_a_pair_or_fields() {
-    ExportConfiguration config = empty();
-    assertThat(config.setPullBefore(true).setPullBeforeOverride(INHERIT).resolvePullBefore(), is(true));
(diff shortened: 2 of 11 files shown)


ggalmazor:
@lognaturel,
> Might it be useful to link to some PRs like this from the readme to show examples of things like adding a new setting that can be overridden by a custom config?

Yeah, as the conf panel gets more and more complicated, it will be useful to explain how the overriding of conf params works. I'll try to summarize it here for later use in the README:
- Any param set in the custom configuration of a form has precedence over the same field in the default configuration.
- All conf params are resolved to their effective values combining the default and custom configurations of a form before exporting.
- Text configuration params like the export dir, PEM file location, and the date filters, which take some text input from the user, have the same UI field controls in the default and custom configuration panels.
  - Leaving a field empty in a custom configuration param will defer its value to the same param in the default configuration.
  - This table summarizes how the effective values of these params are computed:




- YES/NO configuration params like the "export media" or "pull before export" have different field controls:
  - In the default configuration, they have checkboxes: checked means YES and unchecked means NO
  - In the custom configuration, they have dropdown select boxes with the following choices:
    - INHERIT, meaning "take whatever is defined in the default configuration". This is always the default value here. 
    - YES, meaning "override whatever is defined in the default configuration with YES"
    - NO, meaning "override whatever is defined in the default configuration with NO"

- These configurations can be set or unset 
  -

src/org/opendatakit/briefcase/export/ExportMediaOverrideOption.java
```diff
@@ -0,0 +1,51 @@
...
+
+  private final String label;
+  private final Optional<Boolean> value;
+
+  ExportMediaOverrideOption(String label, Boolean value) {
+    this.label = label;
+    this.value = Optional.ofNullable(value);
+  }
+
+  public static ExportMediaOverrideOption from(String name) {
```
dcbriccetti: Do we need this method?

```
out.println(Arrays.toString(ExportMediaOverrideOption.values()));
out.println();

// “valueOf” does the job
out.println(ExportMediaOverrideOption.valueOf("INHERIT").name());
out.println(ExportMediaOverrideOption.valueOf("INHERIT"));
try {
  out.println(ExportMediaOverrideOption.valueOf("x"));
} catch (IllegalArgumentException e) {
  out.println(e.toString());
}

// We don’t need this “from”
out.println();
out.println(ExportMediaOverrideOption.from("INHERIT").name());
out.println(ExportMediaOverrideOption.from("INHERIT"));
try {
  out.println(ExportMediaOverrideOption.from("x"));
} catch (IllegalArgumentException e) {
  out.println(e.toString());
}
```
```
[Use default, Export media, Do not export media]

INHERIT
Use default
java.lang.IllegalArgumentException: No enum constant org.opendatakit.briefcase.export.ExportMediaOverrideOption.x

INHERIT
Use default
java.lang.IllegalArgumentException: Unknown ExportMediaOverrideOption value x
```

src/org/opendatakit/briefcase/export/ExportMediaOverrideOption.java
```diff
@@ -0,0 +1,51 @@
...
+    return Stream.of(values())
+        .filter(value -> value.name().equals(name))
+        .findFirst()
+        .orElseThrow(() -> new IllegalArgumentException("Unknown ExportMediaOverrideOption value " + name));
+  }
+
+  public static ExportMediaOverrideOption from(Optional<Boolean> maybeValue) {
```
dcbriccetti: Are we calling this anywhere?

test/java/org/opendatakit/briefcase/export/ExportConfigurationTest.java
```diff
@@ -171,6 +171,19 @@ public void resolves_if_we_need_to_pull_depending_on_a_pair_or_fields() {
     assertThat(config.setPullBefore(false).setPullBeforeOverride(DONT_PULL).resolvePullBefore(), is(false));
   }
 
+  @Test
+  public void resolves_if_we_need_to_export_media_files_depending_on_a_pair_or_fields() {
+    assertThat(empty().resolveExportMedia(), is(true));
+    assertThat(empty().setExportMedia(true).resolveExportMedia(), is(true));
+    assertThat(empty().setExportMedia(false).resolveExportMedia(), is(false));
+    assertThat(empty().setExportMedia(true).setExportMediaOverride(ExportMediaOverrideOption.INHERIT).resolveExportMedia(), is(true));
```
dcbriccetti: Forgive my repetition, and I accept that you aren’t going to change it, but
`assertTrue` would tell me right away what kind of assert this is without having to read all the way to the end for the `is(true)`. This is verbose for no good reason.

kkrawczyk123:
**Tested with success!**

src/org/opendatakit/briefcase/export/ExportMediaOverrideOption.java
ggalmazor: You're right! I'll remove it.

test/java/org/opendatakit/briefcase/export/ExportConfigurationTest.java
ggalmazor: It doesn't hurt to remind why I want to follow the `assertThat` style of assertions, especially in a case like this one, which is clearly a tradeoff, as you point out:

- I find that the classic Junit4 assertions have some issues:
  - It's common to make mistakes in the order of arguments, by swapping the expected value with the actual value
  - You need to know about a range of different `assert*` methods in advance
- With the `assertThat` style assertions:
  - Assertions are like an English phrase (subject and predicate), which is easier to understand
  - All assertions follow the same structure, have the same arguments, in the same position, and they always mean the same thing.
  - The error messages are more expressive and useful
  - The matchers are composable and you can easily make your own!
    - This is, by far, the best feature. 
    - It means that you can adapt your matchers to your problem's domain
    - It also means that you can break your assertion into smaller & simpler assertions and combine them without making the resulting assertion harder to understand.
    - The error messages get composed as well, producing useful errors that give the much-needed context while fixing something.

The most common tradeoff is that sometimes you could replace your assertion with an `assertTrue` or an `assertNull`, which only takes an argument but this is outweighed by the benefits of being able to express complex, domain related assertions in the long run. 

This is more of a strategic design decision than a practical, down-to-the-code decision. This assertion style also plays well with test scenarios like the ones we're planning to add to JavaRosa, which help to make the tests about asserting the program's behavior.

src/org/opendatakit/briefcase/export/ExportMediaOverrideOption.java
ggalmazor: It wasn't even used! :O

src/org/opendatakit/briefcase/export/ExportMediaOverrideOption.java
ggalmazor: Nope! Removing.

yanokwa:
@ggalmazor Have you and @dcbriccetti settled your differences? That is, can I merge?

dcbriccetti:
>@ggalmazor Have you and @dcbriccetti settled your differences? That is, can I merge?

Each of us was hoping the other’s Twitter armies would prevail and cause a change, but alas, we’ll have to disagree. https://twitter.com/dcbriccetti/status/1025859803178889216

PR #584, opened 2018-07-20
Issue 568 improve default export conf panel

### LINKED ISSUE #568: Change export default conf UI panel
The export panel is starting to be packed with lots of configuration options. Since we need to support more options in the future, we should change the UI to hide the default conf form in order to save some vertical space.

I'd like to propose we reuse the override conf dialogs to show the default conf form as well. The general idea would be this:

_**Note**: At the end of this issue there's a list of changes from user feedback in the forums_

- When there's no default conf, we show a message and a button to let the user define one:

  export - no default conf

- The user clicks the button and we show the default conf form in a dialog:

  export - default conf dialog

- The user sets some default conf and clicks OK. We show a summary of the default export conf params she has set:

  export - default configuration

- The user can click on the edit button to use the dialog again.

**Changes from [user feedback](https://forum.opendatakit.org/t/briefcase-default-export-configuration-ui/13932):**

- To save vertical space, we won't be showing the conf summary after the user sets one.
- We will simplify the language distinction between "default" and "custom" configurations by removing both adjectives.
- We will reorder the last block of fields this way:
  1. Pull before export
  2. Export media
  3. Overwrite files
- We will review vertical spacing between fields to make coherent groups

ggalmazor:
Closes #568 

This PR changes the UI layout of the export panel, hiding the default export configuration UI components and replacing them with a button that will open a dialog to set the default conf instead.

image

image

image

From an empty storage dir/clean prefs scenario:
- Pull some forms and go to the export tab
- Verify that the button says "Set Configuration"
- Check the form and verify that the OK button is disabled (pending some conf to be set)
- Click on the "Set Configuration button". The conf dialog opens. Verify that the OK button isn't enabled until I enter the export directory.
- Click on OK. The dialog closes. The button now says "View/Edit Configuration". Verify that the Export button is enabled.
- Relaunch Briefcase
- Verify that the button still says "View/Edit Configuration".
- Click on the button. Verify that the conf is the same I set before relaunching
- Click on "Cancel"
- Select some form and check that the Export button is enabled.
- Click on the "View/Edit Configuration" button.
- Click on "Clear all". The dialog closes. Verify that the button is back to "Set Configuration" and the export button is disabled.
- Relaunch Briefcase
- Verify that the button still says "Set Configuration"
- Click on the "Set Configuration" button and verify it's an empty conf.

We can't continue adding fields to the form because we will eat all the vertical space we need for the forms table.

This is a narrow implementation that reuses much of what we did to support export custom override configurations.

Nope. Users with saved default configurations should transition without any problem.

Nope.


yanokwa:
Now that I've played with it, I think the `View/Edit configuration` button might lead a user to think they can select an existing form and hit that button to change the config. Do you see what I mean, @ggalmazor?

I'd propose the following change:
* Change the text to "Edit default configuration"
* Below that button, add "To override the default configuration for each form, press the ⚙️icon"

ggalmazor:
Thanks for your feedback, @yanokwa. I tried to remove all default/override language to see how well that played out and you have confirmed what I suspected :)

I think that the text explaining the override conf buttons could be very nice too.

ggalmazor:
So, just to be on the safe side, I haven't used the gear glyph in the label, since I think that will have issues in windows/mac/linux:

image

dcbriccetti:
I can’t get Ok to enable.

screen shot 2018-07-24 at 6 41 08 pm

dcbriccetti:
Why don’t I see “To override the default configuration for each form, press the gear icon”?

screen shot 2018-07-24 at 7 12 09 pm

ggalmazor:
@dcbriccetti, I think you're on an old version of the branch... This is the commit history I have:

image

dcbriccetti:
I have the same.
```
8e3d0d72 (HEAD -> issue_568_improve_default_export_conf_panel, ggalmazor/issue_568_improve_default_export_conf_panel) Fix UI tests
47da1639 Preload an initial default export on the export panel form
9619531d Use the new conf dialog factory to implement the new UI behavior - The new button launches the default conf dialog - The user's interaction with it triggers the set/reset callbacks
8c69d097 Create new factories for a default export conf panel dialog - Now the override conf dialog's title includes the form name for disambiguation - Specified a different dialog title for the default conf dialog
a6de2647 Change export conf dialog's behavior - Now that the conf panel is aware of the mode it's in (default or override conf), use that to toggle the OK button. - Also, disable OK button when the dialog 
```

I’ll try to track it down. This may be a clue: Any changes I make to ExportPanelForm.form (in the GUI tool) are not shown when I run.

ggalmazor:
Oh, sorry about that. I forgot to sync the `.java` and `.form` files of that view. It should be OK now.

kkrawczyk123:
**Tested with success!**

PR #589, opened 2018-07-26
Force utf8 encoding

### LINKED ISSUE #593: Briefcase is not exporting in Unicode format
ODK-Briefcase version 1.11.2

I'm getting "????????" instead of Unicode characters on the csv file exported using briefcase. The Works well from Direct export from Aggregate Server.

ggalmazor:
Closes #593 

This PR explicitly uses UTF-8 for all NIO2 file writing operations and some `InputStreamReader` that was using the default charset.

This reportedly solves a [forum user's issues](https://forum.opendatakit.org/t/csv-file-lao-characters-not-showing-after-export-from-odk-briefcase/14075/17) with Lao characters being incorrectly encoded while exporting forms in Briefcase.

Run the automated tests.

This changes code that depended on the JDK's default charsets while writing to files with an explicit use of UTF-8 in write operations, which will have a consistent behavior across different platforms and environments.

Nope.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/reused/RemoteServer.java
+++ b/src/org/opendatakit/briefcase/reused/RemoteServer.java
@@ -16,6 +16,7 @@
 
 package org.opendatakit.briefcase.reused;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.toList;
 import static org.javarosa.xform.parse.XFormParser.getXMLText;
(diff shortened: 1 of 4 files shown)


yanokwa:
We've had two users try this and it works, but it seems wise to get this tested on a few OSes. Once we have the go ahead, then I'll ship a hot fix.

yanokwa:
I discussed with @kkrawczyk123 and @ggalmazor and we agree that this is difficult to test, so we'll merge and ship as a hot fix tomorrowish.

PR #591, opened 2018-07-27
At startup, ignore failure to connect to Aggregate server that is refusing connections

### LINKED ISSUE #590: Fails to start when Aggregate server refuses connection
#### Software versions 
Briefcase v1.11.0

#### Problem description
If Briefcase had been configured with an Aggregate server, and is later started with the server refusing connections, Briefcase fails to start.

Maybe fix like this?

screen shot 2018-07-26 at 9 51 26 pm

dcbriccetti:
Closes #590 

I ran Briefcase and it started.

Edit: Testing instructions
- Ensure that "remember passwords" is enabled in the Settings tab
- Create a user for the test in Aggregate
- Configure a pull source with the Aggregate server & user credentials
- Exit Briefcase
- Revoke all granted permissions for the user in Aggregate
- Start Briefcase

Expected results:
- The version in master would write an error stacktrace in the logs and refuse to start. 
- The version from this PR should log the error and start normally. 
- The pull source should be filled in
- The forms list should be empty

It may not be, but does solve the immediate problem.

I don’t think so.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/pull/PullPanel.java
+++ b/src/org/opendatakit/briefcase/ui/pull/PullPanel.java
@@ -36,13 +36,17 @@ import org.opendatakit.briefcase.pull.PullEvent;
 import org.opendatakit.briefcase.reused.BriefcaseException;
 import org.opendatakit.briefcase.reused.RemoteServer;
 import org.opendatakit.briefcase.reused.http.Http;
+import org.opendatakit.briefcase.reused.http.HttpException;
 import org.opendatakit.briefcase.transfer.TransferForms;
 import org.opendatakit.briefcase.ui.ODKOptionPane;
 import org.opendatakit.briefcase.ui.reused.Analytics;
 import org.opendatakit.briefcase.ui.reused.source.Source;
 import org.opendatakit.briefcase.ui.reused.transfer.TransferPanelForm;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class PullPanel {
+  private static final Logger log = LoggerFactory.getLogger(PullPanel.class);
   public static final String TAB_NAME = "Pull";
   private final TransferPanelForm view;
   private final TransferForms forms;
@@ -63,11 +67,15 @@ public class PullPanel {
     getContainer().addComponentListener(analytics.buildComponentListener("Pull"));
 
     // Read prefs and load saved remote server if available
-    this.source = RemoteServer.readPreferences(tabPreferences).flatMap(view::preloadSource);
-    this.source.ifPresent(source -> {
-      forms.load(source.getFormList());
-      view.refresh();
-      updateActionButtons();
+    source = RemoteServer.readPreferences(tabPreferences).flatMap(view::preloadSource);
+    source.ifPresent(source -> {
+      try {
+        forms.load(source.getFormList());
+        view.refresh();
+        updateActionButtons();
+      } catch (HttpException e) {
+        log.warn("Unable to get form list from {}: {}", source.getDescription(), e.toString());
+      }
     });
 
     // Register callbacks to view events


src/org/opendatakit/briefcase/ui/pull/PullPanel.java
```diff
@@ -63,11 +63,15 @@ public PullPanel(TransferPanelForm view, TransferForms forms, BriefcasePreferenc
     getContainer().addComponentListener(analytics.buildComponentListener("Pull"));
 
     // Read prefs and load saved remote server if available
-    this.source = RemoteServer.readPreferences(tabPreferences).flatMap(view::preloadSource);
-    this.source.ifPresent(source -> {
-      forms.load(source.getFormList());
-      view.refresh();
-      updateActionButtons();
+    source = RemoteServer.readPreferences(tabPreferences).flatMap(view::preloadSource);
```
ggalmazor: Is this still happenning? I think `SourcePanel.preload(RemoteServer server)` deals with this:
```java
public Optional<Source<?>> preload(RemoteServer server) {
  try {
    Source source = getSource(server);
    source.set(server);
    triggerOnSource(source);
    return Optional.of(source);
  } catch (HttpException e) {
    log.warn("Can't preload source. Resetting view", e);
    reset();
    return Optional.empty();
  }
}
```

ggalmazor: I think we need to review this whole workflow, though. It could be easier to follow or, at least have some comments that explain the sequence of events/invocations.

src/org/opendatakit/briefcase/ui/pull/PullPanel.java
```diff
@@ -63,11 +63,15 @@ public PullPanel(TransferPanelForm view, TransferForms forms, BriefcasePreferenc
...
+    source.ifPresent(source -> {
```
ggalmazor: So, following my rationale from my first comment, what we should avoid is having something in `source` other than an `Optional.empty()` when the stored conf values won't produce a successful connection to the remote Aggregate server.

src/org/opendatakit/briefcase/ui/pull/PullPanel.java
ggalmazor: After looking a little more into this, it turns out that the behavior is not consistent depending on what happens with the HTTP request:
- If it timeouts, it will throw an `HttpException` and the change on this PR will prevent a program failure
- If the server responds with any non-success (2xx) status code, it won't fail but it will show an empty form list, which is really confusing.

Let's leave your `try/catch` but work a little bit more the `catch` part.
I'll open a new issue to remember we need to normalize the behavior when we get a `4xx` or a `5xx`.

src/org/opendatakit/briefcase/ui/pull/PullPanel.java
```diff
@@ -63,11 +63,15 @@ public PullPanel(TransferPanelForm view, TransferForms forms, BriefcasePreferenc
...
+      try {
+        forms.load(source.getFormList());
+        view.refresh();
+        updateActionButtons();
+      } catch (Exception e) {
+        System.err.println(e.toString());
```
ggalmazor: Since this code is only executed when we're in UI mode, log this error instead of using `System.err`.

Also, the side effects of this should be exactly the same as the callback for `view.onReset()` in line 87 (we want to forget about that server conf and reset the UI). Maybe extract to a method and reuse in both places?

src/org/opendatakit/briefcase/ui/pull/PullPanel.java
ggalmazor: After looking a little more into this, it turns out that the behavior is not consistent depending on what happens with the HTTP request:
- If it timeouts, it will throw an `HttpException` and the change on this PR will prevent a program failure
- If the server responds with any non-success (2xx) status code, it won't fail but it will show an empty form list, which is really confusing.

Let's leave your `try/catch` but work a little bit more the `catch` part.
I'll open a new issue to remember we need to normalize the behavior when we get a `4xx` or a `5xx`.

src/org/opendatakit/briefcase/ui/pull/PullPanel.java
dcbriccetti: Are you sure you want to discard the server configuration? What if it’s a temporary condition?

dcbriccetti: Changing to logging:
`2018-08-12 18:57:18,570 [main] WARN  o.o.briefcase.ui.pull.PullPanel - Unable to get form list from http://localhost:8080: org.opendatakit.briefcase.reused.http.HttpException: Connection refused`

dcbriccetti:
Are you sure you want to discard the server configuration? What if it’s a temporary condition?

Changed to logging:
`2018-08-12 18:57:18,570 [main] WARN o.o.briefcase.ui.pull.PullPanel - Unable to get form list from http://localhost:8080: org.opendatakit.briefcase.reused.http.HttpException: Connection refused`

ggalmazor:
> Are you sure you want to discard the server configuration? What if it’s a temporary condition?

I think it's the less confusing than getting an empty list. What do you think, @yanokwa, @lognaturel?

yanokwa:
Sorry about blocking on this. @lognaturel and I discussed this with @ggalmazor and we think the best way forward is..

1. Not throw the configuration away.
1. Send in a PR that adds refresh button to the UI. 

I'll leave @ggalmazor to file the issues and send in the PR :)

ggalmazor:

ggalmazor:
Thanks, @dcbriccetti! Sorry for keeping you waiting so much time :P

kkrawczyk123:
**Tested with success!**

PR #592, opened 2018-07-27
Modified Export overwriting files confirmation message. Fix #586

### LINKED ISSUE #586: Add a better confirmation message when overwriting files
#### Software versions 
Briefcase master

#### Problem description
When I toggle `Overwrite existing files` while configuring an export, I get that same text with a question mark in the dialog. That's not a helpful confirmation message. Instead, it should say, "The default behavior is to append to existing files. Are you sure you want to overwrite existing files?"

kunalmulwani:
Closes #586 

I have checked the confirmation message on Export tab. Once the user clicks on Overwrite Existing FIles checkbox, the confirmation message is displayed as follows:

screenshot_20180727_151639

It is a simple change of confirmation message.

No.


yanokwa:
Thanks so much for sending in this PR, @kunalmulwani!

PR #597, opened 2018-08-06
Issue 596 Specify the correct Java image for CircleCI builds

### LINKED ISSUE #596: Specify the correct Java image for CircleCI builds
#### Problem description
A recent update on the CircleCI image we use to build JavaRosa has bumped it to use Java 10.0.2 (Oracle), which is incompatible with the JaCoCo version our Gradle uses.

This change makes our builds fail.

#### Expected behavior
The build should use Java 8u171 to work correctly

ggalmazor:
Closes #596 

Force CircleCI to use the latest Java8 image
- Instead of using the lastest "latest" version which at the time of this commit is 10.0.2 (Oracle)
- This is to avoid a failure while testing due to JaCoCo incompatibilities

Tested that now the project builds in CircleCI

This is a straightforward change to avoid depending on a CircleCI env that we can't control.

Nope

Nope

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: ~/work
     docker:
-      - image: circleci/openjdk:latest
+      - image: circleci/openjdk:8u171-jdk
     environment:
       - DISPLAY: :99.0
     steps:


ggalmazor:
Maybe we could/should fast-track this one

PR #598, opened 2018-08-06
Use latest minor release to provide early warning of potential proble…

yanokwa:
Addresses #596

Visual inspection.

I like auto-updating to the latest minor release because that's what we expect from our users. It also feels like it gives us early warning of potential problems we'll run into when we upgrade to a major release.


yanokwa:
Merge when you are ready. You are committer now :)

PR #605, opened 2018-08-16
Make the test less brittle by narrowing the assertion

ggalmazor:
This PR fixes a test of the real http adapter. We used to get the main ODK docs page and assert that the text "Welcome to Open Data Kit", but they just changed that :P

Now we narrow the expected text to just "Open Data Kit".

Run the automated test

We could spin a lightweight http server from the test to ensure this problem never happens again, but that's too complex right, although we could eventually consider doing it.

Nope. It's just a fix on an assertion of a test to avoid false negatives.

Nope


ggalmazor:
I'm fast-tracking this one into master to unblock other builds that are failing. This change doesn't require manual testing.

PR #606, opened 2018-08-16
removing information icon from alert message

### LINKED ISSUE #579: Icon on reload forms from storage location dialog window.
#### Software versions 
Briefcase v1.11.x, Java v1.x.x, operating system: Ubuntu, MacOS, Windows 

#### Problem description
After clicking on Reload forms from the storage location dialog message is displayed and it has an icon which if different on all 2 systems, screenshots attached:
screenshot from 2018-07-12 13-29-59
screenshot from 2018-07-12 13-31-26
screenshot from 2018-07-12 13-34-51

I would like to suggest removing it firstly for unifying behavior for all oses and secondly I remember that when the overwrite dialog was added to export tab - we decided to remove a similar icon.

@lognaturel @yanokwa and @ggalmazor 

@opendatakit-bot label "needs discussion"

kunalmulwani:
Closes #579 

message

Changing the type of JOptionPane to display it as Plain Message.

No.

No.


src/org/opendatakit/briefcase/ui/settings/SettingsPanel.java
```diff
@@ -63,7 +63,7 @@ public SettingsPanel(SettingsPanelForm form, BriefcasePreferences appPreferences
     form.onHttpProxy(appPreferences::setHttpProxy, appPreferences::unsetHttpProxy);
     form.onReloadCache(() -> {
       formCache.update();
-      JOptionPane.showMessageDialog(getContainer(), "Forms successfully reloaded from storage location.");
+      JOptionPane.showMessageDialog(getContainer(), "Forms successfully reloaded from storage location.", "", JOptionPane.PLAIN_MESSAGE);
```
dcbriccetti: Did we consider using APP_NAME as the third argument instead of ""?

kunalmulwani: Currently, <code>APP_NAME</code> is defined in <code>org.opendatakit.briefcase.ui.MainBriefcaseWindow</code>. In order to use it in <code>SettingsPanel</code> I would need to define it in one of the constants file so both the filed could use it. Shoud I go ahead and do that?

dcbriccetti: I’d say make APP_NAME public in MainBriefcaseWindow, and statically import it into SettingsPanel. Unless @ggalmazor prefers something else.

kunalmulwani: @ggalmazor Do you agree with this?

dcbriccetti:
It’s night in Spain. If you want to go ahead, I’ll take the consequences.

src/org/opendatakit/briefcase/ui/settings/SettingsPanel.java
lognaturel: @dcbriccetti Is this common? It seems odd to me for a dialog to have the name of the app but I have to admit I'm not sure what other apps do in this case. I would expect the title to be something like "Success"

dcbriccetti: @lognaturel, thanks for jumping in. I may have been wrong about this. My ancient Swing knowledge is mixing in with my ancient OS/2 Presentation Manager knowledge.

Two of the three other places in the code where PLAIN_MESSAGE is used have the application title, and the other has "".

dcbriccetti:
Sorry for misleading you, @kunalmulwani. I’m fine with "", or waiting to see what @ggalmazor says.

src/org/opendatakit/briefcase/ui/settings/SettingsPanel.java
lognaturel: 😊The two that use the application title are the introductory dialogs so that makes a lot of sense to me. I'd still prefer "Success" but don't feel too strongly about it.

@kunalmulwani, thanks for looking into this issue and making the change. @ggalmazor can be the tie breaker on this one. ;)

ggalmazor:
I have no strong opinion on this. I believe there’s also some variation between Linux/Mac/Windows too.

Since @kunalmulwani has already added the app name, let’s go with that.

We can review in a further PR how Briefcase uses dialogs to normalize alerts, confirmation dialogs and error messages.

What I’m not so sure about merging is all those indentation changes. This project uses 2 spaces... Does that pass checkstyle?

ggalmazor:

kunalmulwani:
@dcbriccetti @ggalmazor Is there anything pending from my side?

dcbriccetti:
>Is there anything pending from my side?

When you see that “needs testing” label appear, it waits for testing (usually a small number of days) and then is merged.

kunalmulwani:
@dcbriccetti Thank you so much for the clarification.

kkrawczyk123:
**Tested with success!**
Verified on Ubuntu, MacOs and Windows. 

PR #604, opened 2018-08-13
Issue 599 custom export conf boolean ui control

ggalmazor:
Closes #599 (and #587)

This PR changes the UI of the boolean conf params of the custom export configuration dialog to show three inlined radio buttons:

image

It also brings code changes to reduce duplication around the concept of a boolean (TRUE/FALSE) value than can be overriden. New artifacts (and tests) have been created to support this model:
- `OverridableBoolean`: Class adapted to the context of Briefcase's domain, that represents a Boolean value that could be overriden by some other Boolean value.
- `TriStateBoolean`: Class that represents a Boolean value with a third UNDERTERMINED state.
- `CustomConfBooleanForm`: UI component to choose between INHERIT, YES, and NO options (radio buttons, currently).

Some of this PR's commits feel big. I wasn't sure how to further split them without having commits that would not compile... I think that stepping through the commits and reviewing the changes on IntelliJ would help. Any and all feedback will be welcome so that I can restructure the PR into more manageable pieces.

#### What has been done to verify that this works as intended?
Run automated tests.
Manually verified that all combinations of boolean conf params work as expected between launches (see #581 for reference of which combinations):

| Default | Custom | Resolved to |
|---|---|---|
|ON|INHERIT|ON|
|ON|ON|ON|
|ON|OFF|OFF|
|OFF|INHERIT|OFF|
|OFF|ON|ON|
|OFF|OFF|OFF|

#### Why is this the best possible solution? Were any other approaches considered?
This, one one hand, brings a UI improvement as discussed in the forums. It is conformant to other UI components currently available.

On the other hand, the export configuration is becoming bigger and by solving #587, we were introducing the third instance of a duplicated structure, which forced some cleaning of the design.

#### Are there any risks to merging this code? If so, what are they?
This change is big and I could have introduced new bugs/regressions. Thorough manual tests are recommended.

#### Does this change require updates to documentation? If so, please file an issue at https://github.com/opendatakit/docs/issues/new and include the link below.
No.

PR #607, opened 2018-08-17
Fix default value of the overwrite files param of the export conf

ggalmazor:
This PR fixes a regression introduced by #604, where the overwrite files export conf param would be true instead of false by default.

Manually tested from a clean storage dir, clean prefs scenario:
- Pull a form with submissions
- Export it twice setting only the output dir
- Check that the output CSV file contains dupe lines
- Toggle the overwrite files conf param
- Export it twice
- Check that the output CSV file has no dupe lines

This is a narrow fix for a regression introduced in a previous PR

Nope.

No.


kkrawczyk123:
Tested with success!

Issue #611, opened 2018-08-22
Leave useful info & data about submissions that fail while exporting


ggalmazor:
Sometimes we have difficulties to triage and debug issue when something goes wrong while exporting encrypted forms. Final users often can't share with us their data and private keys, and Briefcase logs can be not enough to pin down the cause of the failure.

We should explore options to help to triage and debug failures.

- Briefcase currently decrypts encrypted submissions into a file in a temporary random directory in the host machine. When something goes wrong, we could create an extra folder inside the output export directory and store the failing submissions there. This way, users reporting problems could send those to us without compromising their private keys.

PR #614, opened 2018-08-23
Issue 611 export save failed submissions

ggalmazor:
Addresses #611

This PR changes the export process to copy failed submissions to a new error directory inside the export directory. This will let users attach the specific submissions that are failing when reporting issues with Briefcase.

To test this PR, use this storage dir which includes two forms with failing submissions.

[storage_dir.zip](https://github.com/opendatakit/briefcase/files/2331987/storage_dir.zip)

- Pull a form with submissions
- Tamper with a submission by introducing XML syntax errors
- Export the form
- Verify that the invalid form is inside an `errors of NAME OF THE FORM` directory inside the choosen export directory. The submission should have a `failed_submission_1.xml` filename. The number should follow a sequence if there are more than one failing submissions.

This is the narrowest change I could come up with to have this, but it's based on new implicit side-effects during the parsing process, which is not desirable.

A proper way to go about this would involve changing the parsing process to return a `Pair<Path, Optional<Submission>>` instead. This would enable us to group "parsing results" by success/failure and explicitly deal with the new side effects as a "first-class citizen" of the exporting process.

In `ExportToCsv` we're currently doing:
```java
// Generate csv lines grouped by the fqdn of the model they belong to
Map<String, CsvLines> csvLinesPerModel = submissionFiles.parallelStream()
    // Parse the submission and leave only those OK to be exported
    .map(path -> parseSubmission(path, formDef.isFileEncryptedForm(), configuration.getPrivateKey(), configuration.getErrorsDir(formDef.getFormName())))
    .filter(Optional::isPresent)
    .map(Optional::get)
    // Track the submission
    .peek(s -> exportTracker.incAndReport())
    // Use the mapper of each Csv instance to map the submission into their respective outputs
    .flatMap(submission -> csvs.stream()
        .map(Csv::getMapper)
        .map(mapper -> mapper.apply(submission)))
    // Group and merge the CsvLines by the model they belong to
    .collect(groupingByConcurrent(
        CsvLines::getModelFqn,
        reducing(CsvLines.empty(), CsvLines::merge)
    ));
```

Now we would do:
```java
// Parse all submission files
Map<Boolean, List<Pair<Path, Optional<Submission>>>> parsingResults = submissionFiles.parallelStream()
    // Parse the submission and leave only those OK to be exported
    .map(path -> parseSubmission(path, formDef.isFileEncryptedForm(), configuration.getPrivateKey()))
    .collect(groupingByConcurrent(pair -> pair.getRight().isPresent()));

// Copy failed submissions to the errors dir
parsingResults.get(false) // failed ones
    .forEach(pair -> {
      log.error("Can't parse submission", e);
      Path errorsDir = configuration.getErrorsDir(formDef.getFormName());
      if (!exists(errorsDir))
        createDirectories(errorsDir);
      copy(pair.getLeft(), errorsDir.resolve("failed_submission_" + errorSeq.getAndIncrement() + ".xml"));
    });

// Generate csv lines grouped by the fqdn of the model they belong to
Map<String, CsvLines> csvLinesPerModel = parsingResults.get(true) // successfully parsed ones
    .map(Pair::getRight)
    .map(Optional::get)
    // Track the submission
    .peek(s -> exportTracker.incAndReport())
    // Use the mapper of each Csv instance to map the submission into their respective outputs
    .flatMap(submission -> csvs.stream()
        .map(Csv::getMapper)
        .map(mapper -> mapper.apply(submission)))
    // Group and merge the CsvLines by the model they belong to
    .collect(groupingByConcurrent(
        CsvLines::getModelFqn,
        reducing(CsvLines.empty(), CsvLines::merge)
    ));
```

I think that the solution on this PR, although technically less sound, could be better at this point.

Nope

No

### DIFF
--- a/src/org/opendatakit/briefcase/export/ExportConfiguration.java
+++ b/src/org/opendatakit/briefcase/export/ExportConfiguration.java
@@ -452,4 +452,8 @@ public class ExportConfiguration {
   public Optional<String> getExportFileName() {
     return exportFileName;
   }
+
+  public Path getErrorsDir(String formName) {
+    return exportDir.map(dir -> dir.resolve(formName + " - errors")).orElseThrow(BriefcaseException::new);
+  }
 }
--- a/src/org/opendatakit/briefcase/export/ExportToCsv.java
+++ b/src/org/opendatakit/briefcase/export/ExportToCsv.java
@@ -82,7 +82,7 @@ public class ExportToCsv {
     // Generate csv lines grouped by the fqdn of the model they belong to
     Map<String, CsvLines> csvLinesPerModel = submissionFiles.parallelStream()
         // Parse the submission and leave only those OK to be exported
-        .map(path -> parseSubmission(path, formDef.isFileEncryptedForm(), configuration.getPrivateKey(), errorsDir, errorSeq))
+        .map(path -> parseSubmission(path, formDef.isFileEncryptedForm(), configuration.getPrivateKey(), errorsDir, errorSeq, p -> {}))
         .filter(Optional::isPresent)
         .map(Optional::get)
         // Track the submission
--- a/src/org/opendatakit/briefcase/export/ExportToCsv.java
+++ b/src/org/opendatakit/briefcase/export/ExportToCsv.java
@@ -73,10 +73,16 @@ public class ExportToCsv {
 
     csvs.forEach(Csv::prepareOutputFiles);
 
+    // Reset errors management for this export attemp
+    Path errorsDir = configuration.getErrorsDir(formDef.getFormName());
+    AtomicInteger errorSeq = new AtomicInteger(1);
+    if (exists(errorsDir))
+      deleteRecursive(errorsDir);
+
     // Generate csv lines grouped by the fqdn of the model they belong to
     Map<String, CsvLines> csvLinesPerModel = submissionFiles.parallelStream()
         // Parse the submission and leave only those OK to be exported
-        .map(path -> parseSubmission(path, formDef.isFileEncryptedForm(), configuration.getPrivateKey(), configuration.getErrorsDir(formDef.getFormName()), new AtomicInteger(1)))
+        .map(path -> parseSubmission(path, formDef.isFileEncryptedForm(), configuration.getPrivateKey(), errorsDir, errorSeq))
         .filter(Optional::isPresent)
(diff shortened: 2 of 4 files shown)


kkrawczyk123:
In general, it works exactly like described above but a few questions come up during the testing process:

- Exporting failed submissions a few times causes that they are multiplied in error folder in the export directory. Is this behavior an expected one? Shouldn't only a new faulty submission be added?

- What about the numeration? I exported 1 faulty submission of Birds and three of All widgets form. In my error of Birds folder there is failed_submission_1 and in errors of All widgets I have failed submissions from 2-4. I exported all widgets, again and again, I have failed submission from 2-10 - I deleted failed submissions of All widgets and export again now I have failed submissions from 11-13. We could think about improving the numeration or adding form name to names of failed submissions. After a few tries, I have like 19 failed submissions. maybe numbers should start from 1 for each form? And when a number of failed submission should be reset? - now they are after clearing prefs. 

- What behavior should be expected for an empty submission.xml file? Isn't it a faulty submission too? 
I see an Unexpected error dialog message and export process is not finished. The same behavior occurs on the master branch. 

ggalmazor:
Thanks for the testing, @kkrawczyk123!

> Exporting failed submissions a few times causes that they are multiplied in error folder in the export directory

I'll review that, good catch!

> What about the numeration?

Yeah, the number sequence right now is global and doesn't reset until you restart Briefcase. Ideally, it should be per form **and** per export operation and any prior error folder should be cleaned.

> What behavior should be expected for an empty submission.xml file?

There's a wide range of conditions that can lead to a partial export and we don't necessarily want to have traces of all of them in the errors dir. 

When there's some problem with an encrypted form, having the failing submissions at hand will be very useful to quickly reproduce the error and decide whether it's a Briefcase problem or a data problem.

On the other hand, having an empty submission is clearly not a bug in Briefcase, but I'll check that too in case this is relevant.

ggalmazor:
I've changed the code to take care of @kkrawczyk123's feedback. Could you check it again, @dcbriccetti?

Regarding empty submissions, they fail before parsing occurs, when Briefcase tries to sort and filter them by their submission date. In this case, the log error gives enough information to know what's going on. I think we don't have to do anything extra for this particular case at this time.

kkrawczyk123:
**Tested with success!**

PR #615, opened 2018-08-23
Issue 613 cli param to pull incomplete subs

### LINKED ISSUE #613: Add CLI param to include incomplete submissions while pulling a form
The PR https://github.com/opendatakit/aggregate/issues/305 provides a new optional param to include incomplete submissions that Briefcase could use while pulling a form.

We should add a new optional CLI param to enable this on the pull from Aggregate CLI op. The new param's long form should be `--include_incomplete` and its short form should be `-ii`

ggalmazor:
Closes #613

This PR adds a new CLI param to enable requesting incomplete submissions when pulling from Aggregate.

It would be easier to test this PR with an Aggregate server that supports the new parameter. (Will post here the PR/version when it's available)

Used the CLI pull from aggregate op with the new flag and verified on the logs that the requested URL included the `&includeIncomplete=true` part in its query string (INFO log level has to be enabled in the logback conf settings).

Although the code is quite coupled and I've had to pass a new arg around changing many classes, this is the narrowest change I could come up with.

Aggregate doesn't complain about the new arg, even if it doesn't support it (maybe it's an older version without support for this). I can't say the same about other compatible servers.

No, since CLI op params are not included in the docs.

### DIFF
--- a/src/org/opendatakit/briefcase/operations/Export.java
+++ b/src/org/opendatakit/briefcase/operations/Export.java
@@ -27,7 +27,6 @@ import java.time.LocalDateTime;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Optional;
-
 import org.opendatakit.briefcase.export.ExportConfiguration;
 import org.opendatakit.briefcase.export.ExportToCsv;
 import org.opendatakit.briefcase.export.FormDefinition;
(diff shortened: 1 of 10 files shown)


kkrawczyk123:
**Tested with success!**

PR #618, opened 2018-08-25
Issue 617 update jr to 2.11.2 and adapt

### LINKED ISSUE #617: Adapt to JavaRosa 2.11.2 breaking changes
#### Problem description
JavaRosa 2.11.2 introduced a change in the `DataType` enum that replaced the `CHOICE_LIST` (deprecated) value with the proper `MULTIPLE_ITEMS` value.

We should upgrade Briefcase to use the latest JR lib version and adapt to this breaking change.

Update the JR dependency to use the 2.11.2 version and try to build the project.

#### Expected behavior
It should build and run the tests correctly

#### Other information 
N/A

ggalmazor:
Closes #617

This PR doesn't change Briefcase's behavior

This PR updates the JR lib version and adapts to a breaking change that was introduced:
- `DataType.CHOICE_LIST` has been replaced by `DataType.MULTIPLE_ITEMS`

Run the automated tests

We could always _not_ update the JR version but the longer we wait, the bigger the adaptation changes will be. Better to do have smaller changesets.

None.

No.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ dependencies {
         exclude group: 'commons-logging'
     }
     compile group: 'net.sf.kxml', name: 'kxml2', version: '2.3.0'
-    compile(group: 'org.opendatakit', name: 'opendatakit-javarosa', version: '2.10.0') {
+    compile(group: 'org.opendatakit', name: 'opendatakit-javarosa', version: '2.11.2') {
         exclude group: 'org.slf4j'
     }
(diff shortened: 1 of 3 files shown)


lognaturel:
> We could always not update the JR version but the longer we wait, the bigger the adaptation changes will be. Better to do have smaller changesets.

Agreed! But without this change the newest All Widgets sample form didn't work anymore because it has `rank`. So great to see the update. 😊

PR #619, opened 2018-08-25
Issue 610 enhanced border effect in windows

### LINKED ISSUE #610: Make headers separators more visible.
#### Software versions 
Briefcase v1.11.x, Java v1.8.x, operating system: Windows 10

#### Problem description
I think that horizontal and vertical separators of the forms table's header row could be more visible on the Windows system on all Pull, Push and Export tabs. It is not a high priority issue just a small visual improvement. 
Screenshot attached: 
screenshot from 2018-08-21 10-02-24

1. Run briefcase on windows os.

huangyz0918:
Closes #610 

Enhanced border effect in windows.
Now, It looks like:

image

No. It only affects Windows.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
+++ b/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
@@ -19,6 +19,7 @@ import static javax.swing.SortOrder.ASCENDING;
 
 import java.awt.Color;
 import java.awt.Dimension;
+import java.awt.Font;
 import java.awt.event.MouseEvent;
 import java.util.Collections;
 import javax.swing.BorderFactory;
@@ -33,7 +34,6 @@ import javax.swing.table.TableModel;
 import javax.swing.table.TableRowSorter;
 import org.opendatakit.briefcase.ui.reused.MouseAdapterBuilder;
 import org.opendatakit.briefcase.ui.reused.UI;
-import org.opendatakit.briefcase.util.FindDirectoryStructure;
 
 public class ExportFormsTableView extends JTable {
   static final String[] HEADERS = new String[]{"", "", "Form Name", "Export Status", "Last Export", ""};
@@ -58,11 +58,8 @@ public class ExportFormsTableView extends JTable {
     Dimension lastExportDims = getHeaderDimension(HEADERS[LAST_EXPORT_COL]);
 
     JTableHeader header = getTableHeader();
-    if (FindDirectoryStructure.isWindows()) {
-      Border border = BorderFactory.createLineBorder(Color.GRAY);
-      header.setBackground(Color.GRAY);
-      header.setBorder(border);
-    }
+    header.setOpaque(false);
+    header.setBackground(new Color(229,229,229));
 
     setRowHeight(28);
 
--- a/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
+++ b/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
@@ -17,7 +17,6 @@ package org.opendatakit.briefcase.ui.export.components;
 
 import static javax.swing.SortOrder.ASCENDING;
 
-import java.awt.Color;
 import java.awt.Dimension;
 import java.awt.event.MouseEvent;
 import java.util.Collections;
@@ -25,11 +24,11 @@ import javax.swing.JButton;
 import javax.swing.JTable;
 import javax.swing.RowSorter;
 import javax.swing.SortOrder;
-import javax.swing.table.JTableHeader;
 import javax.swing.table.TableColumnModel;
 import javax.swing.table.TableModel;
 import javax.swing.table.TableRowSorter;
 import org.opendatakit.briefcase.ui.reused.MouseAdapterBuilder;
+import org.opendatakit.briefcase.ui.reused.TableCustomizer;
 import org.opendatakit.briefcase.ui.reused.UI;
 
 public class ExportFormsTableView extends JTable {
@@ -50,15 +49,12 @@ public class ExportFormsTableView extends JTable {
 
     addMouseListener(new MouseAdapterBuilder().onClick(this::relayClickToButton).build());
 
-    Dimension formNameDims = getHeaderDimension(HEADERS[FORM_NAME_COL]);
-    Dimension exportStatusDims = getHeaderDimension(HEADERS[EXPORT_STATUS_COL]);
-    Dimension lastExportDims = getHeaderDimension(HEADERS[LAST_EXPORT_COL]);
+    Dimension formNameDims = TableCustomizer.getHeaderDimension(HEADERS[FORM_NAME_COL], this);
+    Dimension exportStatusDims = TableCustomizer.getHeaderDimension(HEADERS[EXPORT_STATUS_COL], this);
+    Dimension lastExportDims = TableCustomizer.getHeaderDimension(HEADERS[LAST_EXPORT_COL], this);
 
-    JTableHeader header = getTableHeader();
-    header.setOpaque(false);
-    header.setBackground(new Color(229,229,229));
-
-    setRowHeight(28);
+    TableCustomizer.customizeHeader(getTableHeader());
+    TableCustomizer.customizeTable(this);
 
     TableColumnModel columns = getColumnModel();
     columns.getColumn(SELECTED_CHECKBOX_COL).setMinWidth(40);
@@ -85,14 +81,6 @@ public class ExportFormsTableView extends JTable {
     setRowSorter(sorter);
     sorter.sort();
   }
-
-  private Dimension getHeaderDimension(String header) {
-    return getTableHeader()
-        .getDefaultRenderer()
-        .getTableCellRendererComponent(null, header, false, false, 0, 0)
-        .getPreferredSize();
-  }
-
 
   private void relayClickToButton(MouseEvent event) {
     int column = getColumnModel().getColumnIndexAtX(event.getX());
--- a/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
+++ b/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
@@ -19,15 +19,12 @@ import static javax.swing.SortOrder.ASCENDING;
 
 import java.awt.Color;
 import java.awt.Dimension;
-import java.awt.Font;
 import java.awt.event.MouseEvent;
 import java.util.Collections;
-import javax.swing.BorderFactory;
 import javax.swing.JButton;
 import javax.swing.JTable;
 import javax.swing.RowSorter;
 import javax.swing.SortOrder;
-import javax.swing.border.Border;
 import javax.swing.table.JTableHeader;
 import javax.swing.table.TableColumnModel;
 import javax.swing.table.TableModel;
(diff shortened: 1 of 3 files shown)


dcbriccetti:
Did we look at ways to solve this that wouldn’t involve any OS-specific code?

huangyz0918:
@dcbriccetti The strange thing is that this method does not work in Windows 10, which looks like:

image

So, this is why I like previous solution. I tried adding a background to avoid using any OS-specific code.
The new commit looks like:

macOS:
image
Windows10:
image

What do you think about this ? @dcbriccetti , or just previous solution?

dcbriccetti:
>this method does not work in Windows 10

By “this method” you mean the approach of increasing the font size? Are you saying it doesn’t increase the font size? My suggestion is that we let the larger font size solve the problem of the headings looking small relative to the rows, and that we don’t worry about background or borders, letting the Look and Feel do what it will.

huangyz0918:
I am sorry that I did not make it clear.
> By “this method” you mean the approach of increasing the font size? Are you saying it doesn’t increase the font size?

The font size was increased in Windows 10, but the border still unclear, so I think is adding a background color will be better. What do you think?

If you think it doesn't matter, I can update this PR as your proposed soon. Thanks.

dcbriccetti:
Thanks for the clarification. I’m concerned about making changes that we don’t fully understand. Or perhaps you do understand them.
Why is this necessary?: `header.setOpaque(false);`
Is it sufficient to only do that, and not change the background color?

If we do go ahead with a table header appearance customization, I suggest you factor it out rather than duplicating the code. You could create a class such as `TableCustomizer` with a static method `customizeHeader` where you put the code, along with comments explaining exactly what we are doing and why. Or you could create an abstract class like `CustomTable` that extends `JTable` and put the code there.

Or perhaps we will decide we don’t understand what’s going on and lack the Swing expertise to do this right, respecting multiple platforms and programmable look and feels (https://docs.oracle.com/javase/tutorial/uiswing/lookandfeel/plaf.html). I don’t know you, @huangyz0918, and perhaps you do deeply understand this.

ggalmazor:
Thanks, @huangyz0918!

I agree with @dcbriccetti that we need to explore the solution a little bit more to be more comfortable before merging. I know this can be frustrating but making this kind of customization work with Swing UIs can be challenging.

If I understood correctly, we are aiming to do the smallest possible customization, to avoid having a snowball effect down the road.

I'd suggest going with the `TableCustomizer`. We are trying to stay away from using inheritance for composition purposes. Having the `ExportFormsTable` and the `TransferFormsTable` *use* a `TableCustomizer` seems like an idea that fits the current design.

huangyz0918:
Thanks for your suggestions! @dcbriccetti @ggalmazor 

> Why is this necessary?: header.setOpaque(false);

In macOS, this may be useless, but it can make the background color available in Windows 10.

> If we do go ahead with a table header appearance customization, I suggest you factor it out rather than duplicating the code.

Agreed. I can create a simple util class (like `TableCustomizer.class`) to hold the custom code.

> we are aiming to do the smallest possible customization.

That's true, since this issue we are talking about is just a tiny improvement. So, I'd like to address it in a easier way.

As for the inconsistent issue with table headers on different platforms (macOS and Windows10), it's because the default `TableCellRenderer` are different in different platforms (this is why I add `header.setOpaque(false);` ), so, what we need to do is set a fixed `TableCellRenderer`.

I propose to add a background color in the table header and make it has the same effect in both macOS and Windows 10.

huangyz0918:
Now, it looks better:
| Windows10 | macOS |
|:---------------:|:-----:|
|windows|macOS|

Please review @dcbriccetti @ggalmazor , thanks.

ggalmazor:
The end result seems OK to me. Here's a screenshot from Ubuntu:
image

src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
```diff
@@ -50,15 +49,12 @@
 
     addMouseListener(new MouseAdapterBuilder().onClick(this::relayClickToButton).build());
 
-    Dimension formNameDims = getHeaderDimension(HEADERS[FORM_NAME_COL]);
-    Dimension exportStatusDims = getHeaderDimension(HEADERS[EXPORT_STATUS_COL]);
-    Dimension lastExportDims = getHeaderDimension(HEADERS[LAST_EXPORT_COL]);
+    Dimension formNameDims = TableCustomizer.getHeaderDimension(HEADERS[FORM_NAME_COL], this);
```
ggalmazor: I'd suggest adding a static import of all these methods in TableCustomizer to make the code less verbose.

src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
```diff
@@ -86,14 +82,6 @@
     sorter.sort();
   }
 
-  private Dimension getHeaderDimension(String header) {
```
ggalmazor: Good idea to extract this into the `TableCustomizer` class.

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
```diff
@@ -0,0 +1,61 @@
...
+import java.awt.Color;
+import java.awt.Dimension;
+import javax.swing.JTable;
+import javax.swing.UIManager;
+import javax.swing.table.JTableHeader;
+
+public class TableCustomizer {
+
+  /**
+   * for header customization.
```
ggalmazor: I'd suggest to change this to something that explains what is this method doing. Something like "Sets properties of the given {@link JTableHeader} to improve its visual appearance across all platforms (Mac, Windows, and Linux)"

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @param header
+   */
+  public static void customizeHeader(JTableHeader header) {
+    // to make the background color is available.
+    header.setOpaque(false);
+    header.setBackground(UIManager.getColor("ToolBar.background"));
```
ggalmazor: I'd add a static import for all those `UIManager` methods to reduce code verbosity.

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.swing.JTable;
+import javax.swing.UIManager;
+import javax.swing.table.JTableHeader;
+
+public class TableCustomizer {
+
+  /**
+   * for header customization.
+   *
+   * @param header
```
ggalmazor: I think these params don't add much value. I'd suggest removing them.

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
```diff
@@ -0,0 +1,61 @@
...
+   */
+  public static void customizeHeader(JTableHeader header) {
+    // to make the background color is available.
+    header.setOpaque(false);
+    header.setBackground(UIManager.getColor("ToolBar.background"));
+    header.setBorder(UIManager.getBorder("MenuBar.border"));
+  }
+
+  /**
+   * for table customization.
```
ggalmazor: Same comment as with the `customizeHeader(JTableHeader)` method. Put something more descriptive.

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @param table
```
ggalmazor: Same here, remove it?

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
```diff
@@ -0,0 +1,61 @@
...
+   */
+  public static void customizeTable(JTable table) {
+    table.setGridColor(Color.lightGray);
```
ggalmazor: Same: static import `lightGray` from `Color`?

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
```diff
@@ -0,0 +1,61 @@
...
+    table.setRowHeight(28);
+  }
+
+  /**
+   * to get a dimension from specific header.
```
ggalmazor: Change to "Returns the {@link Dimension} of the given header in the given {@link JTable} using its related cell renderer component's preferred size"

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @param table
```
ggalmazor: Same: remove them params?

src/org/opendatakit/briefcase/ui/reused/transfer/TransferFormsTableView.java
```diff
@@ -49,14 +48,11 @@ public TransferFormsTableView(TransferFormsTableViewModel model, String[] header
 
     addMouseListener(new MouseAdapterBuilder().onClick(this::relayClickToButton).build());
 
-    Dimension formNameDims = getHeaderDimension(getHeader(FORM_NAME_COL));
-    Dimension statusDims = getHeaderDimension(getHeader(STATUS_COL));
+    Dimension formNameDims = TableCustomizer.getHeaderDimension(getHeader(FORM_NAME_COL), this);
```
ggalmazor: Same: static import?

src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
huangyz0918: Improved.

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
huangyz0918: Improved.

src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
huangyz0918: Thank you.

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
```diff
@@ -15,43 +15,43 @@
...
   /**
-   * to get a dimension from specific header.
-   *
-   * @param table
-   * @param header
+   * Returns the {@link Dimension} of the given header in the given {@link JTable}
+   * using its related cell renderer component's preferred size.
    */
-  public static Dimension getHeaderDimension(String header, JTable table) {
+  public static Dimension getHeaderDimension(JTable table, String header) {
```
ggalmazor: Nicer arg ordering, thanks!

src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
dcbriccetti: Hey, he was praising *me*. 😀

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
dcbriccetti: You have mad doc skillz, @ggalmazor.

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
dcbriccetti: Ha! After I just complimented your English. ;-)

src/org/opendatakit/briefcase/ui/reused/TableCustomizer.java
ggalmazor: I don't think the GitHub user @link agrees with you, but ¯\\_(ツ)_/¯

dcbriccetti:
Nice work, @huangyz0918! And good teamwork.

Issue #625, opened 2018-09-04
Export submissions from ODK dir is failing. 


kkrawczyk123:
#### Software versions 
Briefcase v1.11.x, Java v1.x.x, operating system, Aggregate v1.x.x, Collect v1.x.x...

#### Problem description
There is an unexpected error displayed on the newest master when exporting submissions pulled from odk directory. In the log file there is Error while exporting forms
org.opendatakit.briefcase.reused.BriefcaseException: No instance ID found.

On the 1.11.1 beta, there is no error displayed but submissions are not exported, 0 submissions exported is displayed in the status, in the csv file there is only the header visible.

On the briefcase 1.10 submissions are exported, no error occurs, but they don't have SubmissionDate.

#### Steps to reproduce the problem
1. Pull submissions from odk folder
2. Export pulled submissions

ggalmazor:
So, after analyzing this issue, we have identified a particular scenario that we should solve separately: submissions without an instance ID.

These submissions have always been excluded from the export output. We could think of alternatives:
- Briefcase could assign each submission with a random ID, used only when exporting (wouldn't change the source submission.xml files in the storage directory)

ggalmazor:
The hardest problem to solve is when forms have repeat groups. Null IDs in this scenario wouldn't work because it would be impossible to link rows from a repeat group CSV file to the main CSV file.

If Briefcase assigns random IDs, they can track a repeat group CSV file line to it's parent file in the main CSV file.

Also, I think random IDs don't harm the scenario where there are no repeat groups.

We could also use random IDs only if we know there are repeat groups, and use null IDs otherwise, but that feels like it's going to be more confusing for users if they don't realize when are null IDs expected and when not.

ggalmazor:
After discussing this with @yanokwa, and @lognaturel, we decided to redefine the validity rules of a submission:

> This means that for a submission with repeat groups to be valid, it needs to have an instance id. This makes sense because, otherwise, it wouldn't be possible to link the lines from the output repeat CSV file to their corresponding lines on the on the main CSV output files.
>
> When a form doesn't define repeat groups, though, it's perfectly reasonable to not have instance ids, and the output CSV file will have blank values on that column.
(copied from #626)

lognaturel:
[...]

PR #626, opened 2018-09-04
Issue 625 export subs without instance id

ggalmazor:
Addresses #625

This PR fixes a regression that produces an error when some of the exported submissions have no instance ID.

Now, no error will be shown and the invalid submissions will be excluded and copied to the error output directory.

The definition of a valid submission is:
- The form it belongs to doesn't define repeat groups
OR 
- it has an instance id

This means that for a submission with repeat groups to be valid, it needs to have an instance id. This makes sense because, otherwise, it wouldn't be possible to link the lines from the output repeat CSV file to their corresponding lines on the on the main CSV output files.

When a form doesn't define repeat groups, though, it's perfectly reasonable to not have instance ids, and the output CSV file will have blank values on that column.

(Provided you have pulled some form that defines repeat groups with submissions)
- If all the submissions have instance IDs, edit one and remove it
- Export the form
- Verify that no error is shown on the UI
- Verify that the details dialog tells that not all the submissions have been exported
- Verify that there are some submissions missing in the output files
- Verify that there are some submissions in the errors output directory

(Provided you have pulled another form that doesn't define repeat groups with submissions)
- If all the submissions have instance IDs, edit one and remove it
- Export the form
- Verify that no error is shown on the UI
- Verify that the details dialog tells that all the submissions have been exported
- Verify that there are no submissions missing in the output files
- Verify that there are no submissions in the errors output directory

This is the narrowest change to reuse the current system to handle errors while exporting submissions, by adding a new error case.

No.

No.

### DIFF
--- a/src/org/opendatakit/briefcase/export/CsvSubmissionMappers.java
+++ b/src/org/opendatakit/briefcase/export/CsvSubmissionMappers.java
@@ -20,6 +20,7 @@ import static java.text.DateFormat.getDateTimeInstance;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
+
 import static org.javarosa.core.model.DataType.DATE;
 import static org.javarosa.core.model.DataType.DATE_TIME;
 import static org.javarosa.core.model.DataType.GEOPOINT;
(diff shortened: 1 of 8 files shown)


kkrawczyk123:
**Tested with success!**
Verified on Ubuntu, MacOS and WIndows.

@opendatakit-bot  label behavior verified"

PR #629, opened 2018-09-05
Issue 628 change export button behavior

### LINKED ISSUE #628: Export button is disabled after first pull because no default configuration is set
After a user has pulled some forms, they will often want to export those forms. Now that the default configuration is hidden behind a button, it's not as clear that they need to set values there. They may check a form, try to tap the Export button and find that they're stuck. Nothing indicates that the problem is that there's no configuration set.

Some options:
- when a form is selected, if no default configuration is set, add text in a dialog or above the set configuration button to indicate a configuration is required
- enable the export button but when it's clicked, show a dialog indicating a configuration is required

I can see this being a dead end for users so we may want to address it before the release?

ggalmazor:
Closes #628

This PR changes the behavior of the export button on the UI.

Now, the button will be always enabled & clickable by the user, except while an export process is on the way.

When the button is clicked we check that:
- at least one form has been selected
- all selected forms have been configured with an export directory
  - Either by the default configuration
  - Or by custom individual configurations
- all encrypted forms have been configured with a PEM file
  - Either by the default configuration
  - Or by custom individual configurations

If any of these checks fail, we show a dialog informing the user about this and calling for action.

If none of these checks fail, the export process starts.

- Tested manually on the UI each check in isolation and combined with unencrypted forms and encrypted forms. Any will do.
- Run automated tests.

We discarded the alternative of showing some label because users could miss it.

No.

No.

### DIFF
--- a/src/org/opendatakit/briefcase/export/Csv.java
+++ b/src/org/opendatakit/briefcase/export/Csv.java
@@ -3,6 +3,7 @@ package org.opendatakit.briefcase.export;
 import static java.nio.file.StandardOpenOption.APPEND;
 import static java.nio.file.StandardOpenOption.CREATE;
 import static java.nio.file.StandardOpenOption.TRUNCATE_EXISTING;
+
 import static org.opendatakit.briefcase.export.CsvSubmissionMappers.getMainHeader;
 import static org.opendatakit.briefcase.export.CsvSubmissionMappers.getRepeatHeader;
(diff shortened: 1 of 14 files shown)


ggalmazor:
I've rebased master and solved conflicts from merging #626. This would be ready to merge after it gets a review approval. /cc @lognaturel @yanokwa

lognaturel:
Nice to see the venerable `if`s and `for`s doing good work. 😉

PR #630, opened 2018-09-05
Issue 230 ensure safe dir and file names

### LINKED ISSUE #230: CSV filename is not safe on all file systems
Currently `ExportToCsv.safeFilename` removes punctuation and spaces, but it allows Unicode and characters that might cause problems on other file systems. I propose only allowing only ASCII characters, space, hyphen, and underscore (similar to https://github.com/opendatakit/briefcase/pull/224). 

So a CSV called "Båé Førm.csv" would now be "B Frm.csv".

ggalmazor:
Closes #230

This ensures that file and directories created by Briefcase don't have illegal characters:
- Export errors output directory
- A form's directory in the storage dir when it's installed individually

This PR also flags legacy places that should be reviewed when refactored.

Since the `stripIllegalChars` method has been already tested, no special checks have been made, other than compiling and running the automated test.

We could refactor the legacy places but it doesn't feel worth it:
- It feels risky changing a piece of code that nobody's complained for a long time ago
- We will probably have to change it when we implement planned features like the central Database, or when we refactor the low-level form pulling code to improve the concurrency model

My gut feeling is that changing that now will produce rework.

No.

No.


lognaturel:
This seems fine to me to merge now but @ggalmazor would be best for you to check with @kkrawczyk123 on timing.

kkrawczyk123:
**Tested with success!**

PR #633, opened 2018-09-06
Replace risk section with behavior changes section

lognaturel:
It's a text-only change so I manually looked at the diff. It's the same change as https://github.com/opendatakit/collect/pull/2488.

Good to keep the same template across tools.

Where we've had problems in the past was with larger PRs that resulted in many different changes, some carefully considered and others that turned out to be unintentional side effects. If folks doing verification at least know what code has changed, they are more likely to identify unintended changes.

In this case, the users of the template are developers who submit pull requests. This change intends to guide those developers to provide more information on intended results of their code change and possible unexpected side effects. It also intends to help make sure we don't merge changes that break the connected tests.

This does add some complexity to the template so a risk is that developers stop filling it out with care.

#### Do we need any specific form for testing your changes? If so, please attach one.
No

#### Does this change require updates to documentation? If so, please file an issue [here]( https://github.com/opendatakit/docs/issues/new) and include the link below.
No

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,6 +4,6 @@ Closes #
 
 #### Why is this the best possible solution? Were any other approaches considered?
 
-#### Are there any risks to merging this code? If so, what are they?
+#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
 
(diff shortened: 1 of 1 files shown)


yanokwa:
I think `Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes.` is very clear and succinct. To me, the first and last sentences, muddle that clarity. I don't want to be a blocker on merge, but just wanted to put that out there for consideration!

PR #636, opened 2018-09-11
WIP Testing

ggalmazor:


src/org/opendatakit/briefcase/model/OdkCollectFormDefinition.java
```diff
@@ -70,31 +70,16 @@ public String toString() {
     return getFormName();
   }
 
-  /*
-   * (non-Javadoc)
-   *
-   * @see org.opendatakit.briefcase.model.IFormDefinition#getFormName()
-   */
```
dcbriccetti: Thanks for getting rid of these hideous noisemakers.

src/org/opendatakit/briefcase/model/OdkCollectFormDefinition.java
```diff
@@ -117,7 +97,8 @@ public boolean equals(Object obj) {
       String id = getFormId();
       String versionString = getVersionString();
 
-      return (id.equals(lf.getFormId()) && ((versionString == null) ? (lf.getVersionString() == null)
+      return (id.equals(lf.getFormId()) && ((versionString == null)
```
dcbriccetti: I’d remove the outermost, redundant, pair of parentheses. When I see `return (expression)` it suggests the programmer thinks that return is a function and not a statement.

src/org/opendatakit/briefcase/ui/reused/FontUtils.java
```diff
@@ -22,18 +22,19 @@
...
+import java.util.Objects;
 
 public class FontUtils {
 
   public static Font getCustomFont(String name, Float size) {
 
     // if we are in a jar, then use the classloader, if not, load from filesystem
     File file = getResourceAsFile(name);
-    if (!file.exists()) {
+    if (!Objects.requireNonNull(file).exists()) {
```
dcbriccetti: How is the NPE thrown from requireNonNull better than the NPE thrown from dereferencing a null `file`?

src/org/opendatakit/briefcase/util/AggregateUtils.java
```diff
@@ -264,26 +243,21 @@ private static final DocumentFetchResult httpRetrieveXmlDocument(HttpUriRequest
           WebUtils.resetHttpContext();
           ex = new XmlDocumentFetchException("Authentication failure");
         } else if (statusCode == 400) {
-          ex = new XmlDocumentFetchException(description.getFetchDocFailed() + webError + " while accessing: "
-              + uri.toString() + "\nPlease verify that the " + description.getDocumentDescriptionType()
-              + " that is being uploaded is well-formed.");
+          ex = new XmlDocumentFetchException("" +
```
dcbriccetti: I don’t see the value of the `"" +`, here and below.

ggalmazor: I thought you would like it! It lets next lines to be correctly indented and aligned.

src/org/opendatakit/briefcase/ui/reused/FontUtils.java
ggalmazor: This way at least it’s explicit that the value can be null.

It was suggested by IntelliJ

src/org/opendatakit/briefcase/util/AggregateUtils.java
dcbriccetti: You are the human. You can align the lines below however you like. :-)

src/org/opendatakit/briefcase/ui/reused/FontUtils.java
dcbriccetti: Oh, I misread what it does then. Sorry.

src/org/opendatakit/briefcase/util/AggregateUtils.java
ggalmazor: But the IDE does that! Not the human! :D

ggalmazor:
Merge this now, yay or nay, @lognaturel?

lognaturel:
I'm happy with it being merged now! I haven't looked in detail but since there's an approval, either of you can merge when ready.

PR #640, opened 2018-09-13
Remove the sentry appender from the prod example conf file

ggalmazor:

This PR removes the Sentry appender from the Logback production example conf file, since won't be relying on that to get error reports to Sentry after #564.

There are no code changes. Nothing to verify :)

This is just something I forgot to do in #564. It's surfaced after doing a release and realizing that the file still had the appender.

Users won't notice a thing.

Nope.


ggalmazor:
Thanks, Dave!

PR #641, opened 2018-09-13
Issue 637 merge audit files

### LINKED ISSUE #637: [Export] Aggregate audit information into one output CSV file
#### Problem description

As explained in #621, when exporting a form that attaches audit information, the audit CSV output files get saved in the media output folder, like any other binary attachments. Since all submissions use the fixed `audit.csv` filename:

- Exporting only one form will get you all the audit files in the media output directory, and their file names will have a sequence number, making it hard to link them to their corresponding submission without consulting the main/repeat CSV files (all the filenames are like `audit.csv`, `audit-2.csv`, `audit-3.csv`, etc).
- Exporting more than one form with audit information at the same time will produce a failure since the second form will find that the audit files (from the first form) already exist on the output directory.

#### Example Audit data

Event | Node | Start | End
-- | -- | -- | --
form start |   | Tue Sep 11 13:06:26 GMT+200 2018 | N/A
question | /data/some_field | Tue Sep 11 13:06:26 GMT+200 2018 | Tue Sep 11 13:06:28 GMT+200 2018
jump |   | Tue Sep 11 13:06:28 GMT+200 2018 | Tue Sep 11 13:06:30 GMT+200 2018
end screen |   | Tue Sep 11 13:06:30 GMT+200 2018 | Tue Sep 11 13:06:31 GMT+200 2018
form save |   | Tue Sep 11 13:06:31 GMT+200 2018 | N/A
form exit |   | Tue Sep 11 13:06:31 GMT+200 2018 | N/A
form finalize |   | Tue Sep 11 13:06:31 GMT+200 2018 | N/A

- Use these forms: [Audit 1](https://github.com/opendatakit/briefcase/files/2370724/audit-1.txt), [Audit 2](https://github.com/opendatakit/briefcase/files/2370723/audit-2.txt)
- Create some submissions and try to export them individually and together

#### Expected behavior

Instead of having an audit CSV output file per exported submission saved in the media output directory, we want all audit data of the same form to be merged into a single audit CSV output file. 

To do so, Briefcase should append a new column to the audit information with the submission UID it belongs to. The resulting Audit Data would look like this:

Instance ID |Event | Node | Start | End
-- | -- | -- | -- | --
uuid:53a83013-99ff-434d-a096-68fe1464c249 | form start |   | Tue Sep 11 13:06:26 GMT+200 2018 | N/A
uuid:53a83013-99ff-434d-a096-68fe1464c249 | question | /data/some_field | Tue Sep 11 13:06:26 GMT+200 2018 | Tue Sep 11 13:06:28 GMT+200 2018
uuid:53a83013-99ff-434d-a096-68fe1464c249 | jump |   | Tue Sep 11 13:06:28 GMT+200 2018 | Tue Sep 11 13:06:30 GMT+200 2018
uuid:53a83013-99ff-434d-a096-68fe1464c249 | end screen |   | Tue Sep 11 13:06:30 GMT+200 2018 | Tue Sep 11 13:06:31 GMT+200 2018
uuid:53a83013-99ff-434d-a096-68fe1464c249 | form save |   | Tue Sep 11 13:06:31 GMT+200 2018 | N/A
uuid:53a83013-99ff-434d-a096-68fe1464c249 | form exit |   | Tue Sep 11 13:06:31 GMT+200 2018 | N/A
uuid:53a83013-99ff-434d-a096-68fe1464c249 | form finalize |   | Tue Sep 11 13:06:31 GMT+200 2018 | N/A

This way, all audit data of submissions from the same form could be cleanly merged to produce a single output file.

The proposed filename would follow this template: `{FORM NAME} - audit.csv`.

#### Other information 

Audit information is mapped with a binary binding, which is exported by Briefcase by writing their corresponding file names **and, as a side-effect**, by copying the media files from the storage directory to the media output directory.

~A strategy to implement the proposed behavior would be to change `Model.getDataType()` to consider not only the field's bound data type, but also the field's name and ancestry (`meta.audit`) to return a new `DataType.AUDIT` that could be mapped into a new `CsvFieldMapper` that would write the same values to the output CSV files and append the audit file's contents to the new proposed output file.~ This strategy is not a good idea because it would involve adding a new DataType in JR, although the new mapper concept is still valid.

This mapper could take care of:
- Create the audit output file if it doesn't exist, adding the corresponding headers
- Add the submission's instance ID column to each one of the audit lines, before appending them to the audit output file.

ggalmazor:
Closes #637

It appears that GitHub has reordered some commits, which could be a little confusing for the reviewer of this PR. Sorry about that! ¯\\\_(ツ)\_/¯

- Run the automated tests
- Manually exported [Audit 1](https://github.com/opendatakit/briefcase/files/2370724/audit-1.txt), and [Audit 2](https://github.com/opendatakit/briefcase/files/2370723/audit-2.txt) forms with some submissions.

This is the narrowest change I could think of. It ensures that previous behavior doesn't change by adding a new mapper instead of changing the current binary mapper.

There are some ways we could address a performance decrease (see last section):
- we could post-process of the submissions and writing once in the output audit file
- we could change the side-effect of the new field mapper (writing to a file) and make it aggregate the lines into some external/global stream of lines

Both options would force us to change the structure of the main export process and the second option would make us think about concurrency/contingency if we want the audit contents to be ordered.

Taking into account that we don't really know how many people is using & exporting audit data, it feels like a more conservative strategy to wait and intervene if there are performance complaints.

- Users should expect a new output CSV file with merged audit data from all the submissions of each exported form.
- Users could miss the `audit-N.csv` files we were previously writing into the media output directory
- Now users will be able to export two forms with audit data at the same time without breaking Briefcase! :tada:
- The new filenames follow the pattern `{FORM NAME} - audit.csv`
- Exported audit contents follow the same behavior as the main and repeat CSV contents regarding overwriting/appending lines. They are affected by the "overwrite files" export conf param.
- Since we basically have to process all the audit lines of all submissions, big forms with many submissions will experience a decrease in performance while exporting them.

We don't currently document audit data export. Should we?

### DIFF
--- a/src/org/opendatakit/briefcase/export/Csv.java
+++ b/src/org/opendatakit/briefcase/export/Csv.java
@@ -73,7 +73,7 @@ class Csv {
         output,
         false,
         configuration.resolveOverwriteExistingFiles(),
-        CsvSubmissionMappers.repeat(groupModel, configuration)
+        CsvSubmissionMappers.repeat(formDefinition, groupModel, configuration)
     );
   }
 
--- a/src/org/opendatakit/briefcase/export/CsvFieldMapper.java
+++ b/src/org/opendatakit/briefcase/export/CsvFieldMapper.java
@@ -36,4 +36,24 @@ interface CsvFieldMapper {
   // TODO Normalize the weird empty/null value encoding scheme and simplify this method to return a Stream<String> of just csv column values
   // TODO Simplify args by passing the ExportConfiguration object
   Stream<Pair<String, String>> apply(String formName, String localId, Path workingDir, Model model, Optional<XmlElement> maybeElement, ExportConfiguration configuration);
+
+  /**
+   * Returns a new mapper, result of composing this mapper with the given mapper
+   *
+   * @see #compose(CsvFieldMapper, CsvFieldMapper)
+   */
+  default CsvFieldMapper andThen(CsvFieldMapper other) {
+    return compose(this, other);
+  }
+
+  /**
+   * Composes two mappers into one that will execute them in order and
+   * return their combined output
+   */
+  static CsvFieldMapper compose(CsvFieldMapper a, CsvFieldMapper b) {
+    return (formName, localId, workingDir, model, maybeElement, configuration) -> Stream.concat(
+        a.apply(formName, localId, workingDir, model, maybeElement, configuration),
+        b.apply(formName, localId, workingDir, model, maybeElement, configuration)
+    );
+  }
 }
--- a/src/org/opendatakit/briefcase/export/CsvFieldMapper.java
+++ b/src/org/opendatakit/briefcase/export/CsvFieldMapper.java
@@ -24,16 +24,16 @@ import org.opendatakit.briefcase.reused.Pair;
  * This Functional Interface represents the operation of transformation of a
  * submission field's value to a stream of CSV key-value pairs.
  * <p>
- * The {@link CsvFieldMapper#apply(String, Path, Model, Optional, ExportConfiguration)} returns
+ * The {@link CsvFieldMapper#apply(String, String, Path, Model, Optional, ExportConfiguration)} returns
  * a list of column name and value pairs because we need to support a weird empty/null
  * value encoding scheme described <a href="https://github.com/opendatakit/briefcase/blob/master/docs/export-format.md#non-empty-value-codification">in the docs</a>.
  * <p>
- * Normally, the {@link CsvFieldMapper#apply(String, Path, Model, Optional, ExportConfiguration)} should return just a
+ * Normally, the {@link CsvFieldMapper#apply(String, String, Path, Model, Optional, ExportConfiguration)} should return just a
  * {@link Stream} of {@link String} values.
  */
 @FunctionalInterface
 interface CsvFieldMapper {
   // TODO Normalize the weird empty/null value encoding scheme and simplify this method to return a Stream<String> of just csv column values
   // TODO Simplify args by passing the ExportConfiguration object
-  Stream<Pair<String, String>> apply(String localId, Path workingDir, Model model, Optional<XmlElement> maybeElement, ExportConfiguration configuration);
+  Stream<Pair<String, String>> apply(String formName, String localId, Path workingDir, Model model, Optional<XmlElement> maybeElement, ExportConfiguration configuration);
 }
--- a/src/org/opendatakit/briefcase/export/CsvFieldMapper.java
+++ b/src/org/opendatakit/briefcase/export/CsvFieldMapper.java
@@ -17,6 +17,7 @@ package org.opendatakit.briefcase.export;
 
 import java.nio.file.Path;
 import java.util.Optional;
+import java.util.function.Function;
 import java.util.stream.Stream;
 import org.opendatakit.briefcase.reused.Pair;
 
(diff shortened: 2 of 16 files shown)


ggalmazor:
Thanks for the review, @dcbriccetti!

lognaturel:
> Users could miss the audit-N.csv files we were previously writing into the media output directory

An alternate option would be to export both. The advantage is that it means no user has to change their analysis workflow. We know from Collect analytics that not a ton of unique forms make use of audits but there are a few forms that are deployed at quite large scale. I have to assume they use Briefcase because Aggregate doesn't offer much of an opportunity for analysis.

The downsides are that it is one more feature to maintain and I assume it has some kind of impact on performance. How significant do you think these downsides are, @ggalmazor?

ggalmazor:
> The downsides are that it is one more feature to maintain and I assume it has some kind of impact on performance. How significant do you think these downsides are, @ggalmazor?

In terms of the code, I don't think it has an impact because the code that produces the audit files is the same that deals with binary attachments e.g. we wouldn't have to maintain code just for exporting the audit files as we've always been doing.  We would just have to let the code flow instead of preventing the `BINARY` field handlers process the autit files.

Regarding performance, yes, writing twice will always be slower than writing once, but this PR's impact on performance draws mainly from the processing of the unified file, not from copying the individual audit files to the output `media` folder.

In any case, this doesn't affect forms without audit info and if it becomes a problem, we could always add a toggle to enable any of the features.

lognaturel:
👍 Given that information, my preference would be to do both. This feels more inline with how we deal with exports in general -- we keep the source and produce a nicely-formatted artifact.

yanokwa:
@lognaturel your reasoning is not correct. We keep the source (e.g., submission.xml) in the storage data, but we always produce an artifact (e.g., all the submissions in the CSV). We do always pass through the media, but the audit.csv is a special kind of media. Also, while there are links to the individual audit files in the submission CSV, they are named in a fragile way (audit-N.csv). That's why @ggalmazor filed the issue to begin with.

I would like to get rid of the audit.csv in the media folder because:
1. It's hard to imagine a scenario where someone using these individual audit files wouldn't prefer the single file. Also, they are in the storage directory anyway as is.
2. We don't have to worry about what to call the files since they won't be there :)

Note that removing audit media is a breaking change so we should probably increment a major version. But if no one is using it, then we can probably make the change without incrementing a major version. For now, I think we:
1. Merge what we have now and leave the individual audit files.
2. Add analytics to see who is exporting audit files.
    * If we see a lot of users, we add a notice on export that audit.csv is going away.
    * If we don't see a lot of users, we don't add a notice.
3. Remove audit.csv in the next major version.

What say ye?

lognaturel:
> We do always pass through the media, but the audit.csv is a special kind of media.

That's fair.

> Merge what we have now and leave the individual audit files. 

This PR no longer exports individual media file. Do you mean that it should be updated to continue doing that export?

Other than that, the approach seems good.

ggalmazor:
> Merge what we have now and leave the individual audit files.

@yanokwa, do you mean to merge the PR as is? "What we have now (in this PR)" is an aggregated audit file and no individual `audit-N.csv` files in the `media` folder.

yanokwa:
Ugh, I'm sorry. Change the PR to allow individual audit-N.csv files and merge that.

ggalmazor:
There's a little complication I didn't think about...

The csv files have a column for the location of the output audit file. This PR proposed having the same value on each row, with the aggregated audit filename (previously we had the path to the individual audit file in the `media` folder).

Now, since we're having both side effects, we need to decide what to do with the audit column(s) in the output csv file. These are the options:
- Have two columns, one for each file (easier, no-extra-work-for-me, no-missing-columns-only-extra-columns-for-users option)
  - we should add an "-aggregated" suffix to the audit column to hold the path to the aggregated file
- Have the same column as before, with the paths to "old" individual files in the `media` output dir
  - Tradeoff: the aggregated file could be missed by users, as there's no mention about it on any csv.
- Have one column, with the path to the aggregated audit file
  - Tradeoff: the individual files could be missed by users, as there's no mention about them on any csv.

yanokwa:
I'd vote for having the same column as before (Option 2). I don't think people will miss the audit file because it's right there next to the data export. And I think when we remove the individual audit files, we just drop that column.

ggalmazor:
OK, so now it should output only a column with the path to the individual audit file in the `media` output folder.

ggalmazor:
Sorry for not waiting for your feedback, @yanokwa! I wanted to reuse testing artifacts from this PR to start working on the GeoJSON issue.

PR #643, opened 2018-09-13
Issue 595 normalize startup behavior

### LINKED ISSUE #595: Normalize startup behavior when a stored server params for pulling aren't good
#### Software versions 
Briefcase v1.11.2

#### Problem description
After a user pulls forms from a server and exits Briefcase (provided she has toggled the "remember passwords" option), the next launch could fail:
1. The saved credentials could have changed server-side
2. The related granted permissions could have changed server-side
3. The server could be malfunctioning
4. There could be some networking problem
5. The server could be non-responding

When Briefcase is starting up, it will try to use the saved server params to preload the UI with its available forms.

- In scenarios 1, 2, and 3, the server will respond with a non-success (2xx) HTTP status. This will produce an empty forms list which can be confusing for the user.
- In scenarios 4, and 5, there will be a connection timeout, which will produce a program failure.

To reproduce scenarios 1, or 2, you can create a new user in Aggregate and change its password and/or granted permissions between Briefcase runs.

To reproduce scenario 5, you can stop Aggregate between Briefcase runs.

Scenarios 3, and 4 are hard to test and won't really help to triangulate the problem (scenarios 1, 2, and 5 should be enough).

#### Expected behavior
When there's any problem with a saved server conf, the server conf should be removed from prefs and the UI should be reset (no set server params, no forms on the table).

Optionally, we could show the user a dialog telling her about what just happened. Something like "The saved Pull source has been removed due to errors while trying to load the list of forms. This could be due to changes to the saved credentials, or to a problem at the source. Please, set your Pull source again."

#### Other information 
#591 introduces the changes to avoid the program's failure on scenarios 4, and 5.

ggalmazor:
Closes #595

I've manually verified each scenario described in the issue:
- When user credentials change between Briefcase launches
- When the server somehow malfunctions or it's missing

This continues the work started by @dcbriccetti and @jpknox to avoid failures and provide users with options and feedback when something goes wrong.

- Users should expect an error dialog whenever there's an error while reloading (or preloading, if Briefcase is starting) forms from a configured source.
- The underlying errors should appear on the logs
- No sentry events should be sent due to this errors.

No.


src/org/opendatakit/briefcase/ui/pull/PullPanel.java
```diff
@@ -75,17 +78,28 @@ public PullPanel(TransferPanelForm view, TransferForms forms, BriefcasePreferenc
         updateActionButtons();
       } catch (HttpException e) {
         log.warn("Unable to get form list from {}: {}", source.getDescription(), e.toString());
+        new SwingWorker() {
```
dcbriccetti: It’s not clear to me why you’re using another thread here. What happens when you simply call showMessageDialog?  Perhaps a comment? The commit message, `- On startup, we need to launch it into a background thread to avoid blocking the UI.` suggests (to me) you’re doing the `getFormList` in another thread, but you aren’t.

src/org/opendatakit/briefcase/ui/reused/UI.java
```diff
@@ -64,6 +68,33 @@ public static JButton cellWithButton(JTable table, Object value, boolean isSelec
...
+  public static void infoMessage(String message) {
+    infoMessage(APP_NAME, message, false);
+  }
+
+  public static void infoMessage(String title, String message, boolean nonBlocking) {
+    if (nonBlocking) {
+      new SwingWorker() {
+        @Override
+        protected Object doInBackground() {
+          JOptionPane.showMessageDialog(buildDialogParent(), message, title, PLAIN_MESSAGE);
```
dcbriccetti: Aren’t you violating this? https://docs.oracle.com/javase/7/docs/api/javax/swing/package-summary.html#threading

src/org/opendatakit/briefcase/ui/pull/PullPanel.java
ggalmazor: Thanks! I'll add a comment.

Swing's threading model is kind of broken and if I don't do that it will show the dialog before rendering the main frame, which is really confusing.

This way, the main frame gets rendered and the error dialog shows on top.

ggalmazor:
Thanks for the review, @dcbriccetti! I've changed the `UI` following your suggestions.

dcbriccetti:
I think this is happening at the wrong time. I started Briefcase with my Briefcase directory having been deleted.
screen shot 2018-09-17 at 8 36 29 am

ggalmazor:
I'll fix those caps. 

I think you're getting to that state because you've still got the prefs saved from a previous session. A clean reset involved deleting the Briefcase Storage directory *and* clearing the prefs with the CLI `-c` flag.

dcbriccetti:
Did you consider, in the PullPanel constructor, doing *this* inside a SwingWorker?
`forms.load(source.getFormList());`
That would allow the JFrame to initialize and you could skip the dialog threading shenanigans.

ggalmazor:
But then I'd have to run all the `try` block inside the worker, right?

I don't like how asymmetric the code would result comparing to the `view.onSource` callback defined after that. It's the same operation, but without the SwingWorker stuff and we don't have to worry about threading because it's always executed as a result of a UI interaction event.

Actually, what I really want to do is a good-old `view.onReady(callback)`, I could hide the Swing Worker stuff in there. What do you think?

Also, since this is a void operation, would it make sense to use `invokeLater` instead?

dcbriccetti:
I’m working on some code for you.

dcbriccetti:
Here’s what I recommend. What do you think?
https://github.com/dcbriccetti/briefcase/commit/0e8468533e2a682ec16bbdd194a2a1815ed5e091

ggalmazor:
Thanks for the code examples! I'll study them.

ggalmazor:
@dcbriccetti I've just checked that using a null parent for the dialog won't put the dialog on top on Mac :(

ggalmazor:
> Here’s what I recommend. What do you think? dcbriccetti@9d0c5e7?w=1

If I understand correctly, a SwingWorker distributes an operation to the correct threads for each thing:

- Current thread: The execute() method is called on this thread. It schedules SwingWorker for the execution on a worker thread and returns immediately. One can wait for the SwingWorker to complete using the get methods.
- Worker thread: The doInBackground() method is called on this thread. This is where all background activities should happen. To notify PropertyChangeListeners about bound properties changes use the firePropertyChange and getPropertyChangeSupport() methods. By default, there are two bound properties available: state and progress.
- Event Dispatch Thread: All Swing related activities occur on this thread. SwingWorker invokes the process() and done() methods and notifies any PropertyChangeListeners on this thread.

  (copied from this [post](https://www.codementor.io/isaib.cicourel/swingworker-in-java-du1084lyl))

Some extra context:
- The only block that will ever run outside the EDT is the PullPanel constructor, and that's why it's important to launch the dialog with `invokeLater()`, which ensures that it will run on the EDT. (same happens with all the other components' constructors)

- The method `sources.getFormList()` can become a long-running process that could block the UI, and that's why we want to launch it in another thread.

- We have the same problem with other long-running processes:
  - Exporting forms
  - Pulling forms
  - Pushing forms
  - Maybe even refreshing the forms cache

- The same problem exists in the context of the CLI

I'm hesitant to couple ourselves to Swing to solve our concurrency needs. 
- We know we wouldn't be able to use it on the CLI, forcing us to solve the same problem twice, using different stacks.
- We know that Swing's threading model is kinda old and wrong: it's not thread-safe 
 and it feels like everything depends on not forgetting to use SwingWorkers and invokeLaters. It doesn't take complexity away and [threading is hard](https://www.google.es/search?q=threading+is+hard&oq=threading+is+hard&aqs=chrome..69i57j0l5.3382j0j7&sourceid=chrome&ie=UTF-8). I'd prefer a more safer & closed system that "just works the way we want" (the UI class is one example of this).
- Briefcase needs a longstanding revamp of its concurrent code to take advantage of modern Java8 features like `CompletableFutures`. Maybe this could be a first run at this.

ggalmazor:
@dcbriccetti, I think I've improved the code ensuring that all UI side-effects run on the EDT and all long-running processes run in a background thread (using CompletableFuture). What do you think?

src/org/opendatakit/briefcase/ui/reused/UI.java
```diff
@@ -72,62 +72,48 @@ public static JButton cellWithButton(JTable table, Object value, boolean isSelec
...
-  public static void errorMessage(String title, String message, boolean blockUI) {
-    openDialog(blockUI, () -> JOptionPane.showMessageDialog(buildDialogParent(), buildScrollPane(message), title, ERROR_MESSAGE));
+  public static void errorMessage(String title, String message) {
+    openDialog(() -> JOptionPane.showMessageDialog(buildDialogParent(), buildScrollPane(message), title, ERROR_MESSAGE));
   }
 
-  private static void openDialog(boolean blockUI, Runnable dialog) {
-    if (blockUI)
+  private static void openDialog(Runnable dialog) {
+    if (SwingUtilities.isEventDispatchThread())
```
dcbriccetti: Ah, nice.

src/org/opendatakit/briefcase/ui/pull/PullPanel.java
```diff
@@ -69,30 +68,30 @@ public PullPanel(TransferPanelForm view, TransferForms forms, BriefcasePreferenc
...
       Source.clearAllPreferences(tabPreferences);
       source.storePreferences(tabPreferences, getStorePasswordsConsentProperty());
-      try {
-        forms.load(source.getFormList());
+      source.getFormList().thenAccept(formList -> {
+        forms.load(formList);
         view.refresh();
         updateActionButtons();
-      } catch (HttpException e) {
+      }).onError(cause -> {
```
dcbriccetti: My example eliminated this duplication, and I think doing that is important.

src/org/opendatakit/briefcase/ui/pull/PullPanel.java
```diff
@@ -72,11 +72,9 @@ public PullPanel(TransferPanelForm view, TransferForms forms, BriefcasePreferenc
...
         view.refresh();
         updateActionButtons();
-      }).onError(e -> {
-        log.warn("Unable to get form list from {}", source.getDescription(), e);
-        // We need to launch the dialog in the background because,
-        // at this point, the main frame won't be rendered yet
-        errorMessage("Error Preloading Forms", "We haven't been able to preload forms using the saved source. Try reloading it or reset it, please.");
+      }).onError(cause -> {
+        log.warn("Unable to load form list from {}", source.getDescription(), cause);
+        errorMessage("Error Loading Forms", "We haven't been able to load forms using the saved source. Try reloading it or reset it, please.");
```
dcbriccetti: Did you consider my rewording here?

dcbriccetti:
@ggalmazor, I like how you keep the big picture in mind. I like your plan, and will take a closer look at the code later.

ggalmazor:
Awesome! Your suggestions about the JavaDocs and the error messages are in. I've only changed the adjective from "saved" to "configured", since the message has to work in both contexts: first launch and whenever someone hits reload.

dcbriccetti:
About the split infinitives: I generally don’t care about the old rule against them, but this one is split quite far.

src/org/opendatakit/briefcase/ui/reused/UI.java
```diff
@@ -69,7 +69,7 @@ public static JButton cellWithButton(JTable table, Object value, boolean isSelec
   }
 
   /**
-   * Pops an informative dialog up with some sensible defaults:
+   * Pops up an informative dialog with some sensible defaults:
```
dcbriccetti: Now yer talkin’.

src/org/opendatakit/briefcase/ui/pull/PullPanel.java
```diff
@@ -119,7 +119,7 @@ private void onSource(TransferPanelForm view, TransferForms forms, Source<?> sou
       updateActionButtons();
     }).onError(cause -> {
       log.warn("Unable to load form list from {}", source.getDescription(), cause);
-      errorMessage("Error Loading Forms", "We haven't been able to load forms using the saved source. Try reloading it or reset it, please.");
+      errorMessage("Error Loading Forms", "Briefcase hasn’t been able to load forms using the configured source. Try Reload or Reset.");
```
dcbriccetti: Sorry I didn’t think of this earlier, but “hasn’t been able to” suggests repeated attempts. I suggest “Briefcase wasn’t able to ...”.

ggalmazor:
Thanks, @dcbriccetti! I'll change that "hasn't been" and will squash some little commits to make it ready for merging

ggalmazor:
I've squashed all the UI work to save all the back and forths. Now the git history is a little bit more useful, I think :)

kkrawczyk123:
I have verified on Ubuntu, MacOS and Windows scenarios 1,2 and 5 form the issue. Everything worked as described.  On Windows I can see a bit worse dialog than on other systems, screenshot attached:
screenshot from 2018-10-04 10-41-29
I am also not very convinced about suggesting users to use Reload button in that case as they can click, click and click on it and it won't work anyway in situations like that and eventually, they will have to use a Reset button. 

kkrawczyk123:
Successful retesting, so

lognaturel:
@ggalmazor For all these PRs that are approved and verified, I think it makes most sense for you to do the merging because you are more aware of where there may be conflicts and what the best order would be!

ggalmazor:
OK!

Since we've already release v1.12.2, now we should start merging these, right?

lognaturel:
Yep, whenever you want!  🎉 Was about to start doing it but realized it'd be safer for you to pick the order and since someone has approved them, it's totally fine for you to merge from a process standpoint.

ggalmazor:
OK, tomorrow I'll proceed with The Big Merge aka Mergepocalipsis aka The Mergeoning

PR #649, opened 2018-09-25
Issue 616 split choice lists ui

### LINKED ISSUE #616: Add option to split select multiples to export UI
#577 added the functionality to the command line and it would be helpful to also have in the UI.

ggalmazor:
Closes #616 

Manually tested different scenarios of main/override export conf settings for the new "split select multiples" fields.

This PR follows the same patterns as with other overridable boolean export conf params.

Now users should be able to enable splitting select multiple fields into their output CSV export files.

https://github.com/opendatakit/docs/pull/842


yanokwa:
Your PR says this doesn't require an update to docs. That seems unlikely. We'll probably want to add some language around what we split, no?

ggalmazor:
Oh, you're right! We've started explaining export conf params in https://github.com/opendatakit/docs/pull/842

I'll add a bullet point there

ggalmazor:
I've included some language normalization to this PR, because there was a confusing mix of "explode", "split", "choice list", "select multiples", and "multiple selects".

Now everything says "split multiple selects".

This includes the CLI flag, which has changed from `-ssm` and `--split-select-multiples` to `-sms` and `--split-multiple-selects`.

I think we can afford this since this is a recent new feature in v1.12.0.

lognaturel:
I apologize for introducing the rather violent "explode." Agreed "split" is good.

Overall, this looks good to me. 

> I think we can afford this since this is a recent new feature in v1.12.0.

I'm on the fence about this. If someone did start using it as part of a script, the change would be breaking. That said, it's not documented yet so this probably wouldn't affect a ton of people. I think we should perhaps have the policy to add analytics events for new features so we can tell with certainty whether they're used or not. We've started doing that in Collect and it's really helpful for making decisions like this.

I tend to call the constructs "select multiples" because that matches the `select_multiple` type name in XLSForm. Docs seem to [call them "multi select"](https://docs.opendatakit.org/form-question-types/#multi-select-widget). What makes you prefer "multiple select" to "select multiple", @ggalmazor? I guess I don't really have a super strong feeling here but it'd be good to make a principled choice and then try to be consistent elsewhere.

I don't know how relevant this is but it bothers me slightly that `sms` has meaning. I don't think Briefcase will ever do anything related to text messages but still wanted to call it out.

ggalmazor:
Thanks, @lognaturel!

I liked "explode"! :)

I used "split multiple selects" in the UI because "split select multiples" didn't sound right to my non-English-native ears. Then I tried to get everything conforming to what users see on the UI. I could have gone the other way and use what we were showing on the CLI as well.

I think you're right and we should keep it so that it forms `ssm` on the CLI. I'll change everything to "split select multiples".

ggalmazor:
I've changed language to "split select multiples" and I've dropped the commit that made the change to the CLI flag. Now it remains `-ssm` and `--split_select_multiples`

kkrawczyk123:
**Testes with success!**

PR #652, opened 2018-09-26
Issue 651 fixed the busy cursor after error happens

### LINKED ISSUE #651: Busy cursor still display after the errors.
#### Software versions 
macOS 10.13.6 

#### Problem description
I'm trying to connect to the demo server (http://opendatakit.appspot.com), but I got a connections time out error, after click the "Yes" button, the cursor still busy in the dialog, it's uncomfortable.

image
image

1. Let the connection time out or other connection errors happen.
2. Close the error dialog and back to the connection dialog.
3. The connection busy cursor should disappear but it still active.

#### Expected behavior
The progress bar should be disappeared after the error happens.

#### Other information 
@opendatakit-bot claim

huangyz0918:
Closes #651 

We set the cursor to waiting after click the connection button but forget to set it back when an error happens, it's a small improvement but makes the user more comfortable when backing to the configuration dialog. 

It's the best solution, and it's really simple.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/reused/source/RemoteServerDialog.java
+++ b/src/org/opendatakit/briefcase/ui/reused/source/RemoteServerDialog.java
@@ -18,7 +18,6 @@ package org.opendatakit.briefcase.ui.reused.source;
 
 import static org.opendatakit.briefcase.ui.ODKOptionPane.showErrorDialog;
 
-import java.awt.Cursor;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
@@ -38,11 +37,13 @@ public class RemoteServerDialog {
       form.setTestingConnection();
 
       new SwingWorker<Response<Boolean>, Void>() {
-        @Override protected Response<Boolean> doInBackground() {
+        @Override
+        protected Response<Boolean> doInBackground() {
           return serverTester.test(server);
         }
 
-        @Override protected void done() {
+        @Override
+        protected void done() {
           try {
             Response<Boolean> response = get();
             if (response.isSuccess()) {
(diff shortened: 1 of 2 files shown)


src/org/opendatakit/briefcase/ui/reused/source/RemoteServerDialog.java
```diff
@@ -60,6 +61,7 @@ private RemoteServerDialog(RemoteServerDialogForm form, RemoteServer.Test server
             }
           }
           form.unsetTestingConnection();
+          form.setCursor(Cursor.getPredefinedCursor(Cursor.DEFAULT_CURSOR));
```
ggalmazor: Maybe I'm a little bit confused, but is this really needed here?

I mean, you're already dealing with changing to the wait cursor and back to the default cursor in the `RemoteServerDialogForm.java`.

huangyz0918: Yes, but the changes in `RemoteServerDialogForm.java` only resolves the `BriefcaseException` and `MalformedURLException`, unexpected errors like `connection time out` cannot be covered. The change here is for those unexpected.

ggalmazor: OK, I understand now, although we shouldn't really interact directly with the UI (use Swing stuff) outside the `*Form` classes. 

Do you think we could hide resetting the cursor inside the `RemoteServerDialogForm.unsetTestingConnection()` method?

Maybe even setting it to waiting in the `RemoteServerDialogForm.setTestingConnection()` method.

huangyz0918: Haha, you're right, let me do the refactor and test now :)

huangyz0918:
@ggalmazor No problem.

ggalmazor:
Thanks, @huangyz0918! We'll merge this after testing :)

kkrawczyk123:
**Tested with success!**

PR #658, opened 2018-09-30
Issue 657 Fix adding csv extension when given filename doesn't include it

### LINKED ISSUE #657: Fix adding csv extension when given filename doesn't include it
#### Software versions 
Briefcase v1.12.0

#### Problem description

As discussed in [this forum post](https://forum.opendatakit.org/t/csv-file-extension-missing-on-export-via-odk-briefcase-cli-1-12/15449), v1.10 ensured that the main export file included the csv, in case the user provided a filename without it.

- Run the JAR: `java -jar build/libs/ODK-Briefcase-v1.12.1.jar -e -id all-widgets -ed /tmp/coco -f chuchu -sd /home/guillermo`
  (note that the `-f` arg defined a filename without the `csv` extension)

#### Expected behavior
The exported main CSV file should have `csv` extension.

#### Other information 
N/A

ggalmazor:
Closes #657

Add and run a new automated test that ensures the filename has the `csv` extension.
Manually run the JAR to verify it works.

This PR fixes a regression introduced in v1.11

No.

### DIFF
--- a/src/org/opendatakit/briefcase/export/ExportConfiguration.java
+++ b/src/org/opendatakit/briefcase/export/ExportConfiguration.java
@@ -19,7 +19,6 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.nio.file.Files.exists;
 import static java.nio.file.Files.isDirectory;
 import static java.nio.file.Files.newInputStream;
-
 import static org.opendatakit.briefcase.ui.MessageStrings.DIR_INSIDE_BRIEFCASE_STORAGE;
 import static org.opendatakit.briefcase.ui.MessageStrings.DIR_INSIDE_ODK_DEVICE_DIRECTORY;
 import static org.opendatakit.briefcase.ui.MessageStrings.DIR_NOT_DIRECTORY;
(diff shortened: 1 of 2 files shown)


ggalmazor:
It would be great to have a review of this issue to release a v1.12.2 with the fix this week.

src/org/opendatakit/briefcase/export/ExportConfiguration.java
```diff
@@ -449,7 +447,7 @@ public Path getExportMediaPath() {
   }
 
   public Optional<String> getExportFileName() {
-    return exportFileName;
+    return exportFileName.map(filename -> filename.endsWith(".csv") ? filename : filename + ".csv");
```
dcbriccetti: Might there be upper or mixed case “csv”?

ggalmazor: Good point. I’ll take that into account

ggalmazor:
@dcbriccetti, I've added new test cases to ensure we don't add extra extensions when the user provides some uppercase variant of `.csv`.

PR #659, opened 2018-10-01
Issue 646 fix repeat csv key generation

ggalmazor:
As per conversation [in the forums](https://forum.opendatakit.org/t/odk-meta-repeat-groups-not-merging/15308/20?u=ggalmazor) it seems that I missed a detail on the patch we published in v1.12.1.

Nested repeat groups should include all the ancestor group names until we hit another repeat group or the model's root.

Given this model:

* `group1`
  * `group2`
    * `group3`
    * `repeat1`
  * `repeat2`
    * `group4`
    * `repeat3`
* `repeat4`
  * `group5`
    * `group6`
    * `repeat5`
  * `repeat6`
    * `group7`
    * `repeat7`

These are the expected file names:
```
nested.csv
nested-group1-group2-repeat1.csv
nested-group1-repeat2.csv
nested-repeat3.csv
nested-repeat4.csv
nested-group5-repeat5.csv
nested-repeat6.csv
nested-repeat7.csv
```

New test cases have been added ensuring this behavior.

This is a missing piece on the patch released in v1.12.1 for the same issues

This is a bugfix.

However, we're discussing the option of improving this algorithm to make it more robust:
- Currently, dupes can occur. We should replace any "-" character in group names to something else.
- Currently, we don't use an FQN for the repeat group. Maybe we should use an FQN in expense of longer filenames.

No.


lognaturel:
On hold pending final resolution at https://forum.opendatakit.org/t/odk-meta-repeat-groups-not-merging/15308/21. We might leave the current behavior as it is.

ggalmazor:
It looks like we've got consensus about this on the [forum](https://forum.opendatakit.org/t/odk-meta-repeat-groups-not-merging/15308/21) and we should merge this one. I can start work on a v1.12.2 release as soon as we get the PR accepted.

lognaturel:
Sorry about the delay on this!

PR #673, opened 2018-10-22
[WIP] Issue 668 filename sanitizing like Central

### LINKED ISSUE #668: Use the same filename sanitizing as in Central
We should sanitize files using the same algorithm. In Central we're using  https://www.npmjs.com/package/sanitize-filename, which seems to be better tested and documented (and less restrictive without being unsafe) than the one we're using in Briefcase, which was custom made.

ggalmazor:
Add a new test and a new sanitizing method
- Test cases taken from the sources from https://github.com/parshap/node-sanitize-filename

Closes #668


ggalmazor:
Some thoughts:
- Briefcase feels too restrictive with punctuation characters
- Central deals with Windowsy COM, LPT, and friends, although we could argue that this is more important in Briefcase than in Central
- Central cleans heading and trailing spaces and dots, which is nice.
- Both deal correctly with non-visible, whitespacey characters, although `\p{Punct}` seems nicer
- Both support UTF-8 non-alphanumeric characters, which is nice.

Some of the sanitizing cases in Central's library feel somewhat arbitrary, but I wouldn't worry much about it anyway.

lognaturel:
Would love to know more about the feedback you're looking for and what work might be remaining! It looks complete to me. Are you wondering whether the switch should be made? What are some examples of sanitizing cases you might find arbitrary? Can you think of useful/valid repeat names that would no longer be supported?

ggalmazor:
I guess after implementing it I wasn't sure this brings an overall improvement. The regexps we're using now are more robust but the library takes care of trimming the filename and handling reserved filenames in Windows.

The library chooses to remove the offending chars instead of replacing them with a space or an underscore, which is what we're using now.

Also, Briefcase adds the extension to the filename in a later phase, which doesn't play well with the library, which assumes you're sanitizing a full filename.

In an ideal world with infinite time I'd suggest we take what's best from both options and change both Briefcase and Central, but I can see how this could be an inconvenient. I'd take:
- The regexps and character sustitution options from Briefcase
- The trimming and the support for Windows edge-cases from the library

lognaturel:
Thanks for the extra details! I'm not sure I have a strong feeling about this. What do you mean by the original regex being more robust?

I guess the primary goal is for the CSVs to be written at that point as opposed to the system crashing. Do we also want to try to make sure that downstream tools like odkmeta can work with scrubbed filenames? @matthew-white? If not, then I don't think the particular approach matters all that much as long as it fulfills the non-crash goal.

ggalmazor:
The original regexps we use in Briefcase use POSIX character classes like `\p{Punct}` or `\p{Whitespace}` which are more generic/complete/robust

ggalmazor:
Currently we are safe from errors that would make Briefcase crash when writing files into the filesystem. The goal for this PR would be to _have Briefcase and Central follow the same filename sanitizing process_ so that they output the same artifacts.

yanokwa:
I'm OK making Briefcase match Central.

> The library chooses to remove the offending chars instead of replacing them with a space or an underscore, which is what we're using now.

The tests don't suggest this is true. `stripIllegalChars` definitely replaces with a space, but `stripIllegalChars2` only replaces some with a space. Am I missing something?

> Also, Briefcase adds the extension to the filename in a later phase, which doesn't play well with the library, which assumes you're sanitizing a full filename.

This seems fine, that is all we are adding is ".csv" which seems pretty safe. I'd be concerned about length, but the library doesn't do that anyway.

ggalmazor:
I've fixed the tests and implementation according to the results I got from the original JS library. 

If there's no objection, I'm happy with it and I think it could be reviewed. 

We should also discuss whether it should go into v1.14 or v2.0 (since it comes with breaking changes)

test/java/org/opendatakit/briefcase/util/StringUtilsTest.java
```diff
@@ -1,53 +1,120 @@
...
+        {"Reserved Windows filename LPT1", "LPT1", ""},
+        {"Reserved Windows filename LPT2", "LPT2", ""},
+        {"Reserved Windows filename LPT3", "LPT3", ""},
+        {"Reserved Windows filename LPT4", "LPT4", ""},
+        {"Reserved Windows filename LPT5", "LPT5", ""},
+        {"Reserved Windows filename LPT6", "LPT6", ""},
+        {"Reserved Windows filename LPT7", "LPT7", ""},
+        {"Reserved Windows filename LPT8", "LPT8", ""},
+        {"Reserved Windows filename LPT9", "LPT9", ""},
+        {"Reserved Windows filename con", "con", ""},
```
dcbriccetti: I’d rather see these lower-case versions generated rather than duplicated, but I suppose if the items aren’t expected to change much or at all, there’s little risk someone will change one and not the other.

yanokwa:
I'm leaning towards putting this in v2 because
1. it's a breaking change
2. it's unlikely to cause conflicts if the PR sites
3. Central isn't widely used and no one is really asking to match that form

test/java/org/opendatakit/briefcase/util/StringUtilsTest.java
ggalmazor: Thanks for the suggestion! It's better :)

ggalmazor:
Moved to the v2.0 milestone and project.

kkrawczyk123:
My testing results:

| **form name** | **result ubuntu** | **result mac** | **result windows** |
|---|---|---|---|
|trailing spaces| _badForm.xml| _badForm.xml|_badForm.xml|
| <.-??>,. | .-,..xml | .-,..xml | .-,.xml |
|slash/test|slashtest.xml|slashtest.xml|slashtest.xml|
|dotdot.......|dotdot|dotdot|dotdot|
|dot.test|dot.test|dot.test|dot.test|
|PRN|forms.xml|forms.xml|forms.xml|
|CON|forms.xml|forms.xml|forms.xml|
|traling dots|forms.xml|forms.xml|forms.xml|
|traling slashes|forms.xml|forms.xml|forms.xml|

All forms.xml ale created without a folder, just in sd with folder instances. When I already have one forms.xml in my sd and I am pulling another one - nothing is happening, there is no another forms.xml in sd and Success in pull tab is visible. Definitely not sure if this are a success results.
cc @ggalmazor and @yanokwa 

ggalmazor:
@kkrawczyk123 could you make a zip file with all the forms you used and attach it here so that I can work with them, please?

kkrawczyk123:
@ggalmazor, of course! 
[weird names forms.zip](https://github.com/opendatakit/briefcase/files/2667611/weird.names.forms.zip)

PR #674, opened 2018-10-25
Issue 671 export param to remove group names

### LINKED ISSUE #671: Add an option to Briefcase export to remove group names in column headers.
**Background**
When exporting a form through Briefcase, variable names (column headers of the csv file) are not simply the one specified by the user on designing the form, but the eventual groups name the variable belong to, are automatically added before the actual name given to the variable.

In complex forms, with nested groups, this could lead in very long and complex names of column headers of the exported csv file.

These complex names bring complexity in the analysis and generate headache to data managers.

The reason of generating these complex names when exporting data from Aggregate is due to the XForm specs that give the possibility to have variables with same name inside different groups.

Most ODK users use XLSForms or Build to create their forms that will not allow duplicated variable names but data managers at the end need to manually (or with a specific manipulation post export) remove the group names for analysing the data. I

[Forum Post
](https://forum.opendatakit.org/t/allow-more-control-over-variable-naming-syntax/15192/11)

**Goal**

a) Add an option to Briefcase to export without group names.

These option will work for forms that have been created with XLSForm and Build or with forms created manually (xml) without duplicated variable names.

In case of manually written xml files with duplicated variable names, a pop up could be shown to users after parsing the file, informing that due to duplicated variable names this kind of export can not be executed.

reference [Roadmap](https://github.com/opendatakit/roadmap/issues/27)

ggalmazor:
Closes #671

This PR is built on top #669 and the first commit is https://github.com/opendatakit/briefcase/pull/674/commits/d0aef21a97450287396ea9a41ebb0c8e1ff36cea

Try it with this Briefcase JAR: 
[briefcase-674-remove-group-names.jar.zip](https://github.com/opendatakit/briefcase/files/2561259/briefcase-674-remove-group-names.jar.zip)

Added new automated tests
Run the export of All Widgets form

Also exported [field-in-group.zip](https://github.com/opendatakit/briefcase/files/2557123/field-in-group.zip), which that has a field in a non-repeat group.

- Without `-rgn` builds this header: `SubmissionDate,meta-instanceID,group-some-field,KEY`
- With `-rgn` builds this header: `SubmissionDate,instanceID,some-field,KEY`

No other approaches considered. The code that generates the headers is very localized and the changes from this PR felt like the minimum required to get the new behavior.

Users should expect a new CLI `-rgn` (`--remove_group_names`) param to enable using unqualified header column names, which don't include group names.

Yes, we should update the docs to add the new CLI param

### DIFF
--- a/src/org/opendatakit/briefcase/export/Csv.java
+++ b/src/org/opendatakit/briefcase/export/Csv.java
@@ -56,7 +56,7 @@ class Csv {
   private static Csv main(FormDefinition formDefinition, ExportConfiguration configuration) {
     return new Csv(
         formDefinition.getModel().fqn(),
-        getMainHeader(formDefinition.getModel(), formDefinition.isFileEncryptedForm(), configuration.resolveSplitSelectMultiples()),
+        getMainHeader(formDefinition.getModel(), formDefinition.isFileEncryptedForm(), configuration.resolveSplitSelectMultiples(), false),
         buildMainOutputPath(formDefinition, configuration),
         true,
         configuration.resolveOverwriteExistingFiles(),
@@ -67,7 +67,7 @@ class Csv {
   private static Csv repeat(FormDefinition formDefinition, Model groupModel, ExportConfiguration configuration, Path output) {
     return new Csv(
         groupModel.fqn(),
-        getRepeatHeader(groupModel, configuration.resolveSplitSelectMultiples()),
+        getRepeatHeader(groupModel, configuration.resolveSplitSelectMultiples(), false),
         output,
         false,
         configuration.resolveOverwriteExistingFiles(),
--- a/src/org/opendatakit/briefcase/export/Csv.java
+++ b/src/org/opendatakit/briefcase/export/Csv.java
@@ -56,7 +56,12 @@ class Csv {
   private static Csv main(FormDefinition formDefinition, ExportConfiguration configuration) {
     return new Csv(
         formDefinition.getModel().fqn(),
-        getMainHeader(formDefinition.getModel(), formDefinition.isFileEncryptedForm(), configuration.resolveSplitSelectMultiples(), configuration.resolveRemoveGroupNames()),
+        getMainHeader(
+            formDefinition.getModel(),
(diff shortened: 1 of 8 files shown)


yanokwa:
I haven't looked at the PR, but `use_unqualified_names` seems very opaque. I would suggest "remove-group-names" or "ignore-group-names"

ggalmazor:
I've changed it to `--remove_group_names` as suggested

lognaturel:
I get that rm is often short for remove but it feels odd to me to see `-rmg` as the short form of `--remove_group_names`. I'd go with either `-rgn` or `rg`.

I tried this with All Widgets and am getting group names on all the questions. Verified I'm running off commit d63df74 but haven't had a chance to look more deeply yet.

ggalmazor:
Sorry, `rmg` was a mistake. I'll fix that

ggalmazor:
Thanks for the review, @lognaturel!

You can use [field-in-group.zip](https://github.com/opendatakit/briefcase/files/2557123/field-in-group.zip), which that has a field in a non-repeat group.

- Without `-rgn` builds this header: `SubmissionDate,meta-instanceID,group-some-field,KEY`
- With `-rgn` builds this header: `SubmissionDate,instanceID,some-field,KEY`

lognaturel:
Hmm, I know what happened -- I wanted to see the diff so ran it once without `-rgn`. Then I ran it again with `rgn`. But I didn't use `-oc` so the file wasn't recreated. That makes sense in retrospect but could be unexpected. I don't have any good suggestions for dealing with that. It's probably fine.

Approving the new commits! Once #671 is merged, would be good to rebase here and you can mark for a quick QA pass to see if there are any other odd cases to be aware of.

ggalmazor:
I'll wait for #669 to be merged to solve the conflicts on this one.

kkrawczyk123:
I have noticed that group name of geopoints widgets are not removed when using -rgn parameter.  I used All widgets form. The same results on Ubuntu, MacOS and Windows.

kkrawczyk123:
**Tested with success!**

PR #691, opened 2018-12-05
Issue 688 Make pull and push urls clickable in the UI

### LINKED ISSUE #688: Make aggregate's URL clickable
#### Problem description or more improvement suggestion
I was wondering if we could do aggregate's URL clickable on the pull and push tab or at least make it possible to copy. 
It currently looks like: 
screenshot from 2018-11-28 11-26-39
screenshot from 2018-11-28 11-21-57
I think that hyperlink would be a very comfortable solution. It would also make an URL more visible in comparison to the other visible text because now it is not distinguishable nearly at all.
@yanokwa @ggalmazor what are your thoughts on that?

ggalmazor:
Closes #688

Tested manually all possible sources on pull and push tabs.

I researched different options to make stuff clickable in Swing and I went for this one because it was the simplest way to get it done. Other alternatives involve more complex listeners and/or using custom-styled buttons instead of labels.

This PR makes clickable the URLs to the Aggregate sources/targets configured on the Pull and Push tabs. They will open the default browser.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/reused/source/Source.java
+++ b/src/org/opendatakit/briefcase/ui/reused/source/Source.java
@@ -16,11 +16,20 @@
 
 package org.opendatakit.briefcase.ui.reused.source;
 
+import static java.awt.Cursor.HAND_CURSOR;
+import static java.awt.Cursor.getPredefinedCursor;
+import static java.awt.Desktop.getDesktop;
+import static javax.swing.SwingUtilities.invokeLater;
 import static org.opendatakit.briefcase.ui.reused.FileChooser.isUnderBriefcaseFolder;
 import static org.opendatakit.briefcase.ui.reused.UI.errorMessage;
 
 import java.awt.Container;
+import java.awt.Cursor;
+import java.awt.event.MouseListener;
 import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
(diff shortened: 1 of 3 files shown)


src/org/opendatakit/briefcase/ui/reused/source/Source.java
```diff
@@ -267,6 +279,23 @@ public String getDescription() {
...
+    private static void uncheckedBrowse(URL url) {
+      try {
+        getDesktop().browse(url.toURI());
+      } catch (URISyntaxException | IOException e) {
+        throw new BriefcaseException(e);
+      }
+    }
+
+    @Override
+    public void decorate(JLabel label) {
```
dcbriccetti: decorate is getting called twice, so clicking on the link creates two browser tabs.

ggalmazor: Good catch, thanks! I'll take a look

ggalmazor: Solved! Thanks again!

kkrawczyk123:
Hhaha @ggalmazor you know me xd I couldn't wait long haha 
**Tested with success!!**
Verified on Ubuntu, MacOS and Windows and I like it very much :)

PR #692, opened 2018-12-05
Issue 687 export ui conf param remove group names

### LINKED ISSUE #687: Add (export) remove group names conf param to the UI
After Briefcase v1.13.0 we have a new CLI param to enable removing group names from CSV headers when exporting forms.

We should add a new UI conf param to enable it as well:
- It should be a checkbox on the default conf panel
- It should be an `OverridableBoolean` on the override (custom) conf panels.

To go about this, we can imitate what's being done with other boolean conf params like "export media" or "include GeoJSON"

ggalmazor:
Closes #687

Tested manually with [choice-lists.zip](https://github.com/opendatakit/briefcase/files/2648642/choice-lists.zip), which has a field in a group.
Tested all combinations of default conf and overridden conf.

This PR imitates what's being done with all the other overridable boolean export conf params.

Users should be able to enable removing the group names from the CSV header lines thanks to a new checkbox (default conf panel) and overridable boolean (custom conf panel)

image

image

Yes:
- [ ] https://github.com/opendatakit/docs/issues/927

### DIFF
--- a/src/org/opendatakit/briefcase/export/ExportConfiguration.java
+++ b/src/org/opendatakit/briefcase/export/ExportConfiguration.java
@@ -212,6 +212,10 @@ public class ExportConfiguration {
     return includeGeoJsonExport;
   }
 
+  public OverridableBoolean getRemoveGroupNames() {
+    return removeGroupNames;
+  }
+
   public void ifExportDirPresent(Consumer<Path> consumer) {
     exportDir.ifPresent(consumer);
   }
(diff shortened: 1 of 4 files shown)


kkrawczyk123:
**Tested with success!**

ggalmazor:
Thanks for the review, @dcbriccetti! 
Thanks for testing it, @kkrawczyk123! 

:heart:

PR #698, opened 2019-02-01
Issue 681 resume last pull

### LINKED ISSUE #681: Resume last pull
Users with big forms and/or many submissions (10k, 25k, and more), often complain about how much time the pulling and exporting processes take to complete.

([forum conversation](https://forum.opendatakit.org/t/pull-submissions-starting-from-date/16200))

This problem can be worse in scenarios where users have a long-living form that receives periodically new submissions i.e. to receive the last 100 submissions, Briefcase has to navigate through the first 10k to reach them.

(note that Briefcase doesn't actually download the 10k submissions it already pulled before, see the explanation below)

Some insights about how Briefcase pulls forms and their submissions from Aggregate:
- Briefcase pulls data from Aggregate in batches of a maximum of 100 submissions each.
- A batch consists of an XML  with:
  - A list of submission instance IDs
  - A "cursor" that can be used to get the next batch
- A cursor is an XML with:
  - The field used to order submissions. Currently, we're using `LAST_UPDATE_DATE`
  - The last update date (ISO8601) of the last submission from the previous batch
  - The instance ID of the last submission from the previous batch
  - A boolean telling if the cursor is a forward cursor

  ```xml
  <cursor xmlns="http://www.opendatakit.org/cursor">
    <attributeName>_LAST_UPDATE_DATE</attributeName>
    <attributeValue>2018-11-07T14:43:24.644+0000</attributeValue>
    <uriLastReturnedValue>uuid:d9b67b6f-2058-469b-8cb1-c86b9c34b632</uriLastReturnedValue>
    <isForwardCursor>true</isForwardCursor>
  </cursor>
  ```
- In order to get the first batch of submissions, Briefcase sends Aggregate an empty cursor
- Briefcase will continue asking for more batches until an empty batch arrives, which ends the pull operation

(now this is what happens with each non-empty batch received)
- Each batch will contain a maximum of 100 submission instance IDs. 
- Briefcase uses a local database to exclude from the list any submission that's been already downloaded
- For each of those remaining submission instance IDs, Briefcase downloads them from Aggregate, including all their media attachments

**Optimization problem**

Taking how Briefcase works into mind, in scenarios like the one described above, where a long-living form receives new submissions periodically, Briefcase will be forced to "scan" all the submission instance IDs until it reaches a batch where new submissions start to show up.

For a form with 10k submissions, this means 100 requests before reaching a batch that actually contains new submissions. That has a negative impact on the overall performance of the operation and the monthly expenses for users hosting their Aggregate instances in App Engine.

**Proposed solution**

- Briefcase stores the last cursor used for each form
- We add a checkbox to "Always try to resume pulls" in the Settings tab
- When that's enabled, Briefcase sends the stored cursor instead of an empty one, effectively resuming the pull operation.
- Briefcase should also provide a new CLI flag to enable this feature for users launching Briefcase from the command line

ggalmazor:
Closes #681

Tested manually on Linux(Ubuntu) using an Aggregate v2.0 local server

This PR tries to store a byproduct of the pre-existing pull process that was being discarded (the cursor uses during interaction with the Aggregate API). 

There are new user facing options:
- There's a new global setting to enable resuming pulls in the Settings tab
- There's a new button in the settings tab that clears any saved resume cursors
- There's a new CLI flag `-rlp` or `--resume_last_pull` to enable the feature from the command line

Yes, https://github.com/opendatakit/docs/issues/967

### DIFF
--- a/src/org/opendatakit/briefcase/transfer/TransferForms.java
+++ b/src/org/opendatakit/briefcase/transfer/TransferForms.java
@@ -208,6 +208,11 @@ public class TransferForms implements Iterable<FormStatus> {
 
   public void setLastPullCursor(FormStatus fs, String cursor) {
     lastPullCursorsByFormId.put(fs.getFormId(), cursor);
+    triggerOnChange();
+  }
+
+  public Map<String, String> getLastPullCursorsByFormId() {
+    return lastPullCursorsByFormId;
   }
 
   @Override
--- a/src/org/opendatakit/briefcase/transfer/TransferForms.java
+++ b/src/org/opendatakit/briefcase/transfer/TransferForms.java
@@ -38,6 +38,7 @@ import org.opendatakit.briefcase.reused.Pair;
  * selection of those forms, as well as merging changes in the forms cache.
  */
 public class TransferForms implements Iterable<FormStatus> {
+  public static final String LAST_CURSOR_PREFERENCE_KEY_SUFFIX = "-last-cursor";
   private List<FormStatus> forms;
   private Map<String, FormStatus> formsIndex = new HashMap<>();
   private Map<String, String> lastPullCursorsByFormId;
@@ -79,7 +80,7 @@ public class TransferForms implements Iterable<FormStatus> {
   public void load(List<FormStatus> forms, BriefcasePreferences preferences) {
     this.forms = forms;
     this.lastPullCursorsByFormId = forms.stream()
-        .map(form -> Pair.of(form.getFormId(), preferences.nullSafeGet(form.getFormId() + "-last-cursor")))
+        .map(form -> Pair.of(form.getFormId(), preferences.nullSafeGet(form.getFormId() + LAST_CURSOR_PREFERENCE_KEY_SUFFIX)))
         .filter(pair -> pair.getRight().isPresent())
         .map(pair -> pair.map(identity(), Optional::get))
         .collect(toMap(Pair::getLeft, Pair::getRight));
--- a/src/org/opendatakit/briefcase/transfer/TransferForms.java
+++ b/src/org/opendatakit/briefcase/transfer/TransferForms.java
@@ -15,6 +15,7 @@
  */
 package org.opendatakit.briefcase.transfer;
 
+import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
 
@@ -25,9 +26,12 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Stream;
+import org.opendatakit.briefcase.model.BriefcasePreferences;
 import org.opendatakit.briefcase.model.FormStatus;
+import org.opendatakit.briefcase.reused.Pair;
 
 /**
  * This class represents a set of forms to be pulled/pushed. It manages the
@@ -36,7 +40,7 @@ import org.opendatakit.briefcase.model.FormStatus;
 public class TransferForms implements Iterable<FormStatus> {
   private List<FormStatus> forms;
   private Map<String, FormStatus> formsIndex = new HashMap<>();
-  private final Map<String, String> lastPullCursorsByFormId;
+  private Map<String, String> lastPullCursorsByFormId;
   private final List<Runnable> onChangeCallbacks = new ArrayList<>();
 
   private TransferForms(List<FormStatus> forms, Map<String, String> lastPullCursorsByFormId) {
@@ -70,10 +74,15 @@ public class TransferForms implements Iterable<FormStatus> {
 
   /**
    * Replaces the current list of {@link FormStatus} instances with
-   * the incoming list.
+   * the incoming list, and reads any saved cursor from a previous pull
    */
-  public void load(List<FormStatus> forms) {
+  public void load(List<FormStatus> forms, BriefcasePreferences preferences) {
     this.forms = forms;
+    this.lastPullCursorsByFormId = forms.stream()
+        .map(form -> Pair.of(form.getFormId(), preferences.nullSafeGet(form.getFormId() + "-last-cursor")))
+        .filter(pair -> pair.getRight().isPresent())
+        .map(pair -> pair.map(identity(), Optional::get))
+        .collect(toMap(Pair::getLeft, Pair::getRight));
     triggerOnChange();
   }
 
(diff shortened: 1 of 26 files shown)


dcbriccetti:
+2,500 −452
Oh   my   goodness.

ggalmazor:
This is mostly new stuff added with tests. I'm hesitating about splitting this PR into two. One focused on adding the new resume last pull and another one with the improvements to the jobs in parallel. What do you think?

dcbriccetti:
Either is fine with me. But I’m calling Guinness World Records about you and your PRs.

ggalmazor:
OK, then I'd prefer everything to go in one fell sweep :P

dcbriccetti:
That’ll make it easier on Guinness.

ggalmazor:
We are splitting the price, right?

dcbriccetti:
Works for me!

ggalmazor:
@dcbriccetti, after discussing this PR with @yanokwa we've decided that we prefer to go for a codefreeze tomorrow.

To make it easier, I've finally split this PR into two and left just the new feature in this one. It should contain the minimal UI and CLI changes to have the feature into Briefcase.

We will be reviewing the changes to the parallel processes code in #706, probably in the context of an upcoming v2.0.0 with further reaching changes.

src/org/opendatakit/briefcase/transfer/TransferForms.java
```diff
@@ -15,56 +15,74 @@
...
   /**
    * Replaces the current list of {@link FormStatus} instances with
-   * the incoming list.
+   * the incoming list, and reads any saved cursor from a previous pull
    */
-  public void load(List<FormStatus> forms) {
+  public void load(List<FormStatus> forms, BriefcasePreferences preferences) {
     this.forms = forms;
+    this.lastPullCursorsByFormId = forms.stream()
+        .map(form -> Pair.of(form.getFormId(), preferences.nullSafeGet(form.getFormId() + "-last-cursor")))
```
dcbriccetti: "-last-cursor" appears in three places. Did you consider making it a constant?

dcbriccetti:
It’s not part of this PR, but I offer you this anyway:
screen shot 2019-02-12 at 8 29 08 pm

dcbriccetti:
Again outside this PR, but:
```
  /**
   * Empties the {@link FormStatus} list.
   */
  public void clear() {
    forms = Collections.emptyList();
    triggerOnChange();
  }
```
Did you consider `forms.clear()`?

src/org/opendatakit/briefcase/transfer/TransferForms.java
```diff
@@ -36,34 +36,36 @@
 public class TransferForms implements Iterable<FormStatus> {
   private List<FormStatus> forms;
   private Map<String, FormStatus> formsIndex = new HashMap<>();
+  private final Map<String, String> lastPullCursorsByFormId;
```
dcbriccetti: The comment for this commit, 
`Add somewhere to store in memory cursors into TransferForms`
would be more readable as 
`... store in-memory cursors ...`

src/org/opendatakit/briefcase/transfer/TransferForms.java
```diff
@@ -206,6 +206,10 @@ public String getLastCursor(FormStatus fs) {
     return Optional.ofNullable(lastPullCursorsByFormId.get(fs.getFormId())).orElse("");
   }
 
+  public void setLastPullCursor(FormStatus fs, String cursor) {
+    lastPullCursorsByFormId.put(fs.getFormId(), cursor);
```
dcbriccetti: Why not have the form ID as the formal parameter rather than have this method concerned with a FormStatus?

src/org/opendatakit/briefcase/util/ServerFetcher.java
```diff
@@ -261,8 +264,8 @@ private ExecutorService getFetchExecutorService(Boolean pullInParallel) {
     return Executors.newFixedThreadPool(downloadThreads, new DownloadThreadFactory());
   }
 
-  private boolean downloadAllSubmissionsForForm(File formInstancesDir, DatabaseUtils formDatabase, BriefcaseFormDefinition lfd,
-                                                FormStatus fs, String websafeCursorString) {
+  private Pair<Boolean, String> downloadAllSubmissionsForForm(File formInstancesDir, DatabaseUtils formDatabase, BriefcaseFormDefinition lfd,
+                                                              FormStatus fs, String websafeCursorString) {
```
dcbriccetti: I’m trying to decide if I’m concerned that these cursors are Strings, and not of a less general type.

src/org/opendatakit/briefcase/ui/settings/SettingsPanelForm.java
```diff
@@ -52,6 +52,7 @@
   private JButton storageLocationClearButton;
   private JButton storageLocationChooseButton;
   private JCheckBox pullInParallelField;
+  private JCheckBox resumeLasPullField;
```
dcbriccetti: May I offer you a `t`?

src/org/opendatakit/briefcase/transfer/TransferForms.java
ggalmazor: You're right :)
Thanks for pointing it out

src/org/opendatakit/briefcase/transfer/TransferForms.java
ggalmazor: I guess you're right, but I'm struggling to not think about this until we can start working on a central database. All the domain is inside out. The last pull cursor should be the "Form"'s concern...

src/org/opendatakit/briefcase/util/ServerFetcher.java
ggalmazor: The ServerFetcher belongs to the legacy codebase. The new codebase will have a `Cursor` abstraction that deals with that (see #705)

ggalmazor: By the way, the cursor is an XML document that the Aggregate API serializes into another XML's tag value, which is super weird. That's why it's a String at that point.

I think the reasoning was to treat the cursor as an opaque string during HTTP interaction.

src/org/opendatakit/briefcase/ui/settings/SettingsPanelForm.java
ggalmazor: I'll take it ;)

No traps were introduced in any commit! :D

ggalmazor:
@dcbriccetti, thanks for the careful review!

I did the `forms.clear()`. Not only it's better and more expressive, but the old code was hiding a bug, since `Collections.emptyList()` has fixed size and would fail to add elements. I think you were right and that it's worth breaking the symmetry with the `TransferForms.load()` method.

As to using a `forEach` to simplify removing keys, I'd like to keep the symmetry with the previous block that sets them. Also, this code is meant to change whenever we start working on the central database anyway.

kkrawczyk123:
I've fought for long but we have *success* here!

Issue #699, opened 2019-02-05
error on pulling a form with guidance hint but without an hint


aurdipas:
#### Software versions 
Briefcase v1.13.1, Java version "1.7.0_80", operating system Windows 7, Aggregate v1.7.1

#### Problem description
I've a form on aggregate that contains some guidance hints, but without a regular hint like in the picture, when I try to pull the form I get the following error message:

"Fetching form definition
resolving against briefcase form definitions
Error parsing form definition: org.opendatakit.aggregate.exception.ODKIncompleteSubmissionData: Javarosa failed to construct a FormDef. Is this an XForm definition?"

![image](https://user-images.githubusercontent.com/5656097/52286911-6f78bf80-2969-11e9-9bd8-b43109343e70.png)

#### Expected behavior
the form should be pulled successfully.

#### Other information 
Adding a regular hint solves the issue.
It looks related to the issue already resolved for pyxform:
https://github.com/XLSForm/pyxform/issues/225

ggalmazor:
Hi @aurdipas! thanks for reporting the issue.

Briefcase currenlty runs JavaRosa 2.11.2, and I'm being able to pull the form if I update JR to the latest 2.13.0. Does this make sense to you?

ggalmazor:
Closed by #701

PR #701, opened 2019-02-07
Issue 700 Update JavaRosa

### LINKED ISSUE #700: Update to latest JavaRosa
We should take the chance to update JR to the latest release.

This should also take care of #699

ggalmazor:
Closes #700, and #699

The project compiles and Briefcase runs.
I've tested that the issue with #699 gets solved by using this form [guidance.xml.zip](https://github.com/opendatakit/briefcase/files/2840104/guidance.xml.zip)
 which fails in master (the form is from https://github.com/XLSForm/pyxform/issues/225 which was linked in #699)

It's a very small change: version bump, sync serializable class list (copied from JR)

N/A

No.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ dependencies {
         exclude group: 'commons-logging'
     }
     compile group: 'net.sf.kxml', name: 'kxml2', version: '2.3.0'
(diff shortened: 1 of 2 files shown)


kkrawczyk123:
**Tested with success!**

PR #703, opened 2019-02-07
Issue 702 Avoid checking the Location response header if the response is not an HTTP 3xx redirect

### LINKED ISSUE #702: Briefcase should not check locations if the response is not an HTTP redirect
#### Problem description
Briefcase checks the `Location` header to detect spoofing attempts unconditionally, which doesn't make sense unless the HTTP response code is in the 3xx range, which encodes a redirection request from the server.

On a side-note, Aggregate should not add `Location` headers to 2xx responses. https://github.com/opendatakit/aggregate/issues/409

The result is that, although requests reach the server and probably forms are properly pushed, you get a lot of errors in the logs and the status of the pushing operation is "partial success".

- Configure Aggregate's hostname in security.properties to your computer's LAN IP address
- Try to pull a form using a localhost URL

#### Expected behavior
No errors should be logged
Push operations should be marked as successfully completed

ggalmazor:
Closes #702

Reproduce the scenario described in #702 and run a push operation
Verify that there's no error logged and the operation completes with success on the UI.

It's a super narrow change to conditionally do the location spoof check.

This should make Briefcase more robust in cases where the users mistakenly uses the wrong ip address to work with their Aggregate servers

No

### DIFF
--- a/src/org/opendatakit/briefcase/util/AggregateUtils.java
+++ b/src/org/opendatakit/briefcase/util/AggregateUtils.java
@@ -305,30 +305,32 @@ class AggregateUtils {
         }
       }
 
-      // what about location?
-      Header[] locations = response.getHeaders("Location");
-      if (locations != null && locations.length == 1) {
-        try {
-          URL url = new URL(locations[0].getValue());
-          URI uNew = url.toURI();
-          log.info("Redirection to URI {}", uNew);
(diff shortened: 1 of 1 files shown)


kkrawczyk123:
**Tested with success!**
Verified on Ubuntu and MacOS. I haven't tried on Windows as I don't have local aggregate on it. 

PR #704, opened 2019-02-07
Issue 695 Support empty manifest URLs when pulling forms

### LINKED ISSUE #695: Invalid url error when pulling data from server
#### Software versions 
from Briefcase v1.11.0 to v1.13.1

#### Problem description
Pulling data from the server fails. Here's the detailed status 

`Fetching form definition
resolving against briefcase form definitions
Fetching form manifest
Error fetching form definition: Fetch of manifest failed. Detailed reason: Invalid url. Failed with error: no protocol: . Did you forget to prefix the address with 'http://' or 'https://' ?`

Only one of the two api calls are made to the server. 

Configure briefcase with Aggregate Server
Provide right credentials to an account with forms
Pull form

#### Expected behavior
Should pull forms with the final status 'SUCCESS'

ggalmazor:
Closes #695

**On Onadata**
- Create an account at https://odk.ona.io/
- Upload the form [pizza_questionnaire.xlsx](https://github.com/opendatakit/briefcase/files/2755542/pizza_questionnaire.xlsx)

**On Briefcase**
- Configure the aggregate server with `https://odk.ona.io/{username}` (and your Ona credentials)
- Pull form

This is a small change to check that whatever value coming on the manifestUrl is an url. This will prevent trying to get manifests when the server reports a `<manifest/>` tag when downloading a blank form (which is the default behavior in Ona).

N/A

No

### DIFF
--- a/test/java/org/opendatakit/briefcase/util/ServerFetcherTest.java
+++ b/test/java/org/opendatakit/briefcase/util/ServerFetcherTest.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2019 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.opendatakit.briefcase.util;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+import static org.opendatakit.briefcase.util.ServerFetcher.isUrl;
+
+import org.junit.Test;
+
+public class ServerFetcherTest {
+
+  @Test
+  public void knows_if_a_string_contains_a_valid_url() {
+    assertThat(isUrl(""), is(false));
+    assertThat(isUrl("foo.bar"), is(false));
+    assertThat(isUrl("some text"), is(false));
+    assertThat(isUrl("http://foo.bar"), is(true));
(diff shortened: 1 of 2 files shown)


kkrawczyk123:
I have verified pr and master with steps described in #695 on Ubuntu, Mac and Windows and on all I have the same results, successfully pulled form in all tries. I wasn't able to reproduce the issue.
The only Failed try was when I copied ona url directly (extra / at the and of the url) but I had Authentication error then. Interesting was also then even if status was failed, the form was pulled to sd. 
So pr is verified but I am not sure if additional testing won't be necessary.

dcbriccetti:
>Didn't test, but this LGTM. I would have done `fd.getManifestUrl().isEmpty()`. @dcbriccetti whatcha think?

LGTM? I searched for it but doubt you’re saying “looks like garbage to me” because that’s not your style. :-)

isEmpty instead of isUrl? Yes, assuming no self-respecting `RemoteFormDefinition` would accept and save an invalid URL. It would be nice to assume that if one is set, it is valid.

test/java/org/opendatakit/briefcase/util/ServerFetcherTest.java
```diff
@@ -0,0 +1,37 @@
...
+import static org.junit.Assert.assertThat;
+import static org.opendatakit.briefcase.util.ServerFetcher.isUrl;
+
+import org.junit.Test;
+
+public class ServerFetcherTest {
+
+  @Test
+  public void knows_if_a_string_contains_a_valid_url() {
+    assertThat(isUrl(""), is(false));
```
dcbriccetti: One day I’ll bring in proof—perhaps made up—from eye-tracking studies quantifying just how awful the above is compared to the lovely, pleasing, non-noisy `assertFalse(usUrl("")`. ;-)

ggalmazor:
> isEmpty instead of isUrl? Yes, assuming no self-respecting `RemoteFormDefinition` would accept and save an invalid URL. It would be nice to assume that if one is set, it is valid.

We don't have fancy value object features in place yet, but yes: RemoteFormDefinition should do that work for us and reject any offending value. Don't know when, but it will happen :)

kkrawczyk123:

PR #708, opened 2019-02-13
WIP: Update UI with more consistent user language

yanokwa:
Closes #

Nothing! I'm opening this to see if @ggalmazor is open to these changes.

* "Resume last pull" isn't very precise "Resume pull from last submission pulled" feels better to me.
* We use "clear" not clean in other parts of the app
* Remove extra words like All, Always, etc. 

* If someone is using the beta, we've changed the CLI interface. I think that's safe.
* The code variables don't match the UX. I think that's OK.


ggalmazor:
Thanks! The changes on the code are looking good, but I want to test it in case the new texts require some UI tweaking

yanokwa:
Now that I'm seeing the PR, I wonder if "Resume" is the correct word. It really makes it seem like if your pull errors out, then resume it form where it broke.

ggalmazor:
> Now that I'm seeing the PR, I wonder if "Resume" is the correct word. It really makes it seem like if your pull errors out, then resume it form where it broke.

We should mention that in the docs

PR #712, opened 2019-02-14
Ensure that CLI ops select the forms they need to transfer

### LINKED ISSUE #711: Cmd pull from odk dir does not work.
#### Software versions 
Briefcase v1.x14.x, Java v1.8.x, operating system, Aggregate v1.x.x, Collect v1.x.x...

#### Problem description
Cmd pull from odk dir does not work, no matter if -pc prefix used of not. There is error or fail status in cmd response o it gives a false impression of successful pull but form are not pulled into sd.

Try to pull from odk dir using cmd command, -pc prefix,

#### Expected behavior
Forms should be pulled when using -pc prefix and without it (outdated command). User should be able to pull specific form when using -id and whole odk folder content without using -id.

#### Other information 
Pulling from aggregate using cmd works well.

ggalmazor:
Closes #711 

**Context**
https://github.com/opendatakit/briefcase/pull/698/commits/7eccb568b4cba72f1da5ca3ecc42fe9b3841117f introduced a change that migrated the transfer(pull&push) code to use `TransferForms` instead of a generic `List<FormStatus>`. `TransferForms` has the notion of "selected forms" and downstream pull and push operation code does `forms.getSelectedForms()`.

The CLI wasn't adapted to this fact by accident, which introduced the bug that this PR solves.

Run the pull from collect command to ensure that it works as expected:
- no form-id > pull all forms present
- form-id > pull just that form

This PR makes changes to have the CLI doing the same as the UI (select the forms)

Users should be able to pull and push forms again using the CLI

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/operations/ImportFromODK.java
+++ b/src/org/opendatakit/briefcase/operations/ImportFromODK.java
@@ -22,7 +22,6 @@ import static org.opendatakit.briefcase.operations.Common.STORAGE_DIR;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
-import java.util.List;
 import java.util.Optional;
 import org.opendatakit.briefcase.model.FormStatus;
 import org.opendatakit.briefcase.transfer.TransferForms;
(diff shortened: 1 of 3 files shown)


ggalmazor:
Thanks, @dcbriccetti!

kkrawczyk123:
**Tested with success!!**

PR #730, opened 2019-03-25
Issue 729 cli pull start from date

### LINKED ISSUE #729: New CLI argument to start pull from arbitrary date
Now that we can start pulling a form starting from the last date/submission pulled, we can add a new CLI argument to allow users to define a date from where to start a pull. The suggested argument would be `-sfd` `--start_from_date` and the accepted values would be ISO8601 dates.

ggalmazor:
Closes #729 

Used a form in the sandbox with submissions to test this feature:
- Pull the form with `-sfd` and verify that the expected amount of submissions are pulled.
- Send some submissions and pull the form using the same `-sfd` date. Verify that only the new submissions get pulled.

This PR is a narrow change on top of the existing `-sfl` feature.

This PR should not have other effects other than the new added feature described in #729.

"start from date" will update the last cursor so that the next pull with "start from last" continues pulling from the next submission.

The CLI will ignore `-sfd` when combined with `-sfl`.

Yes! https://github.com/opendatakit/docs/issues/993

### DIFF
--- a/src/org/opendatakit/briefcase/operations/Export.java
+++ b/src/org/opendatakit/briefcase/operations/Export.java
@@ -134,7 +134,8 @@ public class Export {
             briefcaseDir,
             appPreferences.getPullInParallel().orElse(false),
             false,
-            false
+            false,
+            Optional.empty()
         );
       }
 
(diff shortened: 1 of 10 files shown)


kkrawczyk123:
**Tested with success!**

PR #736, opened 2019-05-08
Pull & push ops overhaul

ggalmazor:
# v1.6 QA feedback
## Differences between Aggregate and Central

- Pull features that aren't supported (yet):
  - Start from last
  - Start from date
- Export features that aren't supported (yet):
  - Pull before export
- Briefcase can't push submissions belonging to a form version it doesn't know about

## New issues
- #742

## Pending stuff
- [x] There's some inconsistent toggling of tabs & buttons while a pull/push op is going on
  - Verified that buttons disable while allowing navigation between tabs during pull/push work
  - Verified that buttons re-enable once the pull/push op ends, or the user cancels it.
- [x] There's some inconsistency about when a pull/push op actually ends (thus, toggling UI widgets at incorrect times)
  - Ensured that now a new event is published to signal the end of the operation, which re-enables the UI at correct time.
- [x] "Downloading 0 submissions" could probably be replace by something more meaningful, such as "No submissions to pull"
  - Completely reviewed tracking language
- [ ] Create media folders (even when they will be empty) in Aggregate and Central, so that behavior is more consistent
- [ ] See why only one page of Details buttons get enabled after receiving tracking events
- [x] Fix export conf dialog - disabled pull before export with single form
- [x] Make the export operation refresh the pull cursor when using the "pull before export" option (Aggregate only, UI and CLI)

## Guillermo's To Do list
- [x] Test `-plla` without specifying a form. Use Kasia's Aggregate server. She reports having pulled only 7 from 60 available forms.
- [x] Test `-sfl`. Kasia reports `-sfd` works as expected, so the best guess is that cursors aren't being saved correctly.
- [x] Test `-psha`. Kasia reports:
  - [x] On Aggregate < 2 only forms are pushed
  - [x] On Aggregate >= 2 forms and submissions are correctly pushed but the op doesn't end.
- [x] Test `-e -pb`. Kasia reports that no file is created in the export dir.
- [x] Study why pushing a form takes forever but pushing many forms doesn't
- [ ] Study why when pushing a form to Aggregate, if we delete the form from Aggregate, Briefcase thinks that the form still exists in Aggregate
  - Caching problem?
- [x] Study how to load from the old place to make it easier for users to upgrade to v1.6 without losing saved "pull before export" and resume cursors.
- [ ] Study how we could dynamically configure the payload size when pushing to Aggregate with the information it adds on the OpenRosa headers in all its responses.
- [ ] Study how we could identify when a server is an Aggregate server to limit the number of connections to 2 regardless of what the user has configured in the Settings tab

## Solved stuff
- Now all CLI and UI ops that pull forms use and save the last cursor
- Now all CLI and UI ops wait until all tasks have completed
  - On the UI, we also react differently when a form has been pull/pushed, as opposed to when all forms have been pulled/pushed. This lets us precisely sequence when to toggle UI components.
- Refactored how Briefcase deals with saving Cursors and RemoteServers (Aggregate/Central) into Java prefs to avoid current indirection that the TransferForms and ExportForms create
  - Also took the chance to change pref key naming for consistency and collision avoidance. 
- Errors during a job execution are no longer being swallowed by the executor and show up in the logs
- We show a deprecation notice when adding the `-pp` flag in pull ops
- The new `-mhc` (max HTTP conns) flag has been enabled and can be used with pull ops.

## Added stuff
- Support (dummy implementation) for `pulldata` xform function
- Pull and push processes should perform better because now we download attachments and submissions using parallel streams. This should not interfere on job cancellation.
- When getting the list of forms in a remote server, we filter out those without name or form ID to prevent issues in Briefcase. This won't be necessary once we implement a central database and we can have compound keys and/or other ways to identify forms locally.

## Stuff we have learned
- The "open browser" (server url link) feature doesn't work on Ubuntu 16
- Central doesn't add submission date information to submissions
- Central rejects submissions of forms with versions unknown to it
- It looks like AppEngine won't allow more than 2 simultaneous connections. It's not clear if this restriction is applied for the same client IP or for all sources.
- Running Briefcase with Java 1.9 will fail due to a missing database driver

## Work for the future
- We should review the message that we put in the table, as opposed to the last available message in the Details dialog, which can be unpredictable due to parallel code.
- We should add validation to the Aggregate/Central server forms and toggle buttons according to a validation status.
- We should add support for Central operations in the CLI
- Discuss what to do with pushing forms to archived projects in Central
- Improvements and issues of the job tracking code:
  - The UI can group tracking events into each form's detail window, but the UI gets all messages mixed into the std out and err streams, which makes it super hard to track by users (actually, the logs include the form's name, but doing the same on the tracking event messages would result on super bloated user feedback and redundant info in the UI)
  - We currently have an undetermined progress bar in the UI and the user doesn't have a sense of how much of the job is pending to be done
  - It would be easy to get a manifest of the tasks to be performed by Briefcase before actually doing them and produce get a number of tasks that could be use as bound for the progress bar.
  - The tracker would mark tasks as done as progress get tracked.


dcbriccetti:
Hi! This is marked as a draft. Shall I start looking at it anyway? Will there likely be many more changes?

ggalmazor:
Yes!, just a couple more commits to restore HTTP proxy and add more tests coverage. I was hoping to get into a live review session with you. We can discuss on slack if you want

src/org/opendatakit/briefcase/ui/pull/PullPanel.java
```diff
@@ -57,9 +57,9 @@
   private final Analytics analytics;
   private JobsRunner pullJobRunner;
   private TerminationFuture terminationFuture;
-  private Optional<PullSource<?>> source;
+  private Optional<PullSource> source;
 
-  public PullPanel(TransferPanelForm view, TransferForms forms, BriefcasePreferences tabPreferences, BriefcasePreferences appPreferences, TerminationFuture terminationFuture, Analytics analytics) {
+  public PullPanel(TransferPanelForm<PullSource> view, TransferForms forms, BriefcasePreferences tabPreferences, BriefcasePreferences appPreferences, TerminationFuture terminationFuture, Analytics analytics) {
```
ggalmazor: Now the TransferPanelForm has a type parameter that specifies the kind of panel (source or target) it is. This lets methods returning the source or target be typesafe down the chain.

src/org/opendatakit/briefcase/ui/reused/transfer/TransferPanelForm.java
```diff
@@ -34,15 +34,17 @@
...
+import org.opendatakit.briefcase.ui.reused.transfer.sourcetarget.SourceOrTargetPanel;
+import org.opendatakit.briefcase.ui.reused.transfer.sourcetarget.SourceOrTargetPanelForm;
 
 @SuppressWarnings("checkstyle:MethodName")
-public class TransferPanelForm {
+public class TransferPanelForm<T extends SourceOrTarget> {
   public JPanel container;
-  private SourcePanel sourcePanel;
-  private SourcePanelForm sourcePanelForm;
+  private SourceOrTargetPanel<T> sourceOrTargetPanel;
```
ggalmazor: Using "SourceOrTarget" from now on for lack of a better name

ggalmazor:
I realize some commit messages fall somewhat short to explain some whys and that I'd like to make some small changes. I'll be rehashing the commits during my day so that you can have better material in your day, @dcbriccetti. I'll give you a heads up when it's done.

src/org/opendatakit/briefcase/ui/pull/PullPanel.java
ggalmazor: I've added my previous comment into the commit message

ggalmazor:
Notes from live review with Dave:
- Do this in `JobsRunnerTest.sleep()`
  ```java
        try {
          Thread.sleep(10);
        } catch (InterruptedException e) {
        }
  ```

src/org/opendatakit/briefcase/reused/http/response/ServerError.java
```diff
@@ -16,17 +16,36 @@
...
+  private final String serverErrorResponse;
 
-  ServerError(int statusCode, String statusPhrase) {
+  private ServerError(int statusCode, String statusPhrase, String serverErrorResponse) {
     this.statusCode = statusCode;
     this.statusPhrase = statusPhrase;
+    this.serverErrorResponse = serverErrorResponse;
+  }
+
+  static <U> ServerError<U> from(Request<U> request, HttpResponse response) {
```
dcbriccetti: `request` is unused?

src/org/opendatakit/briefcase/operations/Export.java
```diff
@@ -136,7 +136,7 @@ public static void export(String storageDir, String formid, Path exportDir, Stri
             ).map(Credentials::from)
         );
 
-        pullJob = PullForm.pull(CommonsHttp.reusing(), server, briefcaseDir, false, Export::onEvent, formStatus);
+        pullJob = PullForm.pull(CommonsHttp.of(8), server, briefcaseDir, false, Export::onEvent, formStatus);
```
dcbriccetti: I know the literal 8 is temporary, but it’s explained only in a commit comment (that I can see). Perhaps it should be extracted from where it appears into a constant, whose name expresses its temporary nature, or marked with a // todo. (I don’t feel strongly about this.)

src/org/opendatakit/briefcase/reused/http/response/ServerError.java
ggalmazor: That one is there just for conformity with other factories in sibling subtypes of Response, but we could remove it if you think it's confusing.

src/org/opendatakit/briefcase/operations/Export.java
ggalmazor: I think you're right. I'll move it to a constant in `Launcher` (topmost class)

src/org/opendatakit/briefcase/reused/http/response/ServerError.java
dcbriccetti: Perhaps remove the confusion with a comment, or a warning-disabling statement?

test/java/org/opendatakit/briefcase/reused/job/JobsRunnerTest.java
```diff
@@ -93,7 +93,13 @@ public void can_launch_jobs_synchronously() {
 
   private <T> Function<RunnerStatus, T> returnWhenCancelled(T t) {
     return runnerStatus -> {
-      while (runnerStatus.isStillRunning()) { }
+      while (runnerStatus.isStillRunning()) {
+        try {
+          Thread.sleep(10);
```
dcbriccetti: Glad you made this work. Think of the polar ice caps. :-)

ggalmazor:
This PR feels like it's ready for merging, and then we can start addressing some follow-up work, like replacing the "pull in parallel" setting, and adding more tests.

I'm currently working on JavaRosa. If no one complains, I'll merge it next.

lognaturel:
When I try to pull from Central, I see this lovely dialog that I'm guessing is not exactly as you wanted it:
Screen Shot 2019-06-04 at 10 15 29 AM

ggalmazor:
Thanks for catching that bug, @lognaturel!

It seems that I lost an important configuration about IntelliJ's GUI Designer tool in one of the updates. The default only reflects changes to the .form files to the compiled .class files and we need to set it to change the source files instead.

kkrawczyk123:
@ggalmazor Unfortunately, I was not able to test Central integration as the testing instance currently is not working so I have to wait until it will be fixed with deeper testing. I have some questions/suggestions and I have noticed some issues:

1. How should I understand: "Implement new pull & push ops using the new JobsRunner async&cancellable workflow" in the pr description?

2. pull in parallel removed from ui - what about cmd? I am not able to verify if it works? It should work but with info about old command?

3. How that "maximum http simultaneous connections" works? and what is the purpose of changing it? I do not completely understand how to use it so hard to test.

4. SubmissionId next to the downloaded attachment on details window are added on purpose? Actually, some of the media includes ids some are not. Similar situation with the pushing process
Screenshot from 2019-06-06 10-33-57-3
Screenshot from 2019-06-06 09-44-27-2

5. When instance contains media Pull status used to show a number of downloaded instances, now it is more random? once it shows info about submissions and once about the attachments? superfine to show it in details window but is it that important to show it a Status column?
Screenshot from 2019-06-06 15-02-08-5
Screenshot from 2019-06-06 09-41-06-1

6. I pulled from the aggregate form without submissions, form has been pulled, pulled process finished but the form still has status “Downloading 0 submissions” - but it stopped.
Another similar case: with a chosen checkbox: start pull from last submission pulled. I removed the sd contend and expected 0 downloaded submissions but except that I constantly see "Downloading 0 submissions”, progress bar working, cancel button visible, I have “pulling in progress” impression. 
Screenshot from 2019-06-06 11-22-01-4

7. What if I will put aggregate url in “pull from central” and the opposite. ?
I have put Aggregate url in the central connection window. There is no validation on credentials and project id - I was able to click connect and it was constantly loading, no error messages displayed, the cursor is remaining busy, loading and loading. Probably the same results will be with input Central url and choose Connect without input to other fields. Should Project ID, Username, Password be required fields?

8. What about pushing to central? Am I a dreamer? :)
 
9. Push success messages are also not working like it used to before I can see just pushed submission <submission id> and there is “Successful upload!” displayed on master. 

Cmd testing results:
10. Pulling via cmd -plla is not working. nothing happens, no cmd response, nothing visible in log file. 
Other pull results just for awareness:  
-plla recognized correctly as old one - but no pulling action
-pa proper reaction. 
- cmd pull to not existing sd - folder created form did not pull. 
- bulk pull of forms, id not specified not working as well.
- sfd did not worked too. 
-pp info about old command but still no action. 
- pull from odk dir worked properly
-sfl prefix does not work too.

11. -psha didn’t work too :(

12. -pb, export works but the newest submission is not visible. (probably connected to failing pull process). Regular export cmd process works jeahh

kkrawczyk123:
One more thing that I have missed in the previous comment, so sorry about that let's pretend that it is a point 13 xd. On the pull tab I chose Select all button, then Pull button and click on Cancel. All of the forms have "Canceled by user" status (none successes at all??). When I scroll the pull tab not all forms have enabled details button. I clicked on the disabled details button and details window appeared like on the attached screenshot. In my sd there are all forms but none of them has submissions. The same issue is visible on push tab.
Screenshot from 2019-06-06 15-47-30

ggalmazor:
# v1.6 QA feedback

Guillermo's To Do list:
- Test `-plla` without specifying a form. Use Kasia's Aggregate server. She reports having pulled only 7 from 60 available forms.
- Test `-sfl`. Kasia reports `-sfd` works as expected, so the best guess is that cursors aren't being saved correctly.
- Test `-psha`. Kasia reports:
  - On Aggregate < 2 only forms are pushed
  - On Aggregate >= 2 forms and submissions are correctly pushed but the op doesn't end.
- Test `-e -pb`. Kasia reports that no file is created in the export dir.


Pending stuff:
- There's some inconsistent toggling of tabs & buttons while a pull/push op is going on
- There's some inconsistency about when a pull/push op actually ends (thus, toggling UI widgets at incorrect times)
- "Downloading 0 submissions" could probably be replace by something more meaningful, such as "No submissions to pull"

Solved stuff:
- Now the CLI waits until the whole operation has ended (we block on the executor)
- We show a deprecation notice when adding the `-pp` flag in pull ops
- The new `-mhc` (max HTTP conns) flag has been enabled and can be used with pull ops.

Added stuff:
- Pull and push processes should perform better because now we download attachments and submissions using parallel streams. This should not interfere on job cancellation.
- When getting the list of forms in a remote server, we filter out those without name or form ID to prevent issues in Briefcase. This won't be necessary once we implement a central database and we can have compound keys and/or other ways to identify forms locally.

Stuff we have learned:
- The "open browser" (server url link) feature doesn't work on Ubuntu 16

Work for the future:
- We should review the message that we put in the table, as opposed to the last available message in the Details dialog, which can be unpredictable due to parallel code.
- We should add validation to the Aggregate/Central server forms and toggle buttons according to a validation status.
- We should add support for Central operations in the CLI

PR #746, opened 2019-06-26
Fix non Aggregate compatibility

ggalmazor:
This PR solves issues when dealing with Ona servers:
- Can't connect to list forms in the pull tab
- Can't use the "start from last" feature when pulling forms
- Can't use the "pull before export" feature when exporting forms.

I've tested this manually using my Ona account at https://odk.ona.io/ggalmazor
- Pull forms under various "start from last" scenarios: first time pull, second time pull, send new submissions and pull again, etc.
- Export forms with the "pull before export" option:
  - Single form
  - Multiple form, from Ona and Aggregate servers

Subtyping Cursor felt like the simplest way to go. The alternative would have been to revert the changes to Cursor that made it parse its value, but that would have an impact on how we deal with the "start from date" feature.

The changes to Cursor are required to get the submissions list when pulling forms. This is because Ona uses a simple numeric id, instead of a complex XML document like Aggregate.

An alternative to the changes to support the pull before export feature with Ona would be to simply not support it, but it felt like we ought to make the effort this time.  

Every time we change something related to the java prefs system we should be on the lookout for trouble.

This time, we need to save a new key with the cursor type (aggregate/ona) and when reading cursors from prefs, we base on that value to choose which factory to use (defaulting to aggregate)

If I did something wrong, users might have trouble with the "start from last" feature, and the "pull before export" feature, by extension.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/pull/aggregate/AggregateCursor.java
+++ b/src/org/opendatakit/briefcase/pull/aggregate/AggregateCursor.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright (C) 2019 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.opendatakit.briefcase.pull.aggregate;
+
+import static java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME;
+
+import java.time.LocalDate;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.Objects;
+import java.util.Optional;
+import org.opendatakit.briefcase.export.XmlElement;
+import org.opendatakit.briefcase.model.BriefcasePreferences;
+import org.opendatakit.briefcase.model.FormStatus;
+import org.opendatakit.briefcase.reused.Iso8601Helpers;
+
+/**
+ * This class stores information about a cursor to a list of remote submission
+ * instance IDs, or "resumptionCursor" as described in the <a href="https://docs.opendatakit.org/briefcase-api/#returned-document">Briefcase Aggregate API docs</a>
+ * <p>
+ * The contents of a Cursor are basically a date ({@link #lastUpdate} field) and
+ * a submission instanceID ({@link #lastReturnedValue} field), which are used by
+ * Aggregate to define the lower bound of a page of submission instanceIDs (also
+ * called batch or chunk in the documentation). The upper bound is a defined by
+ * another parameter that's not part of the cursor.
+ * <p>
+ * The date ({@link #lastUpdate} field) in a cursor is related to the last update
+ * date of a submission stored in Aggregate's database, not to be mistaken with
+ * the completion or submission dates, which can be different.
+ * <p>
+ * The submission instanceID of a Cursor ({@link #lastReturnedValue} field) is used
+ * to further filter the contents of a submission instanceID page when the existing
+ * submissions from the provided date don't fit in the same page.
+ */
+public class AggregateCursor implements Cursor<AggregateCursor> {
+  /**
+   * This date is used only to compare Cursors that might have an empty value in lastUpdate
+   */
+  private static final OffsetDateTime SOME_OLD_DATE = OffsetDateTime.parse("2010-01-01T00:00:00.000Z");
+  // TODO v2.0 Use a better name, like xml
+  private final String value;
+  private final Optional<OffsetDateTime> lastUpdate;
+  private final Optional<String> lastReturnedValue;
+
+  private AggregateCursor(String value, Optional<OffsetDateTime> lastUpdate, Optional<String> lastReturnedValue) {
+    this.value = value;
+    this.lastUpdate = lastUpdate;
+    this.lastReturnedValue = lastReturnedValue;
+  }
+
+  public static AggregateCursor empty() {
+    return new AggregateCursor("", Optional.empty(), Optional.empty());
+  }
+
+  /**
+   * Parses the provided cursor xml document and returns a new Cursor instance.
+   */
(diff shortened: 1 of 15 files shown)


kkrawczyk123:
**Tested with success!**
 Verified on Mac, Ubuntu and Windows. Pulling from ona confirmed, pull before export and start pull from the last submission. Confirmed that no regression visible in pulling, pull before and start from last for Aggregate and in pulling from Central.  

src/org/opendatakit/briefcase/pull/aggregate/AggregateCursor.java
```diff
@@ -0,0 +1,168 @@
...
+    );
+    return new AggregateCursor(cursorXml, Optional.of(lastUpdate), Optional.of(lastReturnedValue));
+  }
+
+  /**
+   * Returns a synthetic Cursor instance with the provided values.
+   */
+  public static AggregateCursor of(LocalDate date) {
+    OffsetDateTime lastUpdate = date.atStartOfDay().atOffset(ZoneOffset.UTC);
+    String cursorXml = String.format("<cursor xmlns=\"http://www.opendatakit.org/cursor\">" +
```
dcbriccetti: Could we remove this duplication, please?
Screen Shot 2019-06-30 at 11 02 01 AM
```
  /**
   * Returns a synthetic Cursor instance with the provided values.
   */
  public static AggregateCursor of(OffsetDateTime lastUpdate, String lastReturnedValue) {
    String cursorXml = getCursorXml(lastUpdate, lastReturnedValue);
    return new AggregateCursor(cursorXml, Optional.of(lastUpdate), Optional.of(lastReturnedValue));
  }

  /**
   * Returns a synthetic Cursor instance with the provided values.
   */
  public static AggregateCursor of(LocalDate date) {
    OffsetDateTime lastUpdate = date.atStartOfDay().atOffset(ZoneOffset.UTC);
    String cursorXml = getCursorXml(lastUpdate, "");
    return new AggregateCursor(cursorXml, Optional.of(lastUpdate), Optional.empty());
  }

  private static String getCursorXml(OffsetDateTime lastUpdate, String lastReturnedValue) {
    return String.format("<cursor xmlns=\"http://www.opendatakit.org/cursor\">" +
            "<attributeName>_LAST_UPDATE_DATE</attributeName>" +
            "<attributeValue>%s</attributeValue>" +
            "<uriLastReturnedValue>%s</uriLastReturnedValue>" +
            "<isForwardCursor>true</isForwardCursor>" +
            "</cursor>",
        lastUpdate.format(ISO_OFFSET_DATE_TIME),
        lastReturnedValue);
  }
```

src/org/opendatakit/briefcase/reused/http/CommonsHttp.java
```diff
@@ -78,7 +78,7 @@ private static HttpClientBuilder getBaseBuilder(int maxConnections) {
   @Override
   public <T> Response<T> execute(Request<T> request) {
     // Ensure that we will be using fresh credentials
-    executor.clearAuth();
+    //executor.clearAuth();
```
dcbriccetti: Remove?

src/org/opendatakit/briefcase/reused/http/CommonsHttp.java
```diff
@@ -67,7 +67,7 @@ private static HttpClientBuilder getBaseBuilder(int maxConnections) {
         .create()
         .setMaxConnPerRoute(maxConnections)
         .setMaxConnTotal(maxConnections)
-        .disableAuthCaching()
+        //.disableAuthCaching()
```
dcbriccetti: Remove?

src/org/opendatakit/briefcase/reused/http/RequestBuilder.java
```diff
@@ -220,9 +220,14 @@ public static boolean isUri(String candidate) {
...
-      throw new BriefcaseException("Can't apply withPath() twice");
-    URL newBaseUrl = url(baseUrl + (path.startsWith("/") ? path : "/" + path));
+    String cleanPath = path.startsWith("/") && path.endsWith("/")
+        ? path.substring(1, path.length() - 1)
+        : path.startsWith("/")
+        ? path.substring(1)
+        : path.endsWith("/")
+        ? path.substring(0, path.length() - 1)
+        : path;
+    URL newBaseUrl = url(baseUrl + "/" + cleanPath);
```
dcbriccetti: Is this more readable?
Screen Shot 2019-06-30 at 11 44 38 AM

```
    int start = path.startsWith("/") ? 1 : 0;
    int end   = path.endsWith  ("/") ? 1 : 0;
    String cleanPath = path.substring(start, path.length() - end);
```

dcbriccetti: startOffset and endOffset would be better

src/org/opendatakit/briefcase/pull/aggregate/AggregateCursor.java
ggalmazor: We can't remove duplication exactly as you propose, but we can make it better. It's key we have a `<tag/>`, as opposed to the `<tag></tag>` you're proposing. Don't ask me why but I've noticed different behavior with `<tag/>`.

I'll work in the extraction as you've laid it out, but with that detail on mind.

src/org/opendatakit/briefcase/reused/http/RequestBuilder.java
ggalmazor: Much better, thanks!

ggalmazor:
Thanks, @dcbriccetti! All your comments are super helpful. I've implemented the three changes you have suggested. Could you do a sanity check on them, please?

src/org/opendatakit/briefcase/pull/aggregate/AggregateCursor.java
```diff
@@ -128,6 +113,20 @@ public static AggregateCursor of(LocalDate date, String lastReturnedValue) {
...
+    return String.format("<cursor xmlns=\"http://www.opendatakit.org/cursor\">" +
+            "<attributeName>_LAST_UPDATE_DATE</attributeName>" +
+            "<attributeValue>%s</attributeValue>" +
+            "%s" +
+            "<isForwardCursor>true</isForwardCursor>" +
+            "</cursor>",
+        lastUpdate.format(ISO_OFFSET_DATE_TIME),
+        lastUri
+            .map(value -> "<uriLastReturnedValue>" + value + "</uriLastReturnedValue>")
+            .orElse("<uriLastReturnedValue/>")
```
dcbriccetti: Comment that the `<uriLastReturnedValue/>` must be in that form, to help future readers?

dcbriccetti:
Looks good!

PR #753, opened 2019-06-28
Issue #749 Improve layout of the show source panel to better acommodate info

### LINKED ISSUE #749: Improve the display of project ID in Central pulling
Instead of 
Screen Shot 2019-06-27 at 8 59 17 PM

I think it'd be nicer to have
https://sandbox.central.opendatakit.org (Project ID: 23)

This way we match the dialog that the user saw when entering the project ID.

For bonus points, do this instead...
```
Pulling from: Central server
URL: https://sandbox.central.opendatakit.org 
Project: Project Name (ID: 1)
```
My assumption is that Central's API lets you fetch the project name.

ggalmazor:
Closes #749

Manually tested (Ubuntu) that Aggregate and Central servers show the new labels and rows accordingly.

Aggregate servers should say:
- (Pulling from|Pushing to): Aggregate server
- URL: http://foo.bar

Central servers should say:
- (Pulling from|Pushing to): Central server
- URL: http://foo.bar
- Project ID: 42

This is the narrowest solution I could come up with that doesn't require changing the Swing layout, which is always good.

The change is so small, and affects only to the presentation layer, that I don't think there's any regression risks.

Maybe Windows and Mac struggle with <br/> tags in html labels?

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/reused/transfer/sourcetarget/ShowSourceOrTargetForm.form
+++ b/src/org/opendatakit/briefcase/ui/reused/transfer/sourcetarget/ShowSourceOrTargetForm.form
@@ -22,7 +22,7 @@
           <gridbag weightx="1.0" weighty="0.0"/>
         </constraints>
         <properties>
-          <text value="Pulling data from %s"/>
+          <text value="Action placeholder"/>
         </properties>
       </component>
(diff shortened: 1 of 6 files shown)


kkrawczyk123:
@yanokwa @ggalmazor I've checked on Ubuntu, Mac and Windows - and I like it!

PR #758, opened 2019-07-09
Issue 744 validate aggregate & central dialogs

### LINKED ISSUE #744: Provide more useful error when the URL is wrong
https://forum.opendatakit.org/t/briefcase-unable-to-load-from-configured-source/20486 is the post that inspired this issue. I think maybe we should (in my order of preference)

1. If we see `Aggregate.html` we remove it and everything after on connect. This is my preferred approach even though it means you can't have a URL with Aggregate.html. Seems unlikely that someone would have that.

1. Provide a more useful error. Append `Did you enter the correct URL?` to the `Briefcase wasn't able to load forms using the configured source` message. 

1. Add a hint above the `Username cannot be a Google login` that says, `The URL must not include /Aggregate.html or anything after.` I hate this wording and taking up UI with a rare error.

ggalmazor:
Closes #744, #747

This PR brings new validation capabilities to the Aggregate and Central server configuration dialogs. Validation works this way:
- Valid fields and pristine (untouched) fields get a black label
- Dirty, invalid fields get a red label
- The connect button won't be enabled until the whole form is valid
  - In Aggregate, this means a valid URL and either no credentials or filled in credentials
    - The username won't be valid if it's an email, to prevent using Google accounts.
  - In Central, this means a valid URL, valid project ID (a number), valid email and non-empty password.

Peek 09-07-2019 14-34

Added unit tests of the field validation classes
Manually tested all the possible combinations with the UI

The solution on this PR could feel bloated at first, but it brings basic industry standard form validation stuff from HTML web development.

An alternative would have been to inline everything in the form classes, but that would show a lot of duplicated concepts and would make the UI code much more complex and hard to work with.

Users will be able to avoid formal errors while configuring pull sources and push targets.

If my validation regexps are wrong, valid emails and urls could be mistakenly identifed as invalid. I've used super simple regexp to avoid that, but this is something that could happen. The tradeoff is that invalid values could slip through as well, but there's a second layer of validation when Briefcase actually tries to used the configured information to connect to the servers.

No.


ggalmazor:
Using color-blind safe colors now:

Peek 11-07-2019 09-20

src/org/opendatakit/briefcase/ui/reused/FocusAdapterBuilder.java
```diff
@@ -0,0 +1,36 @@
+package org.opendatakit.briefcase.ui.reused;
+
+import java.awt.event.FocusAdapter;
```
dcbriccetti: The Swing and AWT adapters are great when you don’t want to implement all the interface’s methods, but here you do, so the adapter is an unnecessary layer, n'est-ce pas? Also notice that IDEA can now point out that onFocusGained is never called (for whatever that’s worth—I assume you included it for completeness).

Screen Shot 2019-07-28 at 12 34 21 PM

src/org/opendatakit/briefcase/ui/reused/UI.java
```diff
@@ -172,4 +175,18 @@ public static void removeAllMouseListeners(JComponent component) {
     for (MouseListener listener : component.getMouseListeners())
       component.removeMouseListener(listener);
   }
+
+  public static void flash(JComponent component, Duration duration) {
```
dcbriccetti: This method is uncalled for. (Ha ha, a play on words!)

src/org/opendatakit/briefcase/ui/reused/UiFieldValidator.java
```diff
@@ -0,0 +1,74 @@
...
+import javax.swing.JTextField;
+import org.opendatakit.briefcase.reused.http.RequestBuilder;
+
+public class UiFieldValidator {
+  public static final Predicate<String> REQUIRED = s -> s != null && !s.trim().isEmpty();
+  public static final Predicate<String> NUMBER = s -> s == null || s.matches("\\d+");
+  public static final Predicate<String> EMAIL = s -> s == null || s.matches("\\S+@\\S+");
+  public static final Predicate<String> URI = s -> s == null || RequestBuilder.isUri(s);
+  static final Color COLOR_NOT_VALID = new Color(215, 25, 28); // Color-blind safe red
+  static final Color COLOR_VALID = new Color(44, 123, 182); // Color-blind safe blue
```
dcbriccetti: Very thoughtful with the color-blind-safe colors

src/org/opendatakit/briefcase/ui/reused/transfer/sourcetarget/AggregateServerDialogForm.java
```diff
@@ -77,8 +91,25 @@
...
     getRootPane().setDefaultButton(connectButton);
+  }
 
+  private void updateConnectButton() {
+    connectButton.setEnabled(Stream.of(
+        urlValidator.isValid(),
+        usernameValidator.isValid(),
+        passwordValidator.isValid(),
+        usernameAndPasswordValidator.isValid()
+    ).reduce(true, Boolean::logicalAnd));
```
dcbriccetti: Usually I love your streams, man, but this one increases complexity for no benefit that I can see.
Screen Shot 2019-07-28 at 12 57 19 PM

src/org/opendatakit/briefcase/ui/reused/transfer/sourcetarget/CentralServerDialogForm.java
```diff
@@ -77,15 +93,32 @@
...
+        passwordValidator.isValid()
+    ).reduce(true, Boolean::logicalAnd));
   }
 
   private void triggerConnect() {
     setCursor(Cursor.getPredefinedCursor(Cursor.WAIT_CURSOR));
     try {
       Optional<Credentials> credentials = OptionalProduct.all(
-          Optional.ofNullable(usernameField.getText()).map(String::trim).filter(s -> !s.isEmpty()),
+          Optional.ofNullable(emailField.getText()).map(String::trim).filter(s -> !s.isEmpty()),
```
dcbriccetti: At some point, shall we remove this duplication?

Screen Shot 2019-07-28 at 1 03 46 PM

src/org/opendatakit/briefcase/ui/reused/transfer/sourcetarget/AggregateServerDialogForm.java
dcbriccetti: I’d be all for it if it were as simple as
`Seq(urlValidator, usernameValidator, passwordValidator, usernameAndPasswordValidator).forall(_.isValid)`

src/org/opendatakit/briefcase/ui/reused/KeyAdapterBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+    return this;
+  }
+
+  public KeyAdapterBuilder onKeyReleased(Consumer<KeyEvent> consumer) {
+    keyReleasedConsumer = consumer;
+    return this;
+  }
+
+  public KeyAdapter build() {
+    return new KeyAdapter() {
```
dcbriccetti: Please see my comment about the focus adapter. No need for an adapter here, since you’re implementing everything in the listener interface.

src/org/opendatakit/briefcase/reused/transfer/CentralServer.java
```diff
@@ -65,6 +65,12 @@ private static String buildSessionPayload(Credentials credentials) {
     );
   }
 
+  public static String cleanUrl(String url) {
+    return url.contains("/#/")
+        ? url.substring(0, url.indexOf("/#/"))
+        : url;
```
dcbriccetti: May I suggest this?
```
    int i = url.indexOf("/#/");
    return i == -1 
        ? url 
        : url.substring(0, i);
```
It only searches the string once, and more important, doesn’t duplicate the search argument.

src/org/opendatakit/briefcase/ui/reused/transfer/sourcetarget/CentralServerDialogForm.java
```diff
@@ -77,15 +93,32 @@
 
     connectButton.addActionListener(__ -> triggerConnect());
 
+    urlField.addFocusListener(new FocusAdapterBuilder()
+        .onFocusLost(e -> urlField.setText(CentralServer.cleanUrl(urlField.getText())))
+        .build());
+
+    urlValidator = UiFieldValidator.of(urlField, urlFieldLabel, REQUIRED, URI).onChange(this::updateConnectButton);
```
dcbriccetti: I love how you make Swing easier to use with your functional layers.

src/org/opendatakit/briefcase/ui/reused/UiLinkedFieldValidator.java
```diff
@@ -0,0 +1,47 @@
+package org.opendatakit.briefcase.ui.reused;
+
+import static org.opendatakit.briefcase.ui.reused.UiFieldValidator.COLOR_NOT_VALID;
+import static org.opendatakit.briefcase.ui.reused.UiFieldValidator.COLOR_VALID;
+
+public class UiLinkedFieldValidator {
```
dcbriccetti: A doc comment here might help the reader.

test/java/org/opendatakit/briefcase/ui/reused/UiFieldValidatorTest.java
```diff
@@ -0,0 +1,112 @@
...
+    assertThat(label.getForeground(), is(COLOR_VALID));
+  }
+
+  @Test
+  public void calls_a_callback_each_time_there_is_a_change() {
+    AtomicInteger changeCallbackCalledTimes = new AtomicInteger(0);
+    JTextField field = new JTextField();
+    JLabel label = new JLabel();
+    label.setForeground(BLACK);
+    UiFieldValidator validator = UiFieldValidator.of(field, label, s -> s.equals("foo")).onChange(() -> changeCallbackCalledTimes.getAndIncrement());
```
dcbriccetti: Lambda can be replaced with method reference, IDEA points out.

test/java/org/opendatakit/briefcase/ui/reused/UiFieldValidatorTest.java
```diff
@@ -0,0 +1,112 @@
...
+    assertThat(validator.isValid(), is(true));
+    assertThat(changeCallbackCalledTimes.get(), is(2)); // goes from "invalid" to "valid"
+
+    field.setText("bar");
+    simulateKeyStroke(field);
+    assertThat(validator.isValid(), is(false));
+    assertThat(changeCallbackCalledTimes.get(), is(3)); // goes from "valid" to "invalid"
+  }
+
+  private void simulateKeyStroke(JTextField field) {
```
dcbriccetti: I like this.

test/java/org/opendatakit/briefcase/ui/reused/UiLinkedFieldValidatorTest.java
```diff
@@ -14,8 +14,10 @@
 import org.junit.Test;
 
 public class UiLinkedFieldValidatorTest {
-  private JTextField fieldA, fieldB;
-  private JLabel labelA, labelB;
+  private JTextField fieldA;
```
dcbriccetti: I hereby register mild dislike of this style requirement.

src/org/opendatakit/briefcase/ui/reused/UiFieldValidator.java
```diff
@@ -0,0 +1,74 @@
...
+import javax.swing.JLabel;
+import javax.swing.JTextField;
+import org.opendatakit.briefcase.reused.http.RequestBuilder;
+
+public class UiFieldValidator {
+  public static final Predicate<String> REQUIRED = s -> s != null && !s.trim().isEmpty();
+  public static final Predicate<String> NUMBER = s -> s == null || s.matches("\\d+");
+  public static final Predicate<String> EMAIL = s -> s == null || s.matches("\\S+@\\S+");
+  public static final Predicate<String> URI = s -> s == null || RequestBuilder.isUri(s);
+  static final Color COLOR_NOT_VALID = new Color(215, 25, 28); // Color-blind safe red
```
dcbriccetti: Did you consider INVALID instead of NOT_VALID? (There is the potential that in-'val-id could be confused with 'in-vuh-lid, a person made weak or disabled by illness or injury.)

src/org/opendatakit/briefcase/ui/reused/KeyAdapterBuilder.java
ggalmazor: I've opted to remove the unused methods instead :P

src/org/opendatakit/briefcase/ui/reused/UiLinkedFieldValidator.java
ggalmazor: I was hoping that the tests would shed some light. I'll reference them in a doc block

src/org/opendatakit/briefcase/ui/reused/UiFieldValidator.java
ggalmazor: Yes, that's precisely why I went for NOT_VALID. In this case, we're using colors adapted for people with different eyesight and it feels even more important to avoid the "invalid" term.

ggalmazor: I feel like this might be a bigger deal in Spanish than in English. Do you think it's OK to use "invalid"?

src/org/opendatakit/briefcase/ui/reused/UiFieldValidator.java
ggalmazor: It was Yaw's idea :)

src/org/opendatakit/briefcase/ui/reused/UiFieldValidator.java
dcbriccetti: I think, here in the USA at least, the risk of misunderstanding and upset is very small, probably not justifying avoiding the natural antonym. I’m cool with it either way.

kkrawczyk123:
@ggalmazor I have some questions/thoughts/suggestions:

1. On Pull/Push tab: I Input url, input username and password, then remove username and password - connect button is enabled for aggregate (it probably shouldn't be) even if anonymous access is not allowed (the button is disabled for Central, the issue is visible only for aggregate). Here comes a question what if anonymous access allowed and it is expected that user doesn't want to input credentials? 

2. On Pull/Push tab: I Input username and password,  wanted to correct the password so remove it - both username and password are red. When a password is empty - both username and password are red. But ex. when I use an email as username on Aggregate Server Configuration - only username is red, a password is blue. So I am not sure if validation is for empty fields or correct credentials too?

3. On Pull/Push tab: I input Aggregate url on Central window - Error says that Aggregate url nor found. It should probably say that Central url not found. 
Screenshot from 2019-08-02 09-23-35-1

4. On Pull/Push tab: for Central: I input non-existing project id and it is blue - shouldn’t it be red? - I am able to click on Connect and log in but then “Briefcase wasn’t able to reload forms” maybe some other error message should be displayed?. The empty list is displayed and Pulling from so it suggests that you can pull from that project. On Push tab I am able to log in to a non-existing project, list of form is displayed and I am even able to Push a form there and I have “Success with errors” status. 
Screenshot from 2019-08-02 09-27-13-4

5. On Pull/Push tab: On Windows and Ubuntu some space between email and Password lines would look a bit better than right now. (I know it can be harder than it sounds) 
Screenshot from 2019-07-17 13-58-29-2

6. On Pull tab: I put Central url https://test.central.opendatakit.org/#/projects/18 to Aggregate Configuration Window and super ugly error message surprised me, url was not cleaned. 
Screenshot from 2019-08-02 11-31-33-6
When I've tried with  https://test.central.opendatakit.org/ I saw: Error message aggregate not found, what is ok. 

7. On Push tab: I put Central url on Aggregate Conf window  - url is not cleaned I was able to log in :O https://test.central.opendatakit.org/#/projects/18, on url without a project I saw error: Error message aggregate not found, what is ok. 
Screenshot from 2019-08-02 11-08-50-8

8. On Pull/Push tab: I put local aggregate url in Central Server Configuration window and it wasn’t cleared at all - you can have local central instance right? So it probably should be cleared 
Screenshot from 2019-07-17 13-49-03-3
 After using the Connect button, an error message “Wrong Credentials” is visible, but it's more like the wrong url. 

9. Pull tab: Permissions of my user on which I am logging in are “Data Collector” only and I am able to log in - should I be able to do it? As dialog message says I should have “Data viewer”. Anonymous access is not allowed and I am clearing prefs between each try - as credentials cache issue is still available. (Yes I am saving permissions changes on aggregate each time and preloading page to check it they are saved correctly.
Push tab: (requires form manager permissions): For “data collector” permissions only and the same credentials and url I have: “Please review the connection parameters and try again.”.

10. Previously when HTTP protocol was not added to URL there was an error message suggesting that protocol is missing - it has been removed right? Url is highlighted with red when there is protocol missing. Just confirm :)

kkrawczyk123:
@opendatakit-bot unlabel "needs testing'

ggalmazor:
Thanks, @kkrawczyk123! I won't be able to go through the feedback until late Monday, probably.

kkrawczyk123:

ggalmazor:
OK, so much feedback! Thanks!

> On Pull/Push tab: I Input url, input username and password, then remove username and password - connect button is enabled for aggregate (it probably shouldn't be) even if anonymous access is not allowed (the button is disabled for Central, the issue is visible only for aggregate). Here comes a question what if anonymous access allowed and it is expected that user doesn't want to input credentials?

Users are required to provide username/password because Central's API requires an auth token, which requires credentials. I'm not sure how anonymous access would work with Central, though. I'll ask if there's a way to make it work without credentials, though.

ggalmazor:
> 2.On Pull/Push tab: I Input username and password, wanted to correct the password so remove it - both username and password are red. 

This is because username and password fields are linked and one can never be valid if the other is empty

> When a password is empty - both username and password are red. But ex. when I use an email as username on Aggregate Server Configuration - only username is red, a password is blue

OK, the password is blue because you have provided one, so that field is valid. The username is red because we can't let you use emails as credentials in Aggregate since v1.17 (I think), so the field is invalid.

> So I am not sure if validation is for empty fields or correct credentials too?

Validation is for non-empty fields, unless they're a couple of linked fields, with have extra validations to make sure you fill both in.

ggalmazor:
> 3. On Pull/Push tab: I input Aggregate url on Central window - Error says that Aggregate url nor found. It should probably say that Central url not found.

Fix added. Thanks!

ggalmazor:
>4. On Pull/Push tab: for Central: I input non-existing project id and it is blue - shouldn’t it be red? - I am able to click on Connect and log in but then “Briefcase wasn’t able to reload forms” maybe some other error message should be displayed?

So, support for Central is very basic so far. I agree that we should verify that the project exists, etc., but I think that we will have to deal with that in the future. There are other aspects about the interaction with Central that we already know they will change, so we can address these things later.

We need to make sure we open issues with this to not forget about it, though.

ggalmazor:
> 5. On Pull/Push tab: On Windows and Ubuntu some space between email and Password lines would look a bit better than right now. (I know it can be harder than it sounds)

Ugh, layout in Swing is so hard! @dcbriccetti, do you see some way we could fix that without adding a lot of if blocks here and there? (see the screenshot at point 5 of @kkrawczyk123's comment above)

ggalmazor:
> 6. On Pull tab: I put Central url https://test.central.opendatakit.org/#/projects/18 to Aggregate 
Configuration Window and super ugly error message surprised me

Fixed it!

> url was not cleaned.

I'm more inclined to leave it there, so that the user can check it and fix what's wrong.

ggalmazor:
> 7. On Push tab: I put Central url on Aggregate Conf window - url is not cleaned I was able to log in :O https://test.central.opendatakit.org/#/projects/18, on url without a project I saw error: Error message aggregate not found, what is ok.

I think I don't need to do anything about this one, right?

ggalmazor:
> 8.On Pull/Push tab: I put local aggregate url in Central Server Configuration window and it wasn’t cleared at all - you can have local central instance right? So it probably should be cleared

It doesn't work like that. The field will clean URLs belonging to the kind of server you've selected. If you're configuring a Central server, it will clean Central URLs (not Aggregate URLs) and viceversa.

ggalmazor:
> 9. Pull tab: Permissions of my user on which I am logging in are “Data Collector” only and I am able to log in - should I be able to do it? As dialog message says I should have “Data viewer”. Anonymous access is not allowed and I am clearing prefs between each try - as credentials cache issue is still available. (Yes I am saving permissions changes on aggregate each time and preloading page to check it they are saved correctly.

I'll review this at detail later.

ggalmazor:
> 10. Previously when HTTP protocol was not added to URL there was an error message suggesting that protocol is missing - it has been removed right? Url is highlighted with red when there is protocol missing. Just confirm :)

Confirmed!

kkrawczyk123:
@ggalmazor, I have tested again after adding fixes: 

1. Confirmed that fixed. Ubuntu, MacOS and Windows

2. That makes sense.

3. Confirmed that fixed. Ubuntu, MacOS and Windows

4. Reported an issue in https://github.com/opendatakit/briefcase/issues/786

5. I knew it might have been suuper suuper difficult to do. 

6. Confirmed that fixed. Ubuntu, MacOS and Windows

7. Why not doing it? I don’t think that it should be allowed, it is misleading and gives a wrong impression.

8. But if it is a url of local server how briefcase differentiate if it’s url of aggregate or central? Is briefcase that clever and I am not aware? XD

9. Confirmed that fixed. Ubuntu, MacOS and Windows

10. Ok, good.

And I have a new one question, does the URL clear feature should work for ona server too?

ggalmazor:
7. I just think it doesn't make sense to do anything at all when a user enters a Central URL in the Aggregate dialog or vice versa.

8. Briefcase detects the `.../Aggregate.html...` part in the URL and cut what's in front of that. It does something similar with Central URLs.

> does the URL clear feature should work for ona server too?

No :P Just Aggregate URLs :)

kkrawczyk123:
7. But it is not allowed on Pull and allowed on Push, don't you think it that behavior should be compliant? 

8. That makes sense.

ggalmazor:
> 7. But it is not allowed on Pull and allowed on Push, don't you think it that behavior should be compliant?

I think we're mixing things because it's confusing:
- This PR provides input assistance and **format** validation in dialogs used in the pull and push tabs.
- The behavior of these dialogs is exactly the same regardless of you pushing or pulling forms.
- Before this PR, Briefcase already did basic testing of data introduced by the user. These tests are not comprehensive and we favor false positives over false negatives.
- This means that users could enter data that Briefcase thinks it's OK, when it's not.
- When this happens, the following requests will probably fail, and Briefcase will show some kind of error message, or show empty lists, which is fine so far.
- This PR doesn't improve any of these tests, so false positives are expected to continue happening.
  - Any improvement on this front should be discussed in a new issue, with a specific focus on cost-effectiveness.

The difference you're seeing between the pull and push tabs comes from the testing of the data you're entering (valid URLs to wrong servers, valid non-existing project ids, etc.)

PR #760, opened 2019-07-10
Issue 759 fix parsing times with unsupported offsets

### LINKED ISSUE #759: Error parsing time strings with unsupported offset format
#### Software versions 
Briefcase v1.16.0

#### Problem description
Reported by @aurdipas [on the forums](https://forum.opendatakit.org/t/odk-briefcase-v1-16-beta/20600/21?u=ggalmazor)

Briefcase fails to export submissions with time fields encoded as `HH:mm:ss.sss±HH`. Briefcase expects time zone offsets to be expressed as `±HH:mm`.

- Unzip this form and submissions to Briefcase's storage directory under `forms`: [Test form.zip](https://github.com/opendatakit/briefcase/files/3385505/Test.form.zip)
- Reload forms in the Settings tab
- Export the "Test form" form

You'll see that 100% of submission are exported, but after inspecting the output file, you won't find the submission. You'll find an exception stacktrace in the logs as well.

#### Expected behavior
It should complete the export without failing, and the output file should have a row with the submission's data.

#### Other information 
N/A

ggalmazor:
Closes #759

This PR has a stowaway change: removing the `WhitespaceAround` rule from the checkstyle config file:
- This rule is super noisy because every time that someone makes a change to a UI form using IntelliJ's GUI Designer, it generates code in their companion classes that offends the rule at least once. If the form has customized fonts it's even worse because there are multiple offending lines that you need to search and change *every time*.
- This rule can be annoying when dealing with no-op functions as `() -> {}`, forcing to write them with an empty space between the brackets, or simple enums such as `enum Type {A, B}`, forcing you to sandwich the enum values between whitespaces.
- I don't think this particular rule helps us to write particularly easier to read code, so I've removed it to spare us from wasting more time on it.

- Added automated tests that verify that time values that would fail now work
- Manually exported the form attached to the issue using the UI

No other approaches were considered because there's already another precedent of the same normalization with datetime values in submissions.

It feels a pretty safe change. I can't foresee any regression risk on this one.

Nope.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -70,16 +70,6 @@
             <property name="option" value="alone"/>
             <property name="tokens" value="CLASS_DEF, METHOD_DEF, CTOR_DEF, LITERAL_FOR, LITERAL_WHILE, STATIC_INIT, INSTANCE_INIT"/>
         </module>
-        <module name="WhitespaceAround">
-            <property name="allowEmptyConstructors" value="true"/>
-            <property name="allowEmptyMethods" value="true"/>
-            <property name="allowEmptyTypes" value="true"/>
-            <property name="allowEmptyLoops" value="true"/>
-            <message key="ws.notFollowed"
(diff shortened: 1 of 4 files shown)


kkrawczyk123:
**Testes with success!!**

PR #762, opened 2019-07-10
Issue #761 Encoding issues when pulling submissions with Arabic characters in Windows

### LINKED ISSUE #761: Encoding issues when pulling submissions with Arabic characters in Windows
#### Software versions 
Briefcase v1.16.0, Windows

#### Problem description
Problem reported by [forum user](https://forum.opendatakit.org/t/odk-briefcase-v1-16-beta/20600/9?u=ggalmazor).

- Unzip [kii_c5.zip](https://github.com/opendatakit/briefcase/files/3377170/kii_c5.zip) in the `forms` folder of your Briefcase storage directory, and reload forms
- Push the form to the [sandbox](https://sandbox.aggregate.opendatakit.org)
- Remove the form from the storage directory and reload forms
- Pull the form from the sandbox
- Export the form

You should see that the Arabic characters are incorrectly encoded in the export files.

#### Expected behavior
Arabic characters should be exported correctly.

#### Other information 
The source of the issue comes from the way Briefcase reads the submissions when pulling them. On macOS and Linux machines, where UTF-8 encodings are commonplace, this issue isn't observed.

ggalmazor:
Closes #761

Manually tested on a Windows 10 machine using the form attached to the issue
Manually tested on Linux to ensure that it still worked as expected on another system.

This is a narrow fix and I didn't consider any other approach.

It feels like it's a safe change with no risk of affecting in an unpredictable manner.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/reused/http/RequestBuilder.java
+++ b/src/org/opendatakit/briefcase/reused/http/RequestBuilder.java
@@ -32,7 +32,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
 import java.io.UncheckedIOException;
 import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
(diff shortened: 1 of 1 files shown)


kkrawczyk123:
**Tested with success!!**

PR #764, opened 2019-07-10
Issue 763 central cli ops

### LINKED ISSUE #763: Add Central pull and push CLI ops
Now that we have support for working with Central using the UI, we should add new CLI ops to do the same using the CLI.

ggalmazor:
Closes #763

This PR adds new pull from Central and push to Central CLI operations.
This PR deprecates `-url --aggregate_url` (see below)

Run both CLI operations using the `all-widgets` form and the Central sandbox at https://sandbox.central.opendatakit.org

This PR replicates the same things we already are doing for Aggregate server.

Instead of having shared args for things like the server's url and credentials, I considered having a separate set of args, but it felt bloated and not very reasonable overall.

To be able to share args for the server's url and credentials, I had to review the language on their shortcode, longcode, and descriptions, and I found an irreconcilable issue with the `-url --aggregate_url` argument. I had to either change the longcode to `--server_url` or deprecate it and create a new arg. I opted for the latter, which means that users won't be able to continue using `-url` and will be prompted to use the new `-U` CLI arg.

Yes! https://github.com/opendatakit/docs/issues/1069

### DIFF
--- a/src/org/opendatakit/briefcase/operations/PullFormFromCentral.java
+++ b/src/org/opendatakit/briefcase/operations/PullFormFromCentral.java
@@ -23,7 +23,6 @@ import static org.opendatakit.briefcase.reused.http.Http.DEFAULT_HTTP_CONNECTION
 
 import java.net.URL;
 import java.nio.file.Path;
-import java.time.LocalDate;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
@@ -31,8 +30,7 @@ import org.opendatakit.briefcase.model.BriefcasePreferences;
 import org.opendatakit.briefcase.model.FormStatus;
 import org.opendatakit.briefcase.model.FormStatusEvent;
 import org.opendatakit.briefcase.model.RemoteFormDefinition;
-import org.opendatakit.briefcase.pull.aggregate.Cursor;
-import org.opendatakit.briefcase.pull.aggregate.PullFromAggregate;
+import org.opendatakit.briefcase.pull.central.PullFromCentral;
 import org.opendatakit.briefcase.reused.BriefcaseException;
 import org.opendatakit.briefcase.reused.Optionals;
 import org.opendatakit.briefcase.reused.http.CommonsHttp;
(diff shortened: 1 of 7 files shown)


src/org/opendatakit/briefcase/operations/PullFormFromCentral.java
```diff
@@ -0,0 +1,133 @@
...
+import org.opendatakit.briefcase.reused.transfer.CentralServer;
+import org.opendatakit.briefcase.transfer.TransferForms;
+import org.opendatakit.briefcase.util.FormCache;
+import org.opendatakit.common.cli.Args;
+import org.opendatakit.common.cli.Operation;
+import org.opendatakit.common.cli.Param;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PullFormFromCentral {
```
dcbriccetti: Is it still conventional to use nouns for classes? If so, how about CentralFormPuller?

src/org/opendatakit/briefcase/operations/PullFormFromCentral.java
```diff
@@ -0,0 +1,133 @@
...
+    Response<List<RemoteFormDefinition>> response = http.execute(server.getFormsListRequest(token));
+    if (!response.isSuccess()) {
+      System.err.println(response.isRedirection()
+          ? "Error connecting to Central: Redirection detected"
+          : response.isUnauthorized()
+          ? "Error connecting to Central: Wrong credentials"
+          : response.isNotFound()
+          ? "Error connecting to Central: Central not found"
+          : "Error connecting to Central");
+      return;
```
dcbriccetti: I’m starting to like this form of if/else if/…/else if/else. I prefer that the question mark abut the question, as in English, but I accept your style.

src/org/opendatakit/briefcase/operations/PullFormFromCentral.java
ggalmazor: `CentralFormPuller` wouldn't be correct because this class statically describes a pull operation but it doesn't actually perform it (that's the `JobsRunner`'s work).

I consider this class to be part of a group of domain actions. The complete name would be `PullFormFromCentralAction` but it's longer and feels redundant.

kkrawczyk123:
**Tested with success!!**
I have verified pllc with id, pllc without id, pllc with mhc and pshc I was also looking for regression in aggregate commands and I haven't found any !!

PR #771, opened 2019-08-01
Issue 743 update notifications

### LINKED ISSUE #743: Add update notifications to footer for when new versions are released
This is hopefully a minor and non-controversial request so I'm putting it here as an issue instead of a forum post.

We should let people know when there are new versions of Briefcase. My preferred approach here would be to do what we do in Aggregate with a non-blocking call to GitHub. In the lower left-hand side of every tab, we put the following:
* Current version: vx.y.z
* Old version: vx.y.z - [Update available](https://github.com/opendatakit/briefcase/releases/latest)

I'd love to see this not in the upcoming release, but in the following one.

ggalmazor:
Closes #743

- Added automated tests.
- Tested using the UI
  - To test the available update link, I artificially changed the condition `if (versionManager.isUpToDate()) {` at line 162 to `if (!versionManager.isUpToDate()) {`

This is the smallest possible thing we can do. More advanced approaches would involve parsing the semver to take into account major, minor, and patch numbers.

Users might be confused when they see that the version is no longer available in the title.

Other than that, it feels pretty safe because the version check is performed in the background.

Nope.


test/java/org/opendatakit/briefcase/util/BriefcaseVersionManagerTest.java
```diff
@@ -0,0 +1,60 @@
...
+  public void knows_the_url_to_get_the_latest_version() {
+    assertThat(getLatestUrl(), is(url("https://github.com/opendatakit/briefcase/releases/latest")));
+  }
+
+  @Test
+  public void knows_to_get_the_latest_available_version() {
+    FakeHttp http = new FakeHttp();
+
+    http.stub(
+        get(url("https://api.github.com/repos/opendatakit/briefcase/releases/latest")).build(),
```
dcbriccetti: This string appears 4 times in the source. Maybe it could be a constant in `BriefcaseVersionManager`?

test/java/org/opendatakit/briefcase/util/BriefcaseVersionManagerTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.opendatakit.briefcase.util.BriefcaseVersionManager.getLatestUrl;
+
+import org.junit.Test;
+import org.opendatakit.briefcase.reused.http.FakeHttp;
+import org.opendatakit.briefcase.reused.http.response.ResponseHelpers;
+
+public class BriefcaseVersionManagerTest {
+  @Test
+  public void knows_the_url_to_get_the_latest_version() {
+    assertThat(getLatestUrl(), is(url("https://github.com/opendatakit/briefcase/releases/latest")));
```
dcbriccetti: Maybe this could be a constant in `BriefcaseVersionManager`?

src/org/opendatakit/briefcase/util/BriefcaseVersionManager.java
```diff
@@ -0,0 +1,37 @@
...
+import org.opendatakit.briefcase.reused.http.Http;
+import org.opendatakit.briefcase.reused.http.RequestBuilder;
+
+public class BriefcaseVersionManager {
+  private final Http http;
+  private final String current;
+
+  public BriefcaseVersionManager(Http http, String current) {
+    this.http = http;
+    this.current = current.contains("-") ? current.substring(0, current.indexOf("-")) : current;
```
dcbriccetti: Hey, this expression with a conditional operator reads just like English. Is my style influencing you, or is this a slip-up? ;-)

src/org/opendatakit/briefcase/ui/settings/SettingsPanelForm.form
```diff
@@ -384,6 +384,23 @@
...
+      </vspacer>
+      <component id="53404" class="javax.swing.JLabel" binding="versionLabel">
+        <constraints>
+          <grid row="19" column="1" row-span="1" col-span="5" vsize-policy="0" hsize-policy="0" anchor="0" fill="0" indent="0" use-parent-layout="false"/>
+          <gridbag weightx="1.0" weighty="0.0"/>
+        </constraints>
+        <properties>
+          <enabled value="false"/>
+          <font style="0"/>
+          <text value="[version placeholder]"/>
```
dcbriccetti: Could this be an empty string?

test/java/org/opendatakit/briefcase/util/BriefcaseVersionManagerTest.java
ggalmazor: I'm going to leave this duplication. It's not a big deal, since it's just a getter but:
- Using a constant in the implementation would feel like cheating and make the test quite irrelevant (basically, testing a getter).
- This test just makes the url we're using explicit and serves as documentation of the tested class. One day, when GH changes the url, we can come back and know what has to be changes without investing a lot of time on it.

src/org/opendatakit/briefcase/util/BriefcaseVersionManager.java
ggalmazor: Yes! All your help and reviews are super valuable :)

src/org/opendatakit/briefcase/ui/settings/SettingsPanelForm.form
ggalmazor: I guess it could, but a placeholder is kinda less confusing? maybe?

kkrawczyk123:
**Tested with success!!**
The hyperlink is not working on Ubuntu 16 just like with servers URL but we are aware. I confirmed that it is clickable on Ubuntu 18. 
I have also noticed that "Update available" is displayed in two separate lines after choosing storage directory on MacOS only so I will report it as a separate issue. 

src/org/opendatakit/briefcase/util/BriefcaseVersionManager.java
dcbriccetti: I believe the second sentence, but not the first. :-) You would never deliberately switch away from your
```
? something
: something
```

src/org/opendatakit/briefcase/ui/settings/SettingsPanelForm.form
dcbriccetti: Whatever you prefer.

src/org/opendatakit/briefcase/util/BriefcaseVersionManager.java
ggalmazor: Hahaha I normally break lines only if they get too long

PR #776, opened 2019-08-07
Fix #735 - Provide message when pulling non-existent form from odk dir using cmd

### LINKED ISSUE #735: No information when trying to pull not existing form from odk dir using cmd
#### Software versions 
Briefcase v1..x, Java v1.x.x, operating system, Aggregate v1.x.x, Collect v1.x.x...

#### Problem description
There is no information that form does not exist when trying to pull not existing form from odk dir using cmd

1. Use cmd and pull from odk dir command, as id use a random word so that form won't exist in your sd
2. Check the cmd output

#### Expected behavior
In the cmd output, there should be the information that form does not exist in the odk folder. Now it gives you the impression that pull actually happened. 

#### Other information 
When pulling not existing form from Aggregate the information does not exist is visible.

mayank8318:
Closes #735 

Manually tested it as shown in the screenshot:
Screenshot from 2019-08-07 21-05-06

This approach is similar to the one used in `PullFormFromAggregate` class.
https://github.com/opendatakit/briefcase/blob/master/src/org/opendatakit/briefcase/operations/PullFormFromAggregate.java#L118

This change will inform the user in case the form id they entered is invalid. Before this, briefcase gave no error message and the user was under the impression that the desired pull operation was successful.

No

### DIFF
--- a/src/org/opendatakit/briefcase/operations/ImportFromODK.java
+++ b/src/org/opendatakit/briefcase/operations/ImportFromODK.java
@@ -24,6 +24,7 @@ import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Optional;
 import org.opendatakit.briefcase.model.FormStatus;
+import org.opendatakit.briefcase.reused.BriefcaseException;
 import org.opendatakit.briefcase.transfer.TransferForms;
 import org.opendatakit.briefcase.util.FileSystemUtils;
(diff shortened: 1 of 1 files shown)


kkrawczyk123:
**Tested with success!**
Message in the cmd output:
Screenshot from 2019-08-09 14-49-38
Thanks @mayank8318!

PR #778, opened 2019-08-08
Issue 777 Latest Central rejects multiple simultaneous auth methods

### LINKED ISSUE #777: Latest Central rejects multiple simultaneous auth methods
#### Software versions 
Briefcase v1.16

#### Problem description
Latest (bleeding edge) Central will reject request coming with both `Authorization: Bearer {{token}}` and a `__Host-session` cookie.

This results in users not being able to pull forms (although listing the available forms works)

Connect to Central, pull a form.

#### Expected behavior
Pulling forms should work

#### Other information 
We need to decide whether to stick to sending the auth token in all requests or start using the cookie after a successful login.

A quick fix would be to set the `RequestConfig`'s cookie spec to `IGNORE_COOKIES`, which would be safe to do across all supported pull sources and push targets. 

(see `org.opendatakit.briefcase.reused.http.CommonsHttp`)

ggalmazor:
Closes #777

Manually test against bleeding edge Central server at antinod.es

Central (latest) will reject requests arriving with auth tokens and session cookies. Since we don't really need cookies for anything, this is a quick fix for this issue while we decide what to do about auth in Central

There's the risk of undocumented dependency on cookies by aggregate-compatible servers that we don't know about. If any of these servers rely upon cookies that were being sent, pull/push operations could fail due to this change. Testing in known servers (Ona, Kobo...) would be required to ensure we're breaking anything.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/reused/http/CommonsHttp.java
+++ b/src/org/opendatakit/briefcase/reused/http/CommonsHttp.java
@@ -16,7 +16,7 @@
 
 package org.opendatakit.briefcase.reused.http;
 
-import static org.apache.http.client.config.CookieSpecs.STANDARD;
+import static org.apache.http.client.config.CookieSpecs.IGNORE_COOKIES;
 import static org.apache.http.client.config.RequestConfig.custom;
 import static org.opendatakit.briefcase.reused.http.RequestMethod.POST;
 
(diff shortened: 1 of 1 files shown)


kkrawczyk123:
**Tested with success!**

ggalmazor:
Hi, @dcbriccetti! Will you have some cycles this week to do this one? We would like to release v1.17 next week if possible ;)

PR #787, opened 2019-08-15
Fix #786 - Show error message when invalid project id is used for Central

### LINKED ISSUE #786: Add error message when log in to non-existing project on Central on Pull and Push tab
#### Software versions 
Briefcase v1.16.x, Java v1.11.x, operating system, Aggregate v1.x.x, Collect v1.x.x...

#### Problem description
When input Id of non-existing Project on Central Connection Window and click on Connect button and log in but then “Briefcase wasn’t able to reload forms” I think that maybe some other error message should be displayed? The empty list is displayed and Pulling from so it suggests that you can Reload and pull from that project.  
On Push tab I am able to log in to the non-existing project, list of form is displayed and I am even able to Push a form there and I have “Success with errors” status which is totally misleading.

Try to log in to Central using Project Id that you are sure that does not exist on Central server. 

#### Expected behavior
I think that "Central not found" error could be enough or maybe more specific @ggalmazor what do you think?

mayank8318:
Closes #786 

Manually tested with correct and incorrect project IDs as shown below:
https://www.youtube.com/watch?v=Rd0Vuyg1PMg

I have tried to keep it as consistent with the current design as possible. 

This change will warn the user of an invalid project ID for the Central server at the time of setting up the connection. No risks to the users.

No

### DIFF
--- a/src/org/opendatakit/briefcase/reused/transfer/CentralServer.java
+++ b/src/org/opendatakit/briefcase/reused/transfer/CentralServer.java
@@ -28,10 +28,8 @@ import java.util.Optional;
 import org.opendatakit.briefcase.model.BriefcasePreferences;
 import org.opendatakit.briefcase.model.FormStatus;
 import org.opendatakit.briefcase.model.RemoteFormDefinition;
-import org.opendatakit.briefcase.reused.BriefcaseException;
 import org.opendatakit.briefcase.reused.OptionalProduct;
 import org.opendatakit.briefcase.reused.http.Credentials;
-import org.opendatakit.briefcase.reused.http.Http;
 import org.opendatakit.briefcase.reused.http.Request;
 import org.opendatakit.briefcase.reused.http.RequestBuilder;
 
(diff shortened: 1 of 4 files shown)


mayank8318:
@ggalmazor As of now, Central responds with 404.1 for status not found.
Screenshot from 2019-08-19 15-15-24

Also, I tried with random wrong server URL(valid tho) and noticed that the same message is shown as the status code is the same.
Screenshot from 2019-08-19 15-29-49

mayank8318:
If it is an invalid server then it essentially means that the test to get the token has to fail with a 404 and not a 401 (as for invalid credentials). We can leverage this fact in the new test in `PullSource` and `PushTarget` and return a special two-layer Response object back to `CentralServerDialog`?

ggalmazor:
> If it is an invalid server then it essentially means that the test to get the token has to fail with a 404 and not a 401 (as for invalid credentials). We can leverage this fact in the new test in PullSource and PushTarget and return a special two-layer Response object back to CentralServerDialog?

That's a good idea, but I think it would complicate stuff for Aggregate. I don't think it's worth the hassle.

Now that you've confirmed this, I think our best option would be to change that error message from "Project ID does not exist" to "Central server or project not found". What do you think?

mayank8318:
@ggalmazor Yes, that makes sense. Updating the PR now

ggalmazor:
Hi @mayank8318! Could you rebase master and deal with conflicts?

mayank8318:
@ggalmazor Done

kkrawczyk123:
**Testes with success!!**

PR #727, opened 2019-03-23
721 pull all forms

kunalmulwani:
Closes #721 

#### What has been done to verify that this works as intended?
I checked the forms displayed in UI and compared with what was downloaded.
Screenshot from 2019-03-23 01-34-44
Screenshot from 2019-03-23 01-29-08
Screenshot from 2019-03-23 01-29-23
Screenshot from 2019-03-23 01-29-31

#### Why is this the best possible solution? Were any other approaches considered?
Made the form_id argument optional.Similar to ```ImportFromODK```

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
If you run the command ```java -jar ODK-Briefcase-v1.14.0.jar -plla -sd /Users/yanokwa/Desktop -u admin -p admin -url https://example.com``` it should download all the forms from aggregate server. If you specify ```form_id``` argument then it just downloads the specified form.

PR #788, opened 2019-08-18
775: Push all forms to Aggregate if a form ID is not provided

### LINKED ISSUE #775: Push all forms when form_id is not specified
**Problem description**
Similar to https://github.com/opendatakit/briefcase/issues/721, there isn't an easy way to push all forms from the storage directory into a server.

**Expected behavior**
Running a command like (and the Aggregate equivalent) will push all forms.

`java -jar briefcase.jar --push_central --storage_directory /home/ --odk_url test.central.com --project_id 1 --odk_email me@email.com --odk_password helloworld`

acj:
Closes #775

I added a bit of temporary logging code to see which forms were being pushed, and then manually ran a command like this:

```
$ java -jar build/libs/ODK-Briefcase-v1.16.0-1-g91f5f85f-dirty.jar --push_aggregate --aggregate_url https://sandbox.aggregate.opendatakit.org/ --odk_username foo --odk_password bar --storage_directory /tmp
```

All local forms were selected and pushed.

Straightforward change. This is substantially similar to the work done in #727.

The only user-facing change is that the form ID can be omitted, and this will be interpreted as selecting all forms.

I don't think so.

#### Notes

This seems to work for Aggregate, but the issue description mentions pushing to Central. Are parameters like `push_central` being added in 1.17? Is there anything else to be done right now?

### DIFF
--- a/src/org/opendatakit/briefcase/operations/PushFormToAggregate.java
+++ b/src/org/opendatakit/briefcase/operations/PushFormToAggregate.java
@@ -25,7 +25,9 @@ import static org.opendatakit.briefcase.reused.http.Http.DEFAULT_HTTP_CONNECTION
 
 import java.nio.file.Path;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import org.opendatakit.briefcase.model.BriefcasePreferences;
 import org.opendatakit.briefcase.model.FormStatus;
 import org.opendatakit.briefcase.model.FormStatusEvent;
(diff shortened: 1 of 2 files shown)


ggalmazor:
About the Central and the CLI, #764 takes care of that. I suppose we should merge this PR first, then #764 and then look whether we need to slip a couple of commits to make the form id optional there as well. I'd suggest we wait until all related PRs get merged.

kkrawczyk123:
In general, bulk push without specifying id works but I have noticed that pushing a single form when using id does not work at all. This is the cmd output:
Screenshot from 2019-08-20 11-31-05
@acj Can you take a look at that issue? 

cc @ggalmazor 

acj:
@kkrawczyk123 Eek, good catch. It should be fixed now.

@ggalmazor Waiting until the related PRs are merged sounds fine to me. I'll open a PR for the docs change.

Thanks, all, for the feedback.

kkrawczyk123:

kkrawczyk123:
**Tested with success!**

@opendatakit-bot label 'behavior verified"

kkrawczyk123:

PR #790, opened 2019-08-19
Issue #789 Catch NPE while populating dynamic choices

### LINKED ISSUE #789: Handle NPE populating choices of fields that depend on other fields inside repeats
#### Software versions 
Briefcase v1.16.2

#### Problem description
Briefcase has to collect all the controls in a form to support the "Split select multiple" export feature. It does so by traversing the form's model in search of `QuestionDef` nodes. 

When one of those controls has a dynamic choice list that doesn't use external secondary instances, Briefcase will populate it. This is normally because that field uses an internal secondary instance combined with a predicate as a source of choices, which has been fine so far.

With the introduction of https://github.com/opendatakit/javarosa/pull/437 () Briefcase will cause an NPE while traversing the form if there's a control inside a repeat group, with dynamic choices and a predicate that references other fields or the parent relative path `..`.

It's important to note that this is not a JavaRosa bug. Briefcase will cause an NPE because it traverses the form using the model tree's contents without instantiating new repeat groups as a normal client would do. This causes an NPE at https://github.com/opendatakit/javarosa/pull/437/commits/91d32d904d97696c584ffe481627e6fd60dbe850#diff-b60f702426880ccc5f1b079019104bd4R1016 because the `curQRef` reference has invalid multiplicities.

Try to export this form with the "Split select multiple" option enabled [populate_dynamic_choices_npe.zip](https://github.com/opendatakit/briefcase/files/3514698/populate_dynamic_choices_npe.zip)

This test will reproduce the issue:
```java
package org.opendatakit.briefcase.export;

import org.junit.After;
import org.junit.Before;
import org.junit.Test;

public class CocoteroTest {
  private ExportToCsvScenario scenario;

  @Before
  public void setUp() {
    scenario = ExportToCsvScenario.setUp("populate_dynamic_choices_npe");
  }

  @After
  public void tearDown() {
    scenario.tearDown();
  }

  @Test
  public void exports_forms_with_all_data_types() {
    scenario.runExportExplodedChoiceLists();
  }
}
```

#### Expected behavior
- Exports should complete without failure
- The affected controls should be exported without splitting their values.

#### Other information 
I've already considered some courses of action:
- Catch the NPE
- Include a null check at https://github.com/opendatakit/javarosa/pull/437/commits/91d32d904d97696c584ffe481627e6fd60dbe850#diff-b60f702426880ccc5f1b079019104bd4R1016 to avoid the NPE
- Change the way we get the controls list to imitate what a client would do: instantiate a form and visit each question until we hit the form's end. 

I don't like any of the options but the first one (catching) feels like the safest thing to do. Changing the way we traverse the form could get us into other issues with required fields and non-supported form states, or a performance loss. Changing JR feels unjustified and makes the release harder to manage.

ggalmazor:
Closes #789

- Added an automated regression test
- Manually exported the form attached in the issue and verified that it exports correctly when combined with the "split select multiples" feature.

Test this PR with [briefcase_pr_790.zip](https://github.com/opendatakit/briefcase/files/3515126/briefcase_pr_790.zip)

This is what feels safest from the three options described in #789 (maybe there are more options?). It's a straightforward change to prevent an NPE.

There's the risk of masking other sources of NPEs by the added try/catch block.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/export/FormDefinition.java
+++ b/src/org/opendatakit/briefcase/export/FormDefinition.java
@@ -163,7 +163,11 @@ public class FormDefinition {
             // Populate choices of any control using a secondary
             // instance that is not external
             if (secondaryInstance != null && !(secondaryInstance instanceof ExternalDataInstance))
-              formDef.populateDynamicChoices(itemsetBinding, (TreeReference) control.getBind().getReference());
+              try {
+                formDef.populateDynamicChoices(itemsetBinding, (TreeReference) control.getBind().getReference());
+              } catch (NullPointerException e) {
+                // Ignore (see https://github.com/opendatakit/briefcase/issues/789)
+              }
           }
         })
         .collect(toMap(FormDefinition::controlFqn, e -> e));
--- a/test/java/org/opendatakit/briefcase/export/ExportToCsvExplodeChoiceListWithRelativePredicateInRepeatTest.java
+++ b/test/java/org/opendatakit/briefcase/export/ExportToCsvExplodeChoiceListWithRelativePredicateInRepeatTest.java
@@ -1,0 +1,26 @@
+package org.opendatakit.briefcase.export;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ExportToCsvExplodeChoiceListWithRelativePredicateInRepeatTest {
+  private ExportToCsvScenario scenario;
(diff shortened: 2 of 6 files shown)


aurdipas:
tested successfully with the form generating the error. Thanks

kkrawczyk123:
**Tested with success!!**

PR #792, opened 2019-08-19
Issue #791 Match case when finding xml element

### LINKED ISSUE #791: Match case when finding xml elements
#### Software versions 
Briefcase v1.12

#### Problem description
When a form has two fields with the same name in different cases e.g. `Field` and `field`, it can get into problems when they don't share the same data type. In that case, we can't predict the matcher that will be used, which will produce errors when trying to transform incompatible raw values (as in, trying to parse a `2` as a datetime value will produce a parsing error).

Use the following form and try to export it: 
[simple-form-case-sensitivity.zip](https://github.com/opendatakit/briefcase/files/3514876/simple-form-case-sensitivity.zip)

#### Expected behavior
The export should complete correctly having applied correct data transformations.

#### Other information 
The `XmlElement` class introduced in v1.12 has two `equalsIgnoreCase()` calls that are responsible for this errors.

ggalmazor:
Closes #791

Added automated regression test
Manually exported the form attached to the issue

Test this PR with [briefcase_pr_792.zip](https://github.com/opendatakit/briefcase/files/3515127/briefcase_pr_792.zip)

Super narrow fix to avoid confusing field mappers when two fields have the same name with case differences.

This shouldn't change any other behavior.

Nope.

### DIFF
--- a/src/org/opendatakit/briefcase/export/XmlElement.java
+++ b/src/org/opendatakit/briefcase/export/XmlElement.java
@@ -144,7 +144,7 @@ public class XmlElement {
   public List<XmlElement> findElements(String... namesArray) {
 
     if (namesArray.length == 1)
-      return childrenOf().stream().filter(e -> e.getName().equalsIgnoreCase(namesArray[0])).collect(Collectors.toList());
+      return childrenOf().stream().filter(e -> e.getName().equals(namesArray[0])).collect(Collectors.toList());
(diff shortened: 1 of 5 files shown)


aurdipas:
tested successfully with the form that was generating the error. Thanks also for this +1

kkrawczyk123:
**Tested with success!!**

PR #795, opened 2019-09-02
Issue #775 Push all forms when form_id is not specified

### LINKED ISSUE #775: Push all forms when form_id is not specified

ggalmazor:
Closes #775

This PR solves the issue for Central. Aggregate has been already addressed in another PR.

Test manually in my computer using the central sandbox.

This is a narrow fix that uses the same code from the push to aggregate operation.

There shouldn't be any regression risk since this change uses the same code as another PR that was tested and verified.

Yes. https://github.com/opendatakit/docs/issues/1100

### DIFF
--- a/src/org/opendatakit/briefcase/operations/PushFormToCentral.java
+++ b/src/org/opendatakit/briefcase/operations/PushFormToCentral.java
@@ -26,8 +26,8 @@ import static org.opendatakit.briefcase.reused.http.Http.DEFAULT_HTTP_CONNECTION
 
 import java.nio.file.Path;
 import java.util.Arrays;
-import java.util.Collections;
-import java.util.Optional;
+import java.util.List;
+import java.util.stream.Collectors;
 import org.opendatakit.briefcase.model.BriefcasePreferences;
 import org.opendatakit.briefcase.model.FormStatus;
 import org.opendatakit.briefcase.model.FormStatusEvent;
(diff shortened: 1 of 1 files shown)


src/org/opendatakit/briefcase/operations/PushFormToCentral.java
```diff
@@ -78,14 +78,22 @@ private static void pushToCentral(Args args) {
     String token = http.execute(server.getSessionTokenRequest())
         .orElseThrow(() -> new BriefcaseException("Can't authenticate with ODK Central"));
 
-    String formId = args.get(FORM_ID);
-    Optional<FormStatus> maybeFormStatus = formCache.getForms().stream()
-        .filter(form -> form.getFormId().equals(formId))
-        .map(FormStatus::new)
-        .findFirst();
+    List<FormStatus> statuses;
+    if (args.getOptional(FORM_ID).isPresent()) {
```
yanokwa: Curious why the Aggregate code doesn't use optionals.

ggalmazor:
Great idea! I think @acj might overlap with my timezone a little better than @dcbriccetti in case he's off. @acj feel free to open a review and leave some feedback ;)

dcbriccetti:
Looking now.

PR #813, opened 2019-09-13
Issue 812 GUI fails to load after CLI used 

### LINKED ISSUE #812: GUI fails to load after CLI used
### Software versions 
At least since Briefcase v1.16 
java version "10.0.1" 2018-04-17
Java(TM) SE Runtime Environment 18.3 (build 10.0.1+10)
Java HotSpot(TM) 64-Bit Server VM 18.3 (build 10.0.1+10, mixed mode)
OS X 10.14.6

#### Problem description

Since at least 1.16, briefcase GUI fails to run if the CLI has previously been used. 

Error message

>The Java JAR file "ODK-Briefcase-v1.17.0-beta.0.jar" could not be launched" Check the Console for possible error messages.

Console output with keyword ```java``` reads as follows
 
```
default	14:56:54.887449 +0100	com.apple.WebKit.WebContent	Current memory footprint: 188 MB
default	14:56:55.844284 +0100	kernel	AGC:: [java pid 27676 mux-aware] exiting, non-mux-aware app count 4, runtime: 0:00:02.301
default	14:56:55.844439 +0100	kernel	AGC:: [java pid 27676 non-mux-aware] exiting, non-mux-aware app count 3, runtime: 0:00:02.302
default	14:56:55.844277 +0100	hidd	Connection removed: IOHIDEventSystemConnection uuid:80B90C1A-F4BC-4411-A909-B6196179FFE4 pid:27676 process:java type:Passive entitlements:0x0 caller:HIToolbox: ___GetIOHIDEventSystemClient_block_invoke + 26 attributes:(null) inactive:0
default	14:56:58.564398 +0100	com.apple.WebKit.WebContent	Current memory footprint: 281 MB
```

Pull and export with CLI and then try to start GUI. 

#### Other information 

Performing this command on CLI fixes the problem

```
java -jar ODK-Briefcase-v1.17.0-beta.0.jar -c
```

ggalmazor:
Closes #812

In this PR
- [ ] Pending to reproduce and fix the original issue reported in #812
- [x] Avoid losing prefs by runnnig a CLI op

Testing of this PR involves this steps:
- Run the `-c` CLI op to clear all preferences
- Ensure the GUI environment is set and configured by running the GUI once and setting an sd
- Run a pull from Central CLI op
- Run the GUI again

In master, you are asked to set an sd for a second time, which shouldn't happen.

This is a super narrow fix. No other approaches were considered.

Somehow, setting the sd in the pull CLI ops was losing prefs for the GUI, and users were asked to set it again in the next GUI launch.

Also, setting the sd in CLI ops is not required (downstream code gets the briefcaseDir directly), so it's a safe change to do.

I tried to track uses of `BriefcasePreferences.getStorageDir()` in downstream code that would require the call to `BriefcasePreferences.setStorageDir()` I've removed and I found none. I could have missed something, though.

No.

### DIFF
--- a/src/org/opendatakit/briefcase/operations/PullFormFromAggregate.java
+++ b/src/org/opendatakit/briefcase/operations/PullFormFromAggregate.java
@@ -87,7 +87,6 @@ public class PullFormFromAggregate {
     FormCache formCache = FormCache.from(briefcaseDir);
     formCache.update();
     BriefcasePreferences appPreferences = BriefcasePreferences.appScoped();
-    appPreferences.setStorageDir(briefcaseDir);
     FormMetadataPort formMetadataPort = FileSystemFormMetadataAdapter.at(briefcaseDir);
 
     int maxHttpConnections = Optionals.race(
--- a/src/org/opendatakit/briefcase/operations/PullFormFromCentral.java
+++ b/src/org/opendatakit/briefcase/operations/PullFormFromCentral.java
@@ -70,7 +70,6 @@ public class PullFormFromCentral {
     FormCache formCache = FormCache.from(briefcaseDir);
     formCache.update();
     BriefcasePreferences appPreferences = BriefcasePreferences.appScoped();
-    appPreferences.setStorageDir(briefcaseDir);
     FormMetadataPort formMetadataPort = FileSystemFormMetadataAdapter.at(briefcaseDir);
 
     int maxHttpConnections = Optionals.race(


ggalmazor:
I think it doesn't make sense to block the release for more time. If the feedback comes and we can repro the original issue, we will release v1.17.1 with the fix.

chrissyhroberts:
Hi @ggalmazor 
Sorry have been super busy and not managed to keep up with you on this. 
I guess as this flagged up an issue with preferences, it is possible that the fix you've made will change the behaviour I was seeing, so suggest we wait until next release and check to see if this fixes it. Thanks so much.

ggalmazor:
No problem @chrissyhroberts! Thank you for understanding :)

PR #816, opened 2019-09-20
815 main window resizing

### LINKED ISSUE #815: "Update available" text doesn't wrap properly
When I run master, this is what the "Update available" looks like. I think it should always be on one line and not wrap. Not sure if this is a macOS thing. My guess is that it's not easily fixable. Any ideas, @dcbriccetti ?

*On launch* -  😄
Screen Shot 2019-09-17 at 12 39 08 PM

*Make window bigger* - 😟
Screen Shot 2019-09-17 at 12 39 14 PM

*Make window smaller* -  😭
Screen Shot 2019-09-17 at 12 39 21 PM

dcbriccetti:
Closes #815

Please see the commit comments for details of the changes.

Not enough. It looks great on a Mac, but I’ll need help checking on other platforms.
Could someone please check that the underline is gone and that the hyperlink still works when Aggregate or Central is the source or target?

Using a ContentPane and BorderLayout seems best. 

Should have no effect.

No

### DIFF
--- a/src/org/opendatakit/briefcase/ui/reused/transfer/sourcetarget/target/Central.java
+++ b/src/org/opendatakit/briefcase/ui/reused/transfer/sourcetarget/target/Central.java
@@ -85,7 +85,7 @@ public class Central implements PushTarget<CentralServer> {
   public void decorate(JLabel label) {
     label.setText("" +
         "<html>" +
-        "URL: <a href=\"" + server.getBaseUrl().toString() + "\">" + getDescription() + "</a><br/>" +
+        "URL: <a style=\"text-decoration: none\" href=\"" + server.getBaseUrl().toString() + "\">" + getDescription() + "</a><br/>" +
         "Project ID: " + server.getProjectId() + "" +
         "</html>");
     makeClickable(label, () -> uncheckedBrowse(server.getBaseUrl()));
--- a/src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java
+++ b/src/org/opendatakit/briefcase/ui/MainBriefcaseWindow.java
@@ -73,7 +73,7 @@ public class MainBriefcaseWindow {
 
   private final JFrame frame = new JFrame();
   private final JTabbedPane tabbedPane = new JTabbedPane(JTabbedPane.TOP);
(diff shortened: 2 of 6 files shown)


ggalmazor:
> Perhaps @ggalmazor can fix the unchecked call to put in makeClickable.

I don't see a way to remove that unchecked call. I think we need to @Supress("unchecked") to the overloaded method. 

Why not assuming that `boolean underline` will always be true and simplify the `if` block? We can extract a method to get the underlined font and isolate the unchecked operation there.

dcbriccetti:
@ggalmazor Thanks for that. What do you think about removing underlines from links throughout Briefcase?

ggalmazor:
> What do you think about removing underlines from links throughout Briefcase?

I don't have a strong opinion about it. I'll let @yanokwa and @kkrawczyk123 fight for it ;)

kkrawczyk123:
I don't have a strong opinion about that too and I am not protesting :)

yanokwa:
If the blue stays, the underline can go.

src/org/opendatakit/briefcase/ui/reused/transfer/sourcetarget/target/Central.java
```diff
@@ -85,7 +85,7 @@ public void onSelect(Container container) {
   public void decorate(JLabel label) {
     label.setText("" +
         "<html>" +
-        "URL: <a href=\"" + server.getBaseUrl().toString() + "\">" + getDescription() + "</a><br/>" +
+        "URL: <a style='text-decoration: none' href=\"" + server.getBaseUrl().toString() + "\">" + getDescription() + "</a><br/>" +
```
ggalmazor: Sorry to be nitpicky, but we should use double quotes for HTML tag attribute values. 

I know it's annoying to escape them :/

dcbriccetti: Happy to change, but why do you want those ugly double quotes that use twice as much ink?

dcbriccetti:
Yo, yo, yo, this is ready to go.
(I’m in no rush, but the Github prodder has appeared.)

PR #829, opened 2019-10-26
[WIP] - Fix #809: Allow time in start and end date

### LINKED ISSUE #809: Allow times in start and end filters of the export operation in the CLI
As per forum conversation, users would like to be able to define an export date range filter with time resolution.

Currently, the `-start`, and `-end` CLI args allow ISO8601 formatted dates (e.g. `2019-01-01`).

The enhancement would consist in allowing ISO8601 formatted datetimes (as described in [this doc](https://docs.google.com/document/d/1vPuEgapHvYWINnoa9QiG1R0HWMiCS4_8vtP_Zx3bGec/edit#heading=h.7z9gai9ka6ve)) in addition to just dates. Some examples: 
- `-start 2019-01-01T10:20:30.000Z`
- `-end 2019-02-03T20:30:45.123-05:00`

mayank8318:
Closes #809 
Still WIP
Facing a problem right now because the whole system is built around only using dates(just like the UI). This PR will require some major changes according to me so I want to clarify the design before going forward. The main culprit is the DateRange API which is used by the export tasks. It uses only LocalDate. So here are a few options I could think of:
i) Implement it using a Generic <T extends ...>
ii) Make a LocalDateTime API without touching this one
iii) Change this to use LocalDateTime (lot of refactoring needed including the tests)

CC - @ggalmazor

### DIFF
--- a/src/org/opendatakit/briefcase/operations/Export.java
+++ b/src/org/opendatakit/briefcase/operations/Export.java
@@ -96,7 +96,7 @@ public class Export {
       Arrays.asList(PEM_FILE, EXCLUDE_MEDIA, OVERWRITE, START, END, PULL_BEFORE, SPLIT_SELECT_MULTIPLES, INCLUDE_GEOJSON_EXPORT, REMOVE_GROUP_NAMES, SMART_APPEND)
   );
 
-  public static void export(String storageDir, String formid, Path exportDir, String baseFilename, boolean exportMedia, boolean overwriteFiles, boolean pullBefore, Optional<LocalDate> startDate, Optional<LocalDate> endDate, Optional<Path> maybePemFile, boolean splitSelectMultiples, boolean includeGeoJsonExport, boolean removeGroupNames, boolean smartAppend) {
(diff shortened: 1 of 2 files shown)


ggalmazor:
I agree with the assessment of the problem, @mayank8318. Thanks for pointing it out.

This change was intended to support the needs of some high-load operations that required fine-tuning the export filters. Now there's a new `--smart_append` feature that lets users do something similar.

Since the change will require bigger changes than anticipated, and considering the alternative `--smart_append` workaround, I wonder if we should really go ahead with this change.

Having said that, I think we should go with `OffsetDateTime` values anyway, because users might want to use time zones different than the local ones.

Regarding strategy and design, here's how we could do the change using parallel-refactoring:
- Rename `DateRange` into `LocalDateRange`
- Extract a `DateRange` interface from `LocalDateRange`
- Make all the code depend on `DateRange` instead of `LocalDateRange`
- Duplicate `LocalDateRange` into `OffsetDateTimeRange` and start refactoring that one with the help of the tests. Write more tests if needed to work out new behavior.
- At CLI operation level, instantiate a different instance of subclass of `DateRange` according to what the user provides.
- At UI operation level, always use `LocalDateRange`
- Create a test that verifies relationships between date and dateTime ranges: provided that the system is in UTC, verify that start at `2019-01-01` is the same as start at `2019-01-01T00:00:00.000Z`

ggalmazor:
Hi, @mayank8318!

I'm not sure if you had a moment to think about my last comment. To be clear, I'm inclined to close this PR and do nothing about #809 for the moment. How does that sound?

mayank8318:
Hi @ggalmazor sorry I couldn't just glance at the comment but didn't get time to think about the implementation. Yes, I agree with you. I'll close the PR :)

PR #830, opened 2019-11-14
Jksortingissue

jay4kelly:
Closes #331 

#### What has been done to verify that this works as intended?
Manual testing: Tested the three tabs(pull, push and export) using the sandbox aggregate server. Works as stated in the #331  issue. Test were performed on mac os and windows 10 intellij
gradlew ran to completion 390 succeded 6 ignored on mac os intellij.
Gradle failed on windows 10 (see details below)

#### Why is this the best possible solution? Were any other approaches considered?
Custom sorting in swing table using TableRowSorter typically use custom comparators for changing sort order. This solution is the standard one.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
Changes are described in the issue. The change is limited to when the column headers are clicked.

#### Does this change require updates to documentation? If so, please file an issue at https://github.com/opendatakit/docs/issues/new and include the link below.
No

Ran into issue on windows 10 on intellij ide while running gradlew (verification -> check)
 [ant:checkstyle] [ERROR] C:\Users\jay4k\IdeaProjects\briefcase\src\org\opendatakit\common\utils\WebUtils.java:182: Do not use Windows line endings [RegexpMultiline]
> Task :checkstyleMain FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':checkstyleMain'.
This happened for hundreds of lines on many files that were untouched by me.
Tried altering git configuration and intellij configuration for end of line handling. Made sure codestyle for ide was the one supplied in the project.  

On mac os intellij tests all run to completion. Unsure how to fix this on windows.

PR #845, opened 2019-12-04
Buttonclass

### LINKED ISSUE #838: Reduce indirection and loose coupling with UI colors
#830 introduces new UI table sorting methods that are loosely coupled with the colors we're using to represent active/inactive buttons. The methods also introduce indirection between "something having X color" and "that something having Y business meaning".

We should extract all related business concepts into some artifact that can be reused in both placed. For example, a `ExportConfigurationButton` class, could have methods such as `isSelected(JButton button)`, or `sort(JButton o1, JButton o2)` among other. The goal would be to have all the business needs related to export conf buttons in a single place that can be used everywhere.

jay4kelly:
Closes #838 

Tested using UI for all buttons and chevrons. 
Created classes called DetailsStatusButton and ExportConfigurationButton. The selection column didn't need a class for it as there was no business logic incorporated. So we continued to use a compareSelectionfunction inside the UI class.
Localized just to the UI tests conducted.
### Does this change require updates to documentation? If so, please file an issue at https://github.com/opendatakit/docs/issues/new and include the link below.
No

### DIFF
--- a/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
+++ b/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
@@ -65,7 +65,7 @@ public class ExportFormsTableView extends JTable {
     columns.getColumn(SELECTED_CHECKBOX_COL).setMinWidth(40);
     columns.getColumn(SELECTED_CHECKBOX_COL).setMaxWidth(40);
     columns.getColumn(SELECTED_CHECKBOX_COL).setPreferredWidth(40);
-    columns.getColumn(OVERRIDE_CONF_COL).setCellRenderer(UI::cellWithButton);
+    columns.getColumn(OVERRIDE_CONF_COL).setCellRenderer(UI::cellWithExportConfButton);
     columns.getColumn(OVERRIDE_CONF_COL).setMinWidth(40);
     columns.getColumn(OVERRIDE_CONF_COL).setMaxWidth(40);
     columns.getColumn(OVERRIDE_CONF_COL).setPreferredWidth(40);
@@ -83,7 +83,7 @@ public class ExportFormsTableView extends JTable {
     setFillsViewportHeight(true);
 
     TableRowSorter<ExportFormsTableViewModel> sorter = sortBy(getModel(), FORM_NAME_COL, ASCENDING);
-    sorter.setComparator(SELECTED_CHECKBOX_COL, (Comparator<Boolean>) UI::compareSelectionButton);
+    sorter.setComparator(OVERRIDE_CONF_COL, (Comparator<ExportConfigurationButton>) ExportConfigurationButton::compare);
     setRowSorter(sorter);
     sorter.sort();
   }
--- a/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
+++ b/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
@@ -75,7 +75,7 @@ public class ExportFormsTableView extends JTable {
     columns.getColumn(EXPORT_STATUS_COL).setPreferredWidth(exportStatusDims.width + 25);
     columns.getColumn(LAST_EXPORT_COL).setMinWidth(lastExportDims.width + 25);
     columns.getColumn(LAST_EXPORT_COL).setPreferredWidth(lastExportDims.width + 25);
-    columns.getColumn(DETAIL_BUTTON_COL).setCellRenderer(UI::cellWithButton);
+    columns.getColumn(DETAIL_BUTTON_COL).setCellRenderer(UI::cellWithDetailButton);
     columns.getColumn(DETAIL_BUTTON_COL).setMinWidth(40);
     columns.getColumn(DETAIL_BUTTON_COL).setMaxWidth(40);
     columns.getColumn(DETAIL_BUTTON_COL).setPreferredWidth(40);
--- a/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
+++ b/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
@@ -23,7 +23,6 @@ import static org.opendatakit.briefcase.ui.reused.TableCustomizer.getHeaderDimen
 import java.awt.Dimension;
 import java.awt.event.MouseEvent;
 import java.util.Collections;
-import java.util.Comparator;
 import javax.swing.JButton;
 import javax.swing.JTable;
 import javax.swing.RowSorter;
@@ -31,12 +30,13 @@ import javax.swing.SortOrder;
 import javax.swing.table.TableColumnModel;
 import javax.swing.table.TableModel;
 import javax.swing.table.TableRowSorter;
+import org.opendatakit.briefcase.ui.reused.ExportConfigurationButton;
 import org.opendatakit.briefcase.ui.reused.MouseAdapterBuilder;
 import org.opendatakit.briefcase.ui.reused.UI;
 
 public class ExportFormsTableView extends JTable {
   static final String[] HEADERS = new String[]{"", "", "Form Name", "Export Status", "Last Export", ""};
-  static final Class[] TYPES = new Class[]{Boolean.class, JButton.class, String.class, String.class, String.class, JButton.class};
+  static final Class[] TYPES = new Class[]{Boolean.class, ExportConfigurationButton.class, String.class, String.class, String.class, JButton.class};
   static final boolean[] EDITABLE_COLS = new boolean[]{true, false, false, false, false, false};
 
   public static final int SELECTED_CHECKBOX_COL = 0;
@@ -81,9 +81,6 @@ public class ExportFormsTableView extends JTable {
     setFillsViewportHeight(true);
 
     TableRowSorter<ExportFormsTableViewModel> sorter = sortBy(getModel(), FORM_NAME_COL, ASCENDING);
-    sorter.setComparator(OVERRIDE_CONF_COL, (Comparator<JButton>) UI::compareConfButton);
-    sorter.setComparator(SELECTED_CHECKBOX_COL, (Comparator<Boolean>) UI::compareSelectionButton);
-    sorter.setComparator(DETAIL_BUTTON_COL, (Comparator<JButton>) UI::compareDetailsButton);
     setRowSorter(sorter);
     sorter.sort();
   }
--- a/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
+++ b/src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
@@ -65,7 +65,7 @@ public class ExportFormsTableView extends JTable {
     columns.getColumn(SELECTED_CHECKBOX_COL).setMinWidth(40);
     columns.getColumn(SELECTED_CHECKBOX_COL).setMaxWidth(40);
     columns.getColumn(SELECTED_CHECKBOX_COL).setPreferredWidth(40);
-    columns.getColumn(OVERRIDE_CONF_COL).setCellRenderer(UI::cellWithButton);
+    columns.getColumn(OVERRIDE_CONF_COL).setCellRenderer(ButtonProcessing::cellWithButton);
     columns.getColumn(OVERRIDE_CONF_COL).setMinWidth(40);
     columns.getColumn(OVERRIDE_CONF_COL).setMaxWidth(40);
     columns.getColumn(OVERRIDE_CONF_COL).setPreferredWidth(40);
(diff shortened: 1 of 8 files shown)


src/org/opendatakit/briefcase/ui/reused/ExportConfigurationButton.java
```diff
@@ -0,0 +1,62 @@
...
+import java.awt.Color;
+import java.awt.Font;
+import java.awt.Insets;
+import javax.swing.JButton;
+
+
+/**
+ * Contains application logic pertaining to the Export Configuration Button
+ */
+public class ExportConfigurationButton extends JButton implements Comparable<ExportConfigurationButton> {
```
ggalmazor: The following may seem strange, but we always want to stay away from extending classes when we're not implementing Polymorphism, which is not the case in this instance, because we're not really trying to [replace conditional logic](https://refactoring.com/catalog/replaceConditionalWithPolymorphism.html).

I've gone ahead and prepared an alternative implementation of `ExportConfigurationButton` that wraps around a `Jbutton` instance and has the same behavior without extending `JButton`.  You can check it here: https://github.com/ggalmazor/briefcase/commit/e7bb0891403a54a13892f9d6439f753e9536d363

jay4kelly:
Changed DetailStatusButton and ExportConfigurationButton to not extend JButton. Added ButtonProcessing Interface. Could have added a CommonButton class for the button classes to extend instead of using the interface. I think this would require us to remove the static create method on the button classes and use a new operator instead.

src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableViewModel.java
```diff
@@ -115,14 +104,15 @@ private void removeConfiguration(FormStatus form) {
...
-  private void updateDetailButton(FormStatus form, JButton button) {
-    button.setForeground(form.getStatusHistory().isEmpty() ? LIGHT_GRAY : DARK_GRAY);
+  private void updateDetailButton(FormStatus form, DetailsStatusButton button) {
+    button.setStatus(!(form.getStatusHistory().isEmpty()));
   }
 
-  private void updateConfButton(FormStatus form, JButton button) {
-    button.setForeground(forms.hasConfiguration(form) ? NO_CONF_OVERRIDE_COLOR : DARK_GRAY);
+  private void updateConfButton(FormStatus form, ExportConfigurationButton button) {
+    button.setStatus(forms.hasConfiguration(form));
```
ggalmazor: I think `button.setStatus()` is kinda unexpected because the meaning of "status" in a conf button is different than the one in a details button. I believe you had to have the same name for both classes because they extend it from `ButtonProcessing`, but this is not really a requirement, since you're using the subclasses (`DetailStatusButton` and `ExportConfigurationButton`) instead of the interface, thus not doing Polymorphism (which is totally fine here, by the way).

This is telling us that the `setStatus` method is not part of the `ButtonProcessing` interface. I'd suggest you remove it from the interface and take the chance to give it a more specific name in `ExportConfigurationButton` e.g. `setConfigured`.

src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableViewModel.java
```diff
@@ -157,7 +147,7 @@ public Object getValueAt(int rowIndex, int columnIndex) {
   }
 
   @Override
-  // Suppressing next ParameterName checkstyle error becasue 'aValue' param triggers it by mistake
+  // Suppressing next ParameterName checkstyle error because 'aValue' param triggers it by mistake
```
ggalmazor: Good catch! ;)

src/org/opendatakit/briefcase/ui/reused/ButtonProcessing.java
```diff
@@ -0,0 +1,33 @@
...
+package org.opendatakit.briefcase.ui.reused;
+
+import javax.swing.JButton;
+
+/**
+ * Defines common behavior used in buttons
+ */
+public interface ButtonProcessing {
+
+  void onClick(Runnable callback);
```
ggalmazor: I believe that `getJButton()` would be the only method that needs to exist in this interface. All the other methods are accessed using the subclasses.

Another method in this interface could be the one I suggested in a previous comment i.e. `void click(Runnable callback)` method.

src/org/opendatakit/briefcase/ui/reused/DetailsStatusButton.java
```diff
@@ -0,0 +1,72 @@
...
+package org.opendatakit.briefcase.ui.reused;
+
+import static java.awt.Color.DARK_GRAY;
+import static java.awt.Color.LIGHT_GRAY;
+
+import java.awt.Font;
+import java.awt.Insets;
+import javax.swing.JButton;
+
+public class DetailsStatusButton implements ButtonProcessing, Comparable<ButtonProcessing> {
```
ggalmazor: I believe implementing `Comparable<ButtonProcessing>` is not needed because we will use a reference to a static method.

src/org/opendatakit/briefcase/ui/reused/DetailsStatusButton.java
```diff
@@ -0,0 +1,72 @@
...
+  public JButton getJButton() {
+    return button;
+  }
+
+  public void setStatus (boolean value){
+    status = value;
+    button.setForeground(status ?  DARK_GRAY : LIGHT_GRAY );
+  }
+
+  public int compareTo(ButtonProcessing b) {
```
ggalmazor: So, this method should be migrated to `public static int compare(DetailsStatusButton a, DetailsStatusButton b)`

src/org/opendatakit/briefcase/ui/reused/ExportConfigurationButton.java
```diff
@@ -0,0 +1,74 @@
...
+import java.awt.Color;
+import java.awt.Font;
+import java.awt.Insets;
+import javax.swing.JButton;
+
+
+/**
+ * Contains application logic pertaining to the Export Configuration Button
+ */
+public class ExportConfigurationButton implements ButtonProcessing, Comparable<ButtonProcessing> {
```
ggalmazor: Same, here: we shouldn't need to implement `Comparable`, and the `compareTo` method should be static and take two `ExportConfigurationButton` objects

src/org/opendatakit/briefcase/ui/reused/UI.java
```diff
@@ -62,24 +59,18 @@
 public class UI {
   private static final Font ic_receipt = FontUtils.getCustomFont("ic_receipt.ttf", 16f);
 
-  @SuppressWarnings("checkstyle:AvoidEscapedUnicodeCharacters")
-  public static JButton buildDetailButton(FormStatus form) {
+  public static DetailsStatusButton buildDetailButton(FormStatus form) {
```
ggalmazor: I would expect this to be a static factory in `DetailsStatusButton`

src/org/opendatakit/briefcase/ui/reused/transfer/TransferFormsTableView.java
```diff
@@ -72,6 +74,7 @@
     setFillsViewportHeight(true);
 
     TableRowSorter<TransferFormsTableViewModel> sorter = sortBy(getModel(), FORM_NAME_COL, ASCENDING);
+    sorter.setComparator(SELECTED_CHECKBOX_COL, (Comparator<Boolean>) UI::compareSelectionButton);
```
ggalmazor: I believe we need to set another comparator for the details column as in:
```
sorter.setComparator(DETAIL_BUTTON_COL, (Comparator<DetailsStatusButton>) DetailsStatusButton::compare);

File: src/org/opendatakit/briefcase/ui/reused/transfer/TransferFormsTableView.java [hunk 10]
```diff
@@ -86,8 +89,8 @@ private void relayClickToButton(MouseEvent event) {
 
     if (row < getRowCount() && row >= 0 && column < getColumnCount() && column >= 0) {
       Object value = getValueAt(row, column);
-      if (value instanceof JButton)
-        ((JButton) value).doClick();
+      if (value instanceof ButtonProcessing)
+        ((ButtonProcessing) value).getJButton().doClick();
```
ggalmazor: I'd expect `((ButtonProcessing) value).click();` here.

src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
```diff
@@ -80,6 +83,7 @@
     setFillsViewportHeight(true);
 
     TableRowSorter<ExportFormsTableViewModel> sorter = sortBy(getModel(), FORM_NAME_COL, ASCENDING);
+    sorter.setComparator(SELECTED_CHECKBOX_COL, (Comparator<Boolean>) UI::compareSelectionButton);
```
ggalmazor: I believe we need to set a couple more comparators for the details and conf button columns as in:
```
sorter.setComparator(OVERRIDE_CONF_COL, (Comparator<ExportConfigurationButton>) ExportConfigurationButton::compare);
sorter.setComparator(DETAIL_BUTTON_COL, (Comparator<DetailsStatusButton>) DetailsStatusButton::compare);

File: src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java [hunk 12]
```diff
@@ -90,8 +94,8 @@ private void relayClickToButton(MouseEvent event) {
 
     if (row < getRowCount() && row >= 0 && column < getColumnCount() && column >= 0) {
       Object value = getValueAt(row, column);
-      if (value instanceof JButton)
-        ((JButton) value).doClick();
+      if (value instanceof ButtonProcessing)
+        ((ButtonProcessing) value).getJButton().doClick();
```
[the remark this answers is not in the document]
jay4kelly: Tried that but the doClick call was required to be performed by a JButton class. We can recheck that if you think it should work.

src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableViewModel.java
jay4kelly: Sure.

src/org/opendatakit/briefcase/ui/reused/ButtonProcessing.java
jay4kelly: Still have the onClick method in the interface.  DId not fully understand how to implement the change of button.addActionListener(__ ->   to  button.onClick(().
Can you explain how the "__" reference is used? I have not seen that before.

src/org/opendatakit/briefcase/ui/reused/DetailsStatusButton.java
jay4kelly: TableRowSorter is referencing the column class of the table column.  And as long as that class implements the Comparable interface, there is no need to pass in a comparator.

src/org/opendatakit/briefcase/ui/reused/DetailsStatusButton.java
jay4kelly: See earlier comment on Comparable

src/org/opendatakit/briefcase/ui/reused/UI.java
jay4kelly: Agreed. Made the change.

src/org/opendatakit/briefcase/ui/reused/transfer/TransferFormsTableView.java
jay4kelly: See earlier comment on Comparable

src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
jay4kelly: See earlier comment on Comparable

src/org/opendatakit/briefcase/ui/reused/ButtonProcessing.java
ggalmazor: My bad! This is something I do when I'm forced to declare an arg that I will ignore in the lambda's body block. `__` is simply the shortest name a variable can have using `_` characters, and the goal would be to implicitly say that the can be ignored.

src/org/opendatakit/briefcase/ui/reused/DetailsStatusButton.java
ggalmazor: Huh! I didn't notice that... Thanks for explaining :)

src/org/opendatakit/briefcase/ui/export/components/ExportFormsTableView.java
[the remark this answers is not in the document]
ggalmazor: I was thinking about adding a `ButtonProcessing.doClick` method that wraps `button.doClick`, to avoid breaking encapsulation.

ggalmazor:
Hi, @jay4kelly!  

After exploring a bit more the changes I was suggesting, I've come to the conclusion that they don't bring that much value overall.

This PR is looking good to me now :) Thanks for your work!!

jay4kelly:
Closes #838

kkrawczyk123:
**Tested with success!**
I haven't' noticed any regression. I have confirmed that the issue identified in #830 has been fixed. 

ggalmazor:
Hi, @jay4kelly! I've rebased your branch on top of master to be able to merge it. Thanks for your work!

Issue #798, opened 2019-09-04
Form list displayed without connection URL sometimes randomly


kkrawczyk123:
#### Software versions 
Briefcase v1.17.x, Java v1.11.x, operating system, macOS and Ubuntu 16

#### Problem description
Form list seems to display but the connection to specific aggregate is not visible, the user is able to pull forms with using the displayed form list. Screenshot attached: 
![Screenshot from 2019-09-04 14-51-30](https://user-images.githubusercontent.com/33342488/64256861-8d6cc980-cf24-11e9-9050-d335e29b5a3e.png)

#### Steps to reproduce the problem
I am able to reproduce it in steps:
1. Connect to aggregate which has a lot of forms which are loaded slowly
2. Click on Reset button immediately when it is visible

Reproduction steps show a very rare case, but it sometimes happens randomly and I am not 100% sure why. 

#### Expected behavior
Connection URL should be visible on the Pull tab when log in to Aggregate like always.

#### Other information 
the content of log file:
2019-09-04 09:59:29,448 [ForkJoinPool-27-worker-3] ERROR o.o.briefcase.reused.http.Request - Error
org.opendatakit.briefcase.reused.BriefcaseException: org.xmlpull.v1.XmlPullParserException: attr value delimiter missing! (position:START_TAG <html lang='null'>@1:26 in java.io.InputStreamReader@652b7f34)
    at org.opendatakit.briefcase.reused.http.RequestBuilder.readXmlElement(RequestBuilder.java:118)
    at java.base/java.util.function.Function.lambda$andThen$1(Function.java:88)
    at org.opendatakit.briefcase.reused.http.Request.map(Request.java:62)
    at org.opendatakit.briefcase.reused.http.response.Success.from(Success.java:47)
    at org.opendatakit.briefcase.reused.http.response.Response.from(Response.java:43)
    at org.opendatakit.briefcase.reused.http.CommonsHttp.lambda$uncheckedExecute$1(CommonsHttp.java:130)
    at org.apache.http.client.fluent.Response.handleResponse(Response.java:90)
    at org.opendatakit.briefcase.reused.http.CommonsHttp.uncheckedExecute(CommonsHttp.java:130)
    at org.opendatakit.briefcase.reused.http.CommonsHttp.execute(CommonsHttp.java:87)
    at org.opendatakit.briefcase.ui.reused.transfer.sourcetarget.source.PullSource.lambda$aggregate$0(PullSource.java:35)
[...]

PR #851, opened 2020-01-22
A solution for Bug #798

gilbva:
The bug was caused by the fact that the Source/Target panel has no information that the forms are being loaded from the server. Making a call to setWorking() in the current view makes the buttons in the Source/Target panel to go disabled until the unsetWorking method is call. I think this fixes the bug as the user won be able to click on reset (or reload) during the time the forms are being loaded, and also the user will have good feedback.

Closes #

I was able to reproduce the bug before the fix, after the fix i cannot hit the reset button while loading.

It uses the already created method setWorking that is intended for this kind of behavior

The user won be able to cancel the form loading process, but i think that is ok.

I dont think this need to be put explicitly in the documentation as is a minor fix.

### DIFF
--- a/src/org/opendatakit/briefcase/ui/pull/PullPanel.java
+++ b/src/org/opendatakit/briefcase/ui/pull/PullPanel.java
@@ -22,7 +22,8 @@ import static org.opendatakit.briefcase.reused.job.Job.run;
 import static org.opendatakit.briefcase.ui.reused.UI.errorMessage;
 
 import java.util.Optional;
-import javax.swing.*;
+import javax.swing.JPanel;
+import javax.swing.SwingUtilities;
 
 import org.bushe.swing.event.EventBus;
 import org.bushe.swing.event.annotation.AnnotationProcessor;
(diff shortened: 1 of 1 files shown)


src/org/opendatakit/briefcase/ui/pull/PullPanel.java
```diff
@@ -22,7 +22,8 @@
 import static org.opendatakit.briefcase.ui.reused.UI.errorMessage;
 
 import java.util.Optional;
-import javax.swing.JPanel;
+import javax.swing.*;
```
ggalmazor: This makes checkstyle angry :P 

We have a rule that forbids using `.*` imports.

gilbva: Ok, I'm glad to be helpful, I will review my IDE configuration on imports next time. :)

kkrawczyk123:
**Tested with success!**
I was not able to reproduce steps from the issue. I have also performed some exploratory testing of changing servers, url connections, between pull and push tab.

PR #855, opened 2020-02-24
bugFix to be able to push form to aggregate 

### LINKED ISSUE #854: On pushToAggregate request header "X-OpenRosa-Version -> 1.0" is set twice and give a java Error under aggregate cause it parsing it as double "1.0, 1,0"
#### Software versions 
Briefcase v1.7.2, Java v1.8, operating system, Aggregate v2.0.5

#### Problem description
Impossible to push aggregate forms

push form to aggregate 1.7.x or 2.0.5

#### Expected behavior
form should be pushed without error

#### Other information 
`.... Bad Gateway
Sending submission 23 of 40
Error sending submission 23 of 40: Erreur Interne de Servlet`

When debuging 

the method uncheckedExecute in CommonsHttp.java is adding the header ("X-OpenRosa-Version", "1.0") in headerGroup but this headerGroup is already sets by getPushSubmissionRequest in AgrgegateServer.java
This cause  an error in aggregate side when it  gets the Header's value. The value return by getHeader is now "1.0, 1.0" and not "1.0" 
`public String getHeader(String name); 

    /**
     * Returns all the values of the specified request header
     * as an <code>Enumeration</code> of <code>String</code> objects.`

This is understandable because header has 2 entries of "X-OpenRosa-Version" and getHeader concatenating them.

I hope, I am clear.

### LINKED ISSUE #857: Briefcase CLI for pushing update to form definition forces version update in the form definition even for simple media updates
Briefcase CLI for pushing update to form definition forces version update in the form definition even for simple media updates.

Briefcase CLI takes too long and is also finiky for Version update even when there is purely media update. For simple media updates, CLI works for few hours but after that it fails until I update the version of the form in form definition.
This constraint goes away when I upload the form directly into Aggregate GUI using .../formupload feature.

Need your feedback to get this underway.

thanks

#### Software versions 
Briefcase v1.x.x, Java v1.x.x, operating system, Aggregate v1.x.x, Collect v1.x.x...

#### Other information 
Things you tried, stack traces, related issues, suggestions on how to fix it...

yougis:
Closes #854

run on debug, build, create jar File

this is a quick hot fix
 
nothing worst. 

nope

### DIFF
--- a/src/org/opendatakit/briefcase/reused/http/CommonsHttp.java
+++ b/src/org/opendatakit/briefcase/reused/http/CommonsHttp.java
@@ -113,7 +113,9 @@ public class CommonsHttp implements Http {
 
     // Add the declared headers
     // TODO v2.0 remove this header, since this is not a concern of this class
-    commonsRequest.addHeader("X-OpenRosa-Version", "1.0");
+    if(!request.headers.containsKey("X-OpenRosa-Version")){
+      commonsRequest.addHeader("X-OpenRosa-Version", "1.0");
+    }
     request.headers.forEach(commonsRequest::addHeader);
 
     // Set the request's body if it's a POST request


ggalmazor:
Hi, @yougis!

Thanks for documenting this issue! You even tracked down the log in Aggregate! :O

I think that the change you're suggesting works, but I also suspect that the whole line from `CommonsHttp` could be removed, as the `TODO` comment suggests. Could you try that, please?

ggalmazor:
Hi again! It turns out I was wrong and there are some cases where everything relies on the CommonHttp class adding the header, so the PR is perfect as it is, pending further investigation. Ideally, each request would add all the required headers. The CommonsHttp class shouldn't know any specifics about OpenRosa.

yougis:
greate ! I am happy to get my (little) contribution merged !

kkrawczyk123:
**Verified with success!**
Verified on Ubuntu.

Tested scenarios:
- push form to Aggregate 2.0
- push form to Aggregate 1.7
- push to Central
- cmd push to Aggregate
- cmd push to Central

PR #858, opened 2020-03-24
Added Testing

elduwa:
As part of a software testing course in our university we are submitting this pull request.
Our first assignment was to try and increase test coverage (especially branch coverage for now) in an open source project.

We are aware that this contribution might not be as useful, but we would be grateful for some feedback :)

### DIFF
--- a/test/java/org/opendatakit/briefcase/model/FormStatusTest.java
+++ b/test/java/org/opendatakit/briefcase/model/FormStatusTest.java
@@ -5,6 +5,8 @@ import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
 
 import java.nio.file.Path;
+import java.util.Optional;
+import org.junit.Assert;
 import org.junit.Test;
 
 public class FormStatusTest {
@@ -43,4 +45,14 @@ public class FormStatusTest {
     );
   }
 
+  @Test
+  public void gets_manifest_url_from_remoteform() {
+    FormStatus form = new FormStatus(null);
+    String supposedManifestUrl = "manifest/test/url";
+    RemoteFormDefinition remoteForm = new RemoteFormDefinition("name", "1","v1",
+            "manifest/test/url", "/download/test/url");
+    Assert.assertEquals(supposedManifestUrl, remoteForm.getManifestUrl());
+    Assert.assertEquals(Optional.empty(), form.getManifestUrl());
+  }
+  
 }
--- a/test/java/org/opendatakit/briefcase/model/RetrieveAvailableFormsFailedEventTest.java
+++ b/test/java/org/opendatakit/briefcase/model/RetrieveAvailableFormsFailedEventTest.java
@@ -1,0 +1,21 @@
+package org.opendatakit.briefcase.model;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class RetrieveAvailableFormsFailedEventTest {
+
+    @Test
+    public void getReason_unknown() {
+        RetrieveAvailableFormsFailedEvent testEvent = new RetrieveAvailableFormsFailedEvent(null);
+        assertEquals("unknown", testEvent.getReason());
+    }
+
+    @Test
+    public void getReason_exception() {
+        Exception e = new Exception("fail");
+        RetrieveAvailableFormsFailedEvent testEvent = new RetrieveAvailableFormsFailedEvent(e);
+        assertEquals("Exception: fail", testEvent.getReason());
+    }
+}
--- a/test/java/org/opendatakit/briefcase/model/ServerConnectionInfoTest.java
+++ b/test/java/org/opendatakit/briefcase/model/ServerConnectionInfoTest.java
@@ -1,0 +1,34 @@
+package org.opendatakit.briefcase.model;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class ServerConnectionInfoTest {
+
+    @Test
+    public void testEquals_False() {
+        char[] testPassword = {'a'};
+        ServerConnectionInfo testInfo = new ServerConnectionInfo("testURL", "testName", testPassword);
+        Object o = new Object();
+
+        assertFalse(testInfo.equals(o));
+    }
+
+    @Test
+    public void testEquals_True() {
+        char[] testPassword = {'a'};
+        ServerConnectionInfo testInfo = new ServerConnectionInfo("testURL", "testName", testPassword);
+        ServerConnectionInfo testInfo2 = new ServerConnectionInfo("testURL", "testName", testPassword);
+
+        assertTrue(testInfo.equals(testInfo2));
+    }
+
+    @Test
+    public void testEquals_True_Itself() {
+        char[] testPassword = {'a'};
+        ServerConnectionInfo testInfo = new ServerConnectionInfo("testURL", "testName", testPassword);
+        assertTrue(testInfo.equals(testInfo));
+    }
+
+}
--- a/test/java/org/opendatakit/briefcase/model/form/FormKeyTest.java
+++ b/test/java/org/opendatakit/briefcase/model/form/FormKeyTest.java
@@ -2,6 +2,7 @@ package org.opendatakit.briefcase.model.form;
 
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
 import static org.opendatakit.briefcase.ui.pull.FormInstallerTest.getPath;
 import static org.opendatakit.briefcase.util.StringUtils.stripIllegalChars;
 
@@ -9,6 +10,7 @@ import java.io.IOException;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import org.junit.Assert;
 import org.junit.Test;
 import org.opendatakit.briefcase.model.BriefcaseFormDefinition;
 import org.opendatakit.briefcase.model.FormStatus;
@@ -26,5 +28,19 @@ public class FormKeyTest {
     FormKey key1 = FormKey.of("this-title-has-dashes/and/slashes", "this-id-has-dashes/and/slashes");
     FormKey key2 = FormKey.from(new FormStatus(BriefcaseFormDefinition.resolveAgainstBriefcaseDefn(formFile.toFile(), false, briefcaseFolder.toFile())));
     assertThat(key1, is(key2));
+  }
+
+  @Test
+  public void different_and_same_key_check_if_equal() {
+    FormKey key1 = FormKey.of("Key_1", "1", "v1.0");
+    FormKey key2 = FormKey.of("Key_1", "2", "v2.0");
+    FormKey key3 = FormKey.of("Key_1", "1", "v1.0");
+    String keyString = "key_4";
+    FormKey keyNull = null;
+    Assert.assertFalse(key1.equals(key2));
+    assertTrue(key1.equals(key1));
+    assertTrue(key1.equals(key3));
+    Assert.assertFalse(key1.equals(keyString));
+    Assert.assertFalse(key1.equals(keyNull));
   }
 }
--- a/test/java/org/opendatakit/briefcase/pull/aggregate/AggregateAttachmentTest.java
+++ b/test/java/org/opendatakit/briefcase/pull/aggregate/AggregateAttachmentTest.java
@@ -1,13 +1,15 @@
 package org.opendatakit.briefcase.pull.aggregate;
 
-import static org.junit.Assert.assertThat;
-
 import java.io.IOException;
+import java.net.MalformedURLException;
 import java.net.URISyntaxException;
+import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import org.hamcrest.Matchers;
 import org.junit.Test;
+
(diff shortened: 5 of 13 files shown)


test/java/org/opendatakit/briefcase/model/form/FormKeyTest.java
```diff
@@ -27,4 +29,18 @@ public void regression_wrong_form_name_when_creating_keys_from_briefcase_form_de
     FormKey key2 = FormKey.from(new FormStatus(BriefcaseFormDefinition.resolveAgainstBriefcaseDefn(formFile.toFile(), false, briefcaseFolder.toFile())));
     assertThat(key1, is(key2));
   }
+
+  @Test
+  public void different_and_same_key_check_if_equal() {
```
ggalmazor, line 34: Here are my thoughts about this test:
- `FormKey` is a value object, so we should assert whether an instance **is** or **is not** another instance. 
  Language, in this case, is important because in Java, being equal to something and "is"-ing  that something implies different equality rules.
  In this case we want to verify that objects are effectively the same, as if they were primitive values.
- I wouldn't bother asserting the `null` case
- The mixed types case is formally wrong because you're only matching it against a String, which doesn't probe it won't be equal to any other infinite types that you could use for that assertion. I'd remove that assertion.
- Also, in this codebase we use Hamcrest matchers with `assertThat`. it's important to follow conventions and coding standards. In this case, I'd expect to have `assertThat` assertions instead of `assertFalse` or `assertTrue`.
- Another remark about coding is the mix of statically imported `assertTrue` versus the fully qualified `Assert.assertFalse`. I'd expect everything statically imported, or fully qualified. Not a mix.

test/java/org/opendatakit/briefcase/model/FormStatusTest.java
```diff
@@ -43,4 +45,14 @@ public void knows_how_to_build_paths_to_assets_inside_the_storage_directory() {
     );
   }
 
+  @Test
+  public void gets_manifest_url_from_remoteform() {
+    FormStatus form = new FormStatus(null);
+    String supposedManifestUrl = "manifest/test/url";
```
ggalmazor, line 51: Small language tip: Use `expected` instead of `supposed` as in `expectedManifestUrl`.

test/java/org/opendatakit/briefcase/model/FormStatusTest.java
```diff
@@ -43,4 +45,14 @@ public void knows_how_to_build_paths_to_assets_inside_the_storage_directory() {
     );
   }
 
+  @Test
+  public void gets_manifest_url_from_remoteform() {
```
ggalmazor, line 49: This test feels like it's not bringing any value. It doesn't help to describe this the class' behavior with a meaningful example. It doesn't helping prevent bugs either because it only asserts about its initial state.

Test coverage is not a goal by itself, nor it shoule be considered a metrics to get a sense of the quality of a codebase.

Code, on the other hand, is inventory and has inherent cost. That's why I would suggest not writing this test.

test/java/org/opendatakit/briefcase/pull/aggregate/AggregateAttachmentTest.java
```diff
@@ -1,19 +1,39 @@
...
 public class AggregateAttachmentTest {
   @Test
   public void computes_the_string_MD5_hash_of_a_file() throws URISyntaxException, IOException {
     Path file = Paths.get(AggregateAttachmentTest.class.getResource("/org/opendatakit/briefcase/pull/aggregate/lorem-ipsum-40k.txt").toURI());
     String expectedHash = "E79C0D4AD451003BA8CFDC1183AC89E9";
     assertThat(AggregateAttachment.md5(file), Matchers.is(expectedHash));
   }
+
+  @Test
+  public void gets_download_url() throws URISyntaxException, MalformedURLException {
```
ggalmazor, line 23:
```suggestion
  public void gets_download_url() throws MalformedURLException {
```
`UriSyntaxException` is not being thrown by this block

ggalmazor, line 23: This test is not bringing much to the table. The tested method is being tested indirectly for other tests in higher levels of abstraction. I'd suggest removing it.

test/java/org/opendatakit/briefcase/pull/aggregate/AggregateAttachmentTest.java
```diff
@@ -1,19 +1,39 @@
...
+  @Test
+  public void checks_equality_of_attachment() throws URISyntaxException, MalformedURLException {
+    AggregateAttachment testAttachment = AggregateAttachment.of("file_1", "123456", "file://org/opendatakit/briefcase/pull/aggregate/lorem-ipsum-40k.txt");
+    AggregateAttachment testAttachment2 = AggregateAttachment.of("file_1", "123456", "file://org/opendatakit/briefcase/pull/aggregate/lorem-ipsum-40k.txt");
+    AggregateAttachment testAttachment3 = AggregateAttachment.of("file_3", "1234567", "file://org/opendatakit/briefcase/lorem-ipsum-40k.txt");
+    assertFalse(testAttachment.equals(null));
+    assertTrue(testAttachment.equals(testAttachment));
+    assertTrue(testAttachment.equals(testAttachment2));
+    assertFalse(testAttachment.equals(testAttachment3));
+  }
```
ggalmazor, line 38:
```suggestion
  @Test
  public void checks_equality_of_attachment() throws URISyntaxException, MalformedURLException {
    AggregateAttachment attachment1 = AggregateAttachment.of("111112.txt", "111111", "http://foo.bar/111111.txt");
    AggregateAttachment attachment1Clone = AggregateAttachment.of("111112.txt", "111111", "http://foo.bar/111111.txt");
    AggregateAttachment attachment2 = AggregateAttachment.of("222222.txt", "222222", "http://foo.bar/222222.txt");
    assertThat(attachment1, is(attachment1Clone));
    assertThat(attachment1, is(not(attachment2)));
  }
```

ggalmazor: This alternative is has less text to parse, which makes it easier to understand. It also uses consistent numbers and names , and enough assertions to explain what's the behavior we're asserting

test/java/org/opendatakit/briefcase/model/RetrieveAvailableFormsFailedEventTest.java
```diff
@@ -0,0 +1,21 @@
+package org.opendatakit.briefcase.model;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class RetrieveAvailableFormsFailedEventTest {
```
ggalmazor, line 7: Testing getters here, which is nor very useful. I'd suggest removing this. test class

test/java/org/opendatakit/briefcase/model/ServerConnectionInfoTest.java
```diff
@@ -0,0 +1,34 @@
+package org.opendatakit.briefcase.model;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class ServerConnectionInfoTest {
```
ggalmazor, line 7: This tests are only increasing the coverage percentage, which is not useful by itself. This test class doesn't bring much value to the table because equality rules of ServerConnectionInfo are indirectly verified by higher-level of abstraction tests elsewhere. I'd suggest removing this test class

test/java/org/opendatakit/briefcase/reused/http/CommonsHttpTest.java
```diff
@@ -156,4 +157,12 @@ public void can_handle_3xx_errors() throws Exception {
       assertThat(response.isSuccess(), is(false));
     });
   }
+
+  @Test
+  public void the_factories_accept_max_http_connections_with_Proxy(){
```
ggalmazor, line 162: I think the test method name has to be improved. This test verifies that the returned object is of a certain type, which doesn't have to do with the method name. It's very important that the test methods convey with precission what's being verified.

test/java/org/opendatakit/briefcase/reused/http/CommonsHttpTest.java
```diff
@@ -156,4 +157,12 @@ public void can_handle_3xx_errors() throws Exception {
...
+    assertEquals(CommonsHttp.class, CommonsHttp.of(MAX_HTTP_CONNECTIONS, null).getClass());
```
ggalmazor, line 163: Also, this verification by itself is not very useful.

test/java/org/opendatakit/briefcase/reused/http/RequestBuilderTest.java
```diff
@@ -86,4 +89,11 @@ public void can_resolve_paths() {
         is(url("http://foo.com/bar/baz"))
     );
   }
+
+  @Test
+  public void can_resolve_body(){
```
ggalmazor, line 94: This test doesn't belong to this test class because what we're verifying is that the `RequestSpy` class can read a request body. You don't need a `Request` object for that.

test/java/org/opendatakit/briefcase/transfer/TransferFormsTest.java
```diff
@@ -16,14 +16,34 @@
...
+import org.junit.Before;
 import org.junit.Test;
+import org.opendatakit.briefcase.model.FormStatus;
 import org.opendatakit.briefcase.model.FormStatusBuilder;
+import org.opendatakit.briefcase.model.TestFormDefinition;
 
 public class TransferFormsTest {
+
+  private List<FormStatus> forms;
+  private Path briefcaseDir;
```
ggalmazor, line 38: Not used, remove.

test/java/org/opendatakit/briefcase/transfer/TransferFormsTest.java
```diff
@@ -16,14 +16,34 @@
...
+
+  @Before
+  public void setUp() {
+    briefcaseDir = get("/storage/directory");
```
ggalmazor, line 42: Not used, remove.

test/java/org/opendatakit/briefcase/transfer/TransferFormsTest.java
ggalmazor, line 38:
```suggestion
```
This property is not used. Remove

test/java/org/opendatakit/briefcase/transfer/TransferFormsTest.java
ggalmazor, line 42:
```suggestion
```
Not used. Remove

test/java/org/opendatakit/briefcase/transfer/TransferFormsTest.java
```diff
@@ -42,4 +62,27 @@ public void forms_from_varargs_factory_can_be_cleared() {
     forms.clear();
     assertThat(forms.isEmpty(), is(true));
   }
+
+  @Test
+  public void loads_all_forms() {
+    TransferForms transferForms = TransferForms.empty();
+    transferForms.load(forms);
+    assertEquals(10, transferForms.size());
```
ggalmazor, line 70: Why not test that the event `onChange` is triggered when forms are loaded, which is also part of this method's behavior?

test/java/org/opendatakit/briefcase/transfer/TransferFormsTest.java
```diff
@@ -42,4 +62,27 @@ public void forms_from_varargs_factory_can_be_cleared() {
...
+    assertTrue(transferForms.allSelected());
+    assertTrue(transferForms.someSelected());
+  }
+
+  @Test
+  public void get_selected_forms() {
+    TransferForms transferForms = TransferForms.from(forms);
+    transferForms.selectAll();
+    transferForms.getSelectedForms();
+    assertEquals(10, transferForms.size());
```
ggalmazor, line 86:
```suggestion
    TransferForms selectedForms = transferForms.getSelectedForms();
    assertEquals(10, selectedForms.size());
```
This test wasn't doing what you wanted because you weren't using the output of `getSelectedForms()` in the assertion.

test/java/org/opendatakit/briefcase/util/BriefcaseVersionManagerTest.java
```diff
@@ -57,4 +57,18 @@ public void knows_if_we_are_not_up_to_date() {
 
     assertThat(versionManager.isUpToDate(), is(false));
   }
+
+  @Test
+  public void version_contains_hyphen(){
```
ggalmazor, line 62: This test is useful. However, I think the test method requires a bit of work. Maybe something along the lines "ignores beta version information"

test/java/org/opendatakit/briefcase/util/XmlManipulationUtilsTest.java
```diff
@@ -0,0 +1,48 @@
...
+    model.addChild(Node.ELEMENT, instance);
+    instance.addChild(Node.ELEMENT, data);
+    data.addChild(Node.ELEMENT, meta);
+    meta.addChild(Node.ELEMENT, instanceId);
+  }
+
+  @Test
+  public void find_meta_tag_successful() {
+    FormInstanceMetadata metadata = getFormInstanceMetadata(tree);
+    assertNotNull(metadata);
```
ggalmazor, line 45: I'm sure you can come up with more interesting assertions than just asserting that the output is not null!

PR #870, opened 2020-07-04
Push Central attachments to draft endpoint and publish draft

### LINKED ISSUE #860: Central v0.8: media can't be pushed and pushing submissions requires manually publishing form on Central
[Central v0.8](https://github.com/opendatakit/central/issues/113) introduces form definition drafts and versions.

Media attachments can only be pushed to draft form definitions, not to published ones. This means that Briefcase currently can't push media because it attempts to push to the published version.

Relatedly, when Briefcase pushes a form definition to Central, it remains a draft on Central. That means submissions can't be pushed to it. After a form definition is pushed to Central, it should be published by Briefcase.

lognaturel:
Closes #860

Updated the relevant test and did manual testing against https://sandbox.central.opendatakit.org/.

Initially I thought I'd try to support all versions of Central but I decided that it's not worth it. It's relatively easy to upgrade Central and users really shouldn't be running on old versions. As far as we know there's no behavior that was deprecated that would require users to stay on pre-0.8. I did try to give a useful error message in case of trying to push to an older Central so users can recover.

In terms of code structure, I tried to follow existing patterns as much as possible.

The first commit isn't entirely related but it came up because I was going to lower the level of some errors that are showing up in Sentry but then I realized that they weren't intended to show up in Sentry and are only doing so because the last release used the old release logback file. I made sure @yanokwa's release project removes this and updated the readme.

The only intentional change is to fix the bug. The only code touched is push to Central and that operation is at risk of regressions.

No.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ There are example configuration files that you can use while developing:
 - Copy `res/logback.xml.example` to `res/logback.xml`. This conf will be used when launching Briefcase on your machine.
 
 ### Logging tests vs development vs release
-During the release process, we use a specific logback for release. This configuration sends exceptions to Sentry.io and also logs to a `briefcase.log` file, created in the same folder where Briefcase is launched by the user.
+During the release process, we use a specific logback configuration which logs to a `briefcase.log` file in the folder Briefcase is launched from.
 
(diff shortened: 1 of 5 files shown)


lognaturel:
That's totally right, @ggalmazor, thanks. I'm going to merge since I'm starting to get conflicts with a spike for pushing multiple form versions to Central and I'll file an issue to add the test.

PR #872, opened 2020-07-18
When pushing to Central, push form definitions for all versions referenced in submissions

lognaturel:
Central only accepts submissions for form versions that it is aware of. Aggregate allows limited form versioning but only retains the latest form version. Briefcase only has awareness of the last form version it pulled.

There was a v2 planned to address this and other limitations but unfortunately we've been struggling to get the time and resources to make it happen. 

TODO:
- [x] Canceling push should not give Success status
- [x] Investigate key issue -- is it from pulling new form definition version?
- [x] When pulling Kasia's original All Widgets form with v1.17.4, are there blank submission directories? Seems this was related to the server running out of memory
- [x] Attempt to push submission attachments should not result in 404 - Seems this was also related to server running out of memory
- [x] Look into progress bar starting after warning dialog dismissed
- [x] Seems cancel button on warning dialog is totally broken
- [ ] Look into empty/missing form versions. May need updates to Central API to handle fully. Possibly file as separate issue once merged.
- [x] Write docs page - https://github.com/getodk/docs/pull/1272
- [ ] File issue for making docs link in warning dialog clickable
- [ ] In 1.17.4, pull skips media attachments if the submission exists. That doesn't seem right. It means it can be really hard to know a dataset is incomplete.

This PR:
- reads versions from submissions as it pulls them from Aggregate, Central, or a Collect directory and stores a set of all identified versions in the form metadata
- when pushing to Central from the UI, warns the user that the same form definition will be pushed to represent all the versions needed by submissions
- when pushing to Central, for each identified version from submissions, pushes the available form definition as a draft and then publishes it with the version identifier
- when pushing to Central, pushes and publishes the latest form definition last
- only uses formId in metadata lookup instead of formId and version. The implication of using the formId, version pair was that metadata would be lost each time a new version was pulled (e.g. export status or submission versions)
- does not add any kind of warning for the command line push to Central. It's a less common interface and its users are more savvy.

Acceptance criteria:
* Given: I have pulled a form definition and submissions from Aggregate, Central, or a Collect directory from this version of Briefcase
  And: the submissions are all for the active form definition's version
  When: I push that form and submissions to Central
  Then: the form definition and corresponding submissions are pushed without any special behavior

* Given: I have pulled a form definition and submissions from Aggregate, Central, or a Collect directory from this version of Briefcase
  And: some submissions are for the active form definition's version and others for prior form versions
  When: I push that form and submissions to Central
  Then: I see a warning with a description of what will happen, a reference to a documentation page
[...]

PR #879, opened 2020-10-12
Resolve all jr:// URIs to media directory

### LINKED ISSUE #878: Forms with external secondary instances aren't shown on export or push tabs
#### Software versions
All

#### Problem description
Forms with external secondary instances fail to parse so don't show up on export or push tabs.

1. Pull a form with external secondary instance such as [this one](https://test.central.getodk.org/#/projects/149/forms/a) (requires QA server credentials)
1. See that it's not in export or push tabs

#### Expected behavior
Forms with external secondary instances should act the same as any other form.

#### Other information
I think this is a matter of correctly setting the reference manager and/or the root translators.

lognaturel:
Needed to build a FormDef for files with external secondary instances.

Closes #878

Wrote a test that fails with the user-reported error on master. Tried downloading a real form with external secondary instances and then exporting it. @getodk/testers I think it would be good to try with some of the other ones that you have with both xml and csv external instances and a few submissions.

The core of the solution is setting up a `ReferenceFactory` for the singleton `ReferenceManager`. I don't see any alternatives to that. The decision points I saw were around where to actually build the media directory paths and how much of the `ReferenceFactory` to implement. It felt reasonable to make `getMediaDirectory` public and `JavaRosaWrapper` and `BriefcaseFormDefinition` call it.

For the `ReferenceFactory`, I only implemented `getLocalURI` because I believe that's all we need. The other methods throw `UnsupportedOperationException` so we should be able to tell easily if I was wrong. I had all media of every type resolve to the same directory and don't do any check on the `jr://` prefix provided because I consider that a form design concern.

I changed the implementation of `getMediaDirectory` to not create the directory if it didn't exist but just get the path. The only risk I could think of associated with that was when media attachments are present so I verified that. I could have missed something. Other than that the changes are additive so low-risk.

No.

### DIFF
--- a/src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
+++ b/src/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
@@ -17,6 +17,7 @@
 
 package org.opendatakit.aggregate.parser;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
 import java.io.StringReader;
@@ -32,6 +33,7 @@ import org.javarosa.core.model.instance.AbstractTreeElement;
 import org.javarosa.core.model.instance.FormInstance;
 import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.model.instance.TreeReference;
+import org.javarosa.core.reference.ReferenceManager;
 import org.javarosa.core.services.PrototypeManager;
 import org.javarosa.model.xform.XFormsModule;
 import org.javarosa.xform.parse.XFormParser;
@@ -342,7 +344,7 @@ public class BaseFormParserForJavaRosa implements Serializable {
    *
    * @throws ODKIncompleteSubmissionData
    */
-  protected BaseFormParserForJavaRosa(String existingXml, String existingTitle, boolean allowLegacy)
+  protected BaseFormParserForJavaRosa(String existingXml, String existingTitle, File mediaDirectory, boolean allowLegacy)
       throws ODKIncompleteSubmissionData {
     if (existingXml == null) {
       throw new ODKIncompleteSubmissionData(Reason.MISSING_XML);
@@ -351,6 +353,8 @@ public class BaseFormParserForJavaRosa implements Serializable {
     xml = existingXml;
 
     initializeJavaRosa();
+    ReferenceManager.instance().reset();
+    ReferenceManager.instance().addReferenceFactory(new MediaFileReferenceFactory(mediaDirectory));
 
     XFormParserWithBindEnhancements xfp = parseFormDefinition(xml, this);
     try {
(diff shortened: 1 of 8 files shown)


lognaturel:
I don't understand why the test fails on CI. It passes locally. CI says 

>Unable to parse external secondary instance: java.io.FileNotFoundException: /tmp/briefcase_test14560796627712759977/briefcase_test14560796627712759977-media/external-xml.xml (No such file or directory)```

That is indeed the path structure I would expect.

lognaturel:
I SSHed into the build and saw `/tmp/briefcase_test16513623838698692582/forms/nested-repeats/nested-repeats-media/external-xml.xml (No such file or directory)` which led me to see I was building the media URI wrong. Still, I don't know how we could have ended up with that path. I noticed that there is lots of test data left in `/tmp` which suggests cleanup is incomplete for some tests.

@ggalmazor I'd love it if you could take a quick peek when you have a moment. Hopefully this one will be very quick.

lognaturel:
Now I have no idea why the test is failing. 😭

ggalmazor:
Hi, @lognaturel! I'd need to wait until the weekend to take a look. Does that work for you?

lognaturel:
That works, @ggalmazor!

kkrawczyk123:
**Tested with success!**
Verified on Ubuntu.
Tested scenarios:
- pull form Central
- export
- pull with cmd
- export with cmd
- unable to push to Central manually and using cmd both: what is probably connected to xml media file and we already encountered that issue in #872 that should be fixed now. Error message displayed:
```
Form with external secondary instance - Start pushing form and submissions
Form with external secondary instance - Can't determine form encryption:Unable to parse external secondary instance: org.javarosa.core.reference.InvalidReferenceException: The reference "jr://file/towns.xml" was invalid and couldn't be understood. The javarosa jr:// reference root "file" is not available on this system and may have been mis-typed. Some available roots: 
    Problem found at nodeset: /html/head/model/instance
    With element <instance id="towns" src="jr://file/towns.xml"/>
Form with external secondary instance - Success
All operations completed
```
- download on Collect
- send submissions to Central
- pull before export doesn' work (reported in #882 as it's also visible on 1.17.4)
- form with external csv file
- xlsx form with external csv 
- form with external csv and xml: what is actually weird I was able to push that form https://test.central.getodk.org/#/projects/214/forms/formWithExternalFiles to central despite of xml media attachment. 
cc @lognaturel

mmarciniak90:
Verified on Mac. The main issue is no longer visible.
I didn't notice more unexpected issues than described by Kasia.

lognaturel:
> what is actually weird I was able to push that form test.central.getodk.org/#/projects/214/forms/formWithExternalFiles to central despite of xml media attachment

That seems good and expected to me! Are you thinking about the issue you identified in https://github.com/getodk/briefcase/pull/872#issuecomment-709371710? That only applies to XML files that are attached to instances, not to form definitions.

PR #881, opened 2020-10-14
URI-encode spaces in URLs

### LINKED ISSUE #880: Crash when pulling forms with spaces in formId from Central
#### Software versions
Briefcase v1.17.4

#### Problem description
Crash when pulling forms with spaces in formId from Central

1. Put a form on Central that has a space in the form ID
1. Try to pull that form and see a crash

e.g. [this one](https://test.central.getodk.org/#/projects/149/forms/formid%20with%20spaces) (requires QA server access)

#### Expected behavior
No crash.

lognaturel:
Closes #880

Ran it with the form from the issue, wrote an automated test. I also pushed that same form and exported it to confirm there were no other issues with a form id with a space.

I initially thought I could use some kind of Java API for this but I don't think so. See https://stackoverflow.com/questions/4737841/urlencoder-not-able-to-translate-space-character. I decided to be very surgical and only handle spaces for now and we can deal with other characters that need to be URI-encoded if they actually come up.

The only thing this does is replace spaces in URIs with `%20`. It should be very safe.

No.

### DIFF
--- a/src/org/opendatakit/briefcase/reused/http/Request.java
+++ b/src/org/opendatakit/briefcase/reused/http/Request.java
@@ -83,7 +83,7 @@ public class Request<T> {
   // TODO v2.0 Move this to RequestBuilder, with uri() and isUri()
   URI asUri() {
     try {
-      return url.toURI();
+      return new URI(url.toString().replaceAll(" ", "%20"));
     } catch (URISyntaxException e) {
       throw new BriefcaseException(e);
     }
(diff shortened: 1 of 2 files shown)


kkrawczyk123:
@lognaturel pulling form with spaces in id doesn't work via cmd, error message `Error: Form formid not found` displays so only the first word in that kind of id is considered when using cmd. 
other actions that has been verified and works well for that kind of form:
- pull from central
- push to Central
- export 
- push to aggregate
- download on Collect
- fill submission sent to Central

lognaturel:
> pulling form with spaces in id doesn't work via cmd

This has more to do with how the command line works in general, I think. That is, spaces are meaningful separators between so they must be escaped in some way if they're part of option text. On zsh, I can successfully pull the form from the issue by either escaping spaces with `\` (`formid\ with\ spaces`) or wrapping the id in single quotes (`'formid with spaces'`). I think those are pretty standard. My sense is that this is fine.

kkrawczyk123:
Oh yes, I should think about that. I was able to pull, push and export using cmd right now. So we have success here!

PR #884, opened 2020-10-22
On export, skip submissions that can't be parsed

### LINKED ISSUE #874: Crash on export when submission is empty
#### Software versions
Briefcase v1.17.4

#### Problem description
When a submission is empty (or presumably when one is corrupt), the whole export crashes with no real hints as to what is going on.

1. Start a pull operation and cancel it. Verify that the last submission started is empty
1. Start an export
1. See that export files only have headers. Export details stop after some percentage.
1. Debug to see crash in `parseAttribute` 

#### Expected behavior
Canceling should probably not leave an empty submission.
If there is an empty submission, that should be logged, the submission should be skipped, and export should continue. There is a `failed_submission` file generated, but it's empty so it doesn't really help.

lognaturel:
Closes #874

Tried different exports with and without empty submission.xml files. Clearing the contents of a submission.xml file reproduces the original issue. With this PR, the export should be successful, there should be a message in the details saying `Can't export submission ...`, the `-errors` folder should include the failed submission with its instanceID.

This is the only solution I considered once I realized that submissions that were leading to parse errors were included in the submissions to export. This is the simplest way to exclude those submissions.

Users should get fewer export errors and should have more information to recover from them when they do happen. The regression risk is low because it only affects the path where there's an exception on parse.

No.

### DIFF
--- a/src/org/opendatakit/briefcase/export/ExportToCsv.java
+++ b/src/org/opendatakit/briefcase/export/ExportToCsv.java
@@ -37,7 +37,6 @@ import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.atomic.AtomicInteger;
 import org.bushe.swing.event.EventBus;
 import org.opendatakit.briefcase.model.FormStatus;
 import org.opendatakit.briefcase.model.form.FormMetadata;
@@ -143,14 +142,13 @@ public class ExportToCsv {
   }
 
   private static SubmissionExportErrorCallback buildParsingErrorCallback(Path errorsDir) {
-    AtomicInteger errorSeq = new AtomicInteger(1);
     // Remove errors from a previous export attempt
     if (exists(errorsDir))
       deleteRecursive(errorsDir);
     return (path, message) -> {
       if (!exists(errorsDir))
         createDirectories(errorsDir);
-      copy(path, errorsDir.resolve("failed_submission_" + errorSeq.getAndIncrement() + ".xml"));
+      copy(path, errorsDir.resolve("failed_submission_" + path.getParent().getFileName() + ".xml"));
       log.warn("A submission has been excluded from the export output due to some problem ({}). If you didn't expect this, please ask for support at https://forum.getodk.org/c/support", message);
     };
   }
--- a/src/org/opendatakit/briefcase/export/SubmissionParser.java
+++ b/src/org/opendatakit/briefcase/export/SubmissionParser.java
@@ -86,11 +86,12 @@ public class SubmissionParser {
         .forEach(instanceDir -> {
           Path submissionFile = instanceDir.resolve("submission.xml");
           try {
-            Optional<OffsetDateTime> submissionDate = readSubmissionDate(submissionFile, onParsingError);
+            Optional<OffsetDateTime> submissionDate = readSubmissionDate(submissionFile);
             paths.add(Pair.of(submissionFile, submissionDate.orElse(OffsetDateTime.of(1970, 1, 1, 0, 0, 0, 0, ZoneOffset.UTC))));
           } catch (Throwable t) {
-            log.error("Can't read submission date", t);
+            log.error("Parse error attempting to read instance date", t);
             EventBus.publish(ExportEvent.failureSubmission(formDef, instanceDir.getFileName().toString(), t));
+            onParsingError.accept(submissionFile, "Parse error attempting to read instance date");
           }
         });
     return paths.parallelStream()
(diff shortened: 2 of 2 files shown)


lognaturel:
@getodk/testers I've rebased on master, please take a look.

kkrawczyk123:
@lognaturel After long investigation we finally figure out the problem and agreed that the export crash issue is visible on PR's build only on java 8. I was also able to reproduce it (I am normally on java 11 this is why it wasn't visible for me previously), see the gif:
Peek 2020-10-28 11-54

The issue is not visible on java 9, 10 and 11. I am not sure which one is the most frequent for users. As the issue is visible only on the old java version it shouldn't block this or. It has been verified on Ubuntu and MacOS.

Here comes a question on which java version we should do regression pass?
In the last release, I was using java 11 for building and testing. There are also cases to run briefcase on java 8, 9 and 10 and perform basic actions like pull, push and export. 

lognaturel:
We build releases with Java 11 and we don't typically support users making their own builds. So if it's the **build** version that makes a difference, I don't think it matters. If it's the **run** version that makes a difference, then I think we should look into it more deeply.

Java 11 from https://adoptopenjdk.net/ is what we recommend in the docs for running Briefcase so let's QA with that. Still, it should run with Java 8.

kkrawczyk123:
It's a run version issue, jar has been build on java 11.

lognaturel:
Got it, so I will look into it. It's specifically linked to this PR? And the export just hangs?

lognaturel:
I'll merge and then investigate on master. I'm guessing that somewhere there's an API not available on Java 8 that's being used but I don't think it's in this PR specifically.

kkrawczyk123:
It works better on 1.17.4 where export is starting, % progress is displayed and csv is created but it's empty and shows exactly the issue that has been fixed here.

lognaturel:
I just merged, pulled from master, built with java 11, ran with java oracle64-1.8.0.202, and everything worked as expected. If you're seeing a crash, do you get a stack trace on the console?

kkrawczyk123:
Actually what is even weirder nothing meaningful seems to show in log: [briefcase .log.txt](https://github.com/getodk/briefcase/files/5453556/briefcase.log.txt)
I wasn't able to reproduce for Central, Aggregate only.

mmarciniak90:
[console.txt](https://github.com/getodk/briefcase/files/5458104/console.txt)
export

lognaturel:
@mmarciniak90 the server source makes no difference, right? You have that strange thing where forms aren’t immediately showing up in push or export after a pull. Is that also the case on v1.17.4? And can you confirm everything works as expected when you use Java 11?

mmarciniak90:
> the server source makes no difference, right?

That's right. I was able to reproduce it on the Aggregate server and Central, as you can see.

> You have that strange thing where forms aren’t immediately showing up in push or export after a pull. Is that also the case on v1.17.4?

Yes, when I'm using Java 8.

> And can you confirm everything works as expected when you use Java 11?

When I use Java 11 I don't see problem where forms aren’t immediately showing up in push or export after a pull.

lognaturel:
>  I don't see problem where forms aren’t immediately showing up in push or export after a pull.

Great. And what about on 1.17.4? Does export work as expected with Java 11?

Another though for Java 8. What if you pull, close Briefcase, reopen Briefcase then try the export?

mmarciniak90:
> And what about on 1.17.4? Do the forms show up immediately?
>Does export work as expected with Java 11?

When I use Java 11 and Briefcase with these changes,
* forms are immediately showing up in push or export after a pull
* no crash on export when submission is empty
* details status shows correct information
 Screenshot from 2020-10-29 16-46-01

When I use Java 11 and Briefcase 1.17.4
* forms are immediately showing up in push or export after a pull
* crash on export when submission is empty
   * Caused by: java.lang.RuntimeException: Document has no root element!

> What if you pull, close Briefcase, reopen Briefcase then try the export?

When I use Java 8 and Briefcase with these changes,
* forms aren’t immediately showing up in push or export after a pull (also visible on 1.17.4)
* closing Briefcase and reopening it doesn't cause that forms are showing up in push or export. Only reloading the storage directory fixes this issue.

When I use Java 8 and Briefcase 1.17.4
* forms aren’t immediately showing up in push or export after a pull (also visible on 1.17.4)
* closing Briefcase and reopening it doesn't cause that forms are showing up in push or export. Only reloading the storage directory fixes this issue.

lognaturel:
Great, so we don't worry about forms showing up in the other tabs since that's an ongoing issue.

> closing Briefcase and reopening it doesn't cause that forms are showing up in push or export. Only reloading the storage directory fixes this issue.

And when you reload after re-opening on this PR, does the export work or does it still hang/crash? Does it hang for all forms or only forms with empty submissions?

mmarciniak90:
Yeah!! We have a workaround! :tada: 
Reopend and reload storage directory cause that export is possible with Java 8. 
Finally, I have the expected status.
Screenshot from 2020-10-29 17-19-58

lognaturel:
Phew! Ok, so the problem you experience is that on Java 8, *if there is an empty submission*, export fails unless Briefcase is closed and reopened? Export works in other cases, right? I think that means we're no worse off than with v1.17.4 so we can file this as an issue and deal with it if users run into trouble.

mmarciniak90:
I completely agree!

PR #887, opened 2020-10-27
When pulling from Central, download submission or attachment files that don't exist

### LINKED ISSUE #886: On pull from Central, submission attachments are not requested if submission folder exists
#### Software versions
Briefcase v1.17.4

#### Problem description
Pulling a submission is not an atomic action so it's possible to only get a subset of media attachments or even not get the submission XML. This makes it really important to pull all files on subsequent attempts.

1. Pull a form with media
1. Cancel while some media attachments for a submission are being downloaded
1. Confirm that the submission directory only has partial media
1. Attempt a new pull, let it succeed
1. Confirm that the submission directory still only has partial media

#### Expected behavior
Either pull needs to be atomic for a submission or pulling again needs to check whether media files already exist before skipping them.

lognaturel:
Closes #886

Prior to this change, submissions whose instanceIDs were in the instance database were filtered out when doing a pull from Central. The problem is that presence in the db does not mean any file was successfully saved for that instance. One change I made to address this is to not save an instanceID to the database if the pull task has been cancelled. **I made this change to pull from Aggregate as well.** What this should mean is that if a pull is cancelled, the submission(s) that are being downloaded at the time should not be skipped on a subsequent pull.

For Central, I also changed when the database is checked for an instance. Prior to this change, if an instanceID were in the database, the submission would be skipped entirely. Now, if the submission is in the database, we:
- check for the presence of a submission file on disk and download it if it doesn't exist
- download the attachment list
- for each attachment, check if it exists on disk and download it if not

This means that there's going to be an extra disk check and an extra request for every single submission to get the attachment list. As a result, subsequent pulls will be slower, even for forms with no attachments (we still need to make the request to know there are no expected attachments). I think the tradeoff is worth it to ensure getting all media files in case of "success with errors".

I did **not** make the corresponding change for Aggregate. This is because there is no API endpoint to get a submission attachment listing. Getting it requires pulling the submission contents as well. We know we have Aggregate users for whom performance is quite important so the tradeoff did not seem worth it in this case. This means that when there's a "success with errors" case in Aggregate, some media could be missing even after a subsequent pull.

Pulled from https://test.central.getodk.org/#/projects/149/forms/build_media_1556624635/submissions, cancelled, verified some media was missing, pulled again, verified the media was present. I also manually deleted all contents of an instance's directory, just the submission file, just an attachment and confirmed that the missing files were downloaded on a subsequent pull.

See discussion of tradeoffs above. I also changed the logging messages for submission attachments a little bit to visually link them to their submission and to provide a submission filename and instanceID in case of failure so those are easier to trace.

The biggest risk is that subsequent pulls from Central could be too slow to be practical in certain contexts. I don't think that's the case but it would depend on connection latency to a specific server and disk speed.

I did have to make a number of changes to the pull from Central flow so there's regression risk there. There should be no risk to pull from Aggregate because the only change I made was an added check on whether the pull was cancelled.

No.

### DIFF
--- a/src/org/opendatakit/briefcase/pull/aggregate/PullFromAggregate.java
+++ b/src/org/opendatakit/briefcase/pull/aggregate/PullFromAggregate.java
@@ -149,7 +149,9 @@ public class PullFromAggregate {
                   submissionAttachments.parallelStream().forEach(attachment ->
                       downloadSubmissionAttachment(form, submission, attachment, rs, tracker, currentSubmissionNumber, totalSubmissions, submissionAttachmentNumber.getAndIncrement(), totalSubmissionAttachments)
                   );
-                  db.putRecordedInstanceDirectory(submission.getInstanceId(), form.getSubmissionDir(briefcaseDir, submission.getInstanceId()).toFile());
+                  if (!rs.isCancelled()) {
+                    db.putRecordedInstanceDirectory(submission.getInstanceId(), form.getSubmissionDir(briefcaseDir, submission.getInstanceId()).toFile());
+                  }
                 });
           });
 
--- a/src/org/opendatakit/briefcase/pull/central/PullFromCentral.java
+++ b/src/org/opendatakit/briefcase/pull/central/PullFromCentral.java
@@ -40,7 +40,7 @@ import org.opendatakit.briefcase.model.FormStatusEvent;
 import org.opendatakit.briefcase.model.form.FormKey;
 import org.opendatakit.briefcase.model.form.FormMetadataPort;
 import org.opendatakit.briefcase.pull.PullEvent;
-import org.opendatakit.briefcase.reused.Triple;
+import org.opendatakit.briefcase.reused.Pair;
 import org.opendatakit.briefcase.reused.http.Http;
 import org.opendatakit.briefcase.reused.http.response.Response;
 import org.opendatakit.briefcase.reused.job.Job;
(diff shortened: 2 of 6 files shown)


mmarciniak90:
@lognaturel I noticed that the second pull overwrites previously downloaded media. I observe that the modified date is changed. The funniest thing is that only I see this behavior on macOS and Ubuntu. @kkrawczyk123 didn't observe it.

lognaturel:
Thanks, @mmarciniak90. That's definitely not supposed to happen.

mmarciniak90:
I hope you can reproduce it :crossed_fingers:

lognaturel:
@mmarciniak90 I added some automated tests to verify that if a `submission.xml` file or submission attachment file exists after a pull attempt, it does not get re-downloaded. I don't see timestamps changing in macOS. I also ran it in debug mode and verified there's only a call to download an attachment if either the submission isn't in the database or the attachment file doesn't exist on disk.

Is it possible that you cleared your instances database between pulls somehow? I can reproduce the behavior you describe by deleting the info.hsqldb folder for my form, restarting Briefcase, and then pulling again. That behavior should be the same as in v1.17.4 and prior. That seemed like the safest approach to me -- if the db doesn't know about a specific instance, something weird happened so it makes sense to me to download all of it.

There's one more edge case I can think of. If you download a submission with many attachments and cancel part way, that submission won't be written to the db. So next time you do a pull, all of its attachments will be downloaded again. Again, the thinking there is that if there was a cancel part way through, we can't be confident of the integrity of the submission on disk so let's redownload all of it.

mmarciniak90:
> Is it possible that you cleared your instances database between pulls somehow?

I recorded my steps. Take a look.

kkrawczyk123:
And it's the second thing that we have conflict about is I can't see that replacing that @mmarciniak90 is encountering on my device. The only difference that we are aware of is that I am on Ubuntu 16 and Marzena uses 18.
Peek 2020-10-29 12-21

lognaturel:
@mmarciniak90 this looks like the last edge case I mentioned. I see that you cancel after two media files are downloaded even though there are more. That means the submission is not recorded in the database as being downloaded. So the next time you pull, this whole submission is pulled again regardless of what is on disk. Does that make sense?

I’m going to make another quick change so “Success” doesn’t show up after a cancel, that’s annoying.

mmarciniak90:
> Does that make sense?

Yes, so, media are not redownloaded only if the whole submission has been downloaded before cancelation.

Peek 2020-10-29 16-23

lognaturel:
There's some amount of parallelism so I think what's going on is that your cancel stopped things before a few of the submissions were marked as downloaded. If you let the download go all the way the first time, are there any media files redownloaded the second time? The other thing to try is that even if you let the download go all the way, if you then delete one media file in any instance, it should be redownloaded. The other ones in that instance should not be. If those two things work, I'm pretty confident that what you're experiencing is just several pulls that were cancelled before they were marked as complete.

lognaturel:
> media are not redownloaded only if the whole submission has been downloaded before cancelation.

Reading this over now, I think you're saying the behavior is as-expected? That is, media are redownloaded for submissions that were not fully downloaded. For submissions that were fully downloaded, media is not redownloaded. If that's the case, all is well. And my comment above explains why there might be more than one incomplete submission after a cancel.

kkrawczyk123:
I was also able to reproduce the edge case. We have verfied on Ubuntu and MacOS different pull and cancel scenarios form Central and Aggregate. As the situation is clear now we should merge it and start regression testing as fast as possible. 

PR #896, opened 2020-12-07
Small fix to remove the csv qualifier and \n from text before generating csv file.

anaskhl:
Closes #


lognaturel:
Thanks for your interest in Briefcase! As you can see, this change breaks behavior that we have explicit tests for. If you're having a problem with an export CSV, please start with describing that issue at https://forum.getodk.org/c/support/6 so we can discuss what possible actions to take would be.

anaskhl:
Thanks, Helene, I realized that. The problem for me is the repeated NewLine values in string, causing my CSV columns values to shift. When I parse it with SSIS. I wanted to perform the fix, I generated the jar file, and it worked for me, but when I pushed the code, I realized you have the tests implemented for this.
I am new to the briefcase, I tried to follow the instructions, but I missed some. My apologies.

Issue #895, opened 2020-11-13
On export, skip encrypted submissions with manifest but no .enc file and mark as failed


lognaturel:
#### Software versions
Briefcase v1.18 and prior

#### Problem description
Submissions to encrypted forms can be missing the submission file itself. In that case, export crashes.

See https://forum.getodk.org/t/odk-briefcase-1-18-export-failing-with-encrypted-forms/31069

#### Steps to reproduce the problem
Set up an encrypted form with submissions. Delete the `.enc` file for some submissions. Try to export.

#### Expected behavior
Export should not crash. Incomplete submissions should be marked as problematic.

#### Other information
Similar to https://github.com/getodk/briefcase/pull/884

PR #897, opened 2021-01-19
On export, skip encrypted submissions with manifest but no .enc file …

anaskhl:
…and mark as failed #895

Closes #


anaskhl:
When reading the list of submissions for parsing, A check is done if the form is encrypted, and submission.xml.enc is not available then log that the file is not available and exclude from the submissions.

anaskhl:
It looks like the build failed because of java version issue. How can we fix this?

